PR #9, opened 2016-07-25
Fix nested bullet formatting

LegalizeAdulthood:


asuessenbach:
Has already been fixed with #11.

Issue #16, opened 2016-08-07
Use cmake external projects and cmake build rules for tinyxml and the spec source


jherico:
Many cmake projects will build their upstream dependencies if necessary by downloading either a binary or a source zip file, using the cmake `ExternalProject_Add` functionality.  However, this library can't be used in this way because the downloaded ZIP file doesn't contain the git submodules required to build.  

Instead of using a git submodule for the tinyxml dependency, you can use an `ExternalProject_Add` command of your own.  Instead of  using a submodule for the spec dependency, you can use a `file(DOWNLOAD ...)` command to fetch the spec XML.  This also has the advantage of allowing the build process to target a specific version of the spec at cmake configuration time by passing a command line argument to override the default spec URL.

mtavenrath:
I understand the need to be able to download the project as zip and keep it alive for those cases. On the other side the git submodules are also quite convenient keeping the project as complete stable part as it is without having to run external scripts.

One thing I can imagine is running the 3rdparty download/build scripts if the dependencies are not in the folder hierarchy. 

Another option would be a (cmake) generator script which gets a vulkan version and which takes care of the rest.

What do you think of those options?

jherico:
>  On the other side the git submodules are also quite convenient keeping the project as complete stable part as it is without having to run external scripts.

CMake is already a required component for creating a project file, so having CMake also download the dependencies and build them as well isn't really an 'external script'.  It's built in functionality of CMake.  You can keep the dependency as stable as desired.  If you look #17 , you'll see that the version of TinyXML it downloads is the precise commit that used to be referenced as a git submodule.  

Although I've targeted `https://raw.githubusercontent.com/KhronosGroup/Vulkan-Docs/1.0/src/spec/vk.xml` for `VK_SPEC_URL`, meaning a build will fetch the most recent version of the spec, you can easily maintain stability by setting it to point at a given commit URL as well.

> Another option would be a (cmake) generator script which gets a vulkan version and which takes care of the rest.

I don't understand the distinction between what you're suggesting here and what I've done in #17 .   Can you elaborate?  

[...]

PR #17, opened 2016-08-07
Use cmake instead of git submodules for tinyxml and spec dependencies.

jherico:
Fix for #16

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -30,7 +30,7 @@ project(VulkanHppGenerator)
 
 # If you wish to target a specific version of the spec, override this value like so 
 # cmake -DVK_SPEC_URL:STRING=https://example.com/some/other/url/vk.xml
-set(VK_SPEC_URL "https://raw.githubusercontent.com/KhronosGroup/Vulkan-Docs/1.0/src/spec/vk.xml" CACHE STRING "URL from which to fetch the Vulkan Spec XML")
+set(VK_SPEC_URL "https://raw.githubusercontent.com/KhronosGroup/Vulkan-Docs/f4c4113d070c0c9b55b4482055be054dd6466c96/src/spec/vk.xml" CACHE STRING "URL from which to fetch the Vulkan Spec XML")
 
 # Download the request URL specification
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -28,42 +28,56 @@ cmake_minimum_required(VERSION 3.2)
...
+set(VK_SPEC_URL "https://raw.githubusercontent.com/KhronosGroup/Vulkan-Docs/1.0/src/spec/vk.xml" CACHE STRING "URL from which to fetch the Vulkan Spec XML")
 
-file(TO_NATIVE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/vulkan/vulkan.hpp vulkan_hpp)
-string(REPLACE "\\" "\\\\" vulkan_hpp ${vulkan_hpp})
-add_definitions(-DVULKAN_HPP="${vulkan_hpp}")
-
-set(HEADERS
-)
+# Download the request URL specification
+file(DOWNLOAD ${VK_SPEC_URL} ${CMAKE_CURRENT_BINARY_DIR}/vk.xml )
```
mtavenrath: This way of fetching vk.xml does not update the spec automatically if you change the CMakeLists.txt. To update one would have to delete the cache and configure the project again.

I'm wondering if it'd make sense to keep the vk.xml as external subproject and let an application download a different spec if required. The build process would be:

Download Vulkan-Hpp project (as zip)
Download your favorite vk.xml to some location
unzip Vulkan-Hpp
configure & vbuild vulkan.hpp
run VulkanHppGenerator <path to your favorite vk.xml>

As alternate solution the url of the spec could be in a separate file and a custom target with dependency on that file would download vk.xml during the build process. VulkanHppGenerator would have to depend on that custom target.

oddhack:
@mtavenrath putting vk.xml (really, the entire src/spec area of the Vulkan-Docs tree) into a separate repo and transcluding it via submodule is on my queue for the spec already. Won't happen immediately, but probably within 6-8 weeks.

CMakeLists.txt
ioquatix: Why not download the vk.xml spec with a version number in the file name so it's clear what version you are getting/building?

Issue #14, opened 2016-08-03
Error compiling header, == not defined for certain types


FancyBanana:
I had setup a basic project from [this Vulkan tutorial](https://vulkan-tutorial.com/) and i've replaced  
`vulkan.h` with `vulkancpp.h`( which is `vulkan.hpp`) and i get these numerous compiler errors:  
Visual Studio 2015 Community, toolset v140.  

```
1>d:\vulkansdk\1.0.21.1\include\vulkan\vulcancpp.h(2598): error C2678: binary '==': no operator found which takes a left-hand operand of type 'const vk::Buffer' (or there is no acceptable conversion)
1>  d:\vulkansdk\1.0.21.1\include\vulkan\vulcancpp.h(2598): note: could be 'built-in C++ operator==(VkBuffer, VkBuffer)'
1>  d:\vulkansdk\1.0.21.1\include\vulkan\vulcancpp.h(2598): note: or       'built-in C++ operator==(VkBuffer, bool)'
1>  d:\vulkansdk\1.0.21.1\include\vulkan\vulcancpp.h(2598): note: or       'built-in C++ operator==(bool, VkBuffer)'
1>  d:\vulkansdk\1.0.21.1\include\vulkan\vulcancpp.h(2598): note: or       'built-in C++ operator==(bool, bool)'
1>  d:\vulkansdk\1.0.21.1\include\vulkan\vulcancpp.h(2598): note: while trying to match the argument list '(const vk::Buffer, const vk::Buffer)'
1>d:\vulkansdk\1.0.21.1\include\vulkan\vulcancpp.h(3776): error C2678: binary '==': no operator found which takes a left-hand operand of type 'const vk::DisplayKHR' (or there is no acceptable conversion)
1>  d:\vulkansdk\1.0.21.1\include\vulkan\vulcancpp.h(3776): note: could be 'built-in C++ operator==(VkDisplayKHR, VkDisplayKHR)'
1>  d:\vulkansdk\1.0.21.1\include\vulkan\vulcancpp.h(3776): note: or       'built-in C++ operator==(VkDisplayKHR, bool)'
1>  d:\vulkansdk\1.0.21.1\include\vulkan\vulcancpp.h(3776): note: or       'built-in C++ operator==(bool, VkDisplayKHR)'
1>  d:\vulkansdk\1.0.21.1\include\vulkan\vulcancpp.h(3776): note: or       'built-in C++ operator==(bool, bool)'
[...]

PR #21, opened 2016-08-12
Made the generated header 32-bit safe by explicitly casting vk::* obj…

asuessenbach:
…ects to Vk* in all operator==().


asuessenbach:
This resolves issue #14

mtavenrath:
I'd prefer not to merge this change since it fixes a sympton, but not the root cause which is that operator== and operator!= are missing for the handles. I've created a new pull request which adds those.

PR #31, opened 2016-09-06
Added template class scoped_handle, which is default enabled.

xXSeraphimXx:
Added template class scoped_handle, which supports auto deletion of the objects, after they have been released (must!) and gone out of scope. Also added option to disable enabled scope handling with simple defining the macro VULKAN_HPP_DISABLE_SCOPED_HANDLING.

PS:
I hope the code is right. I tested the code as much as it was possible with my current knowledge! If there are any questions according code, feel free to ask. I didn't added any comments, because the header itself hadn't such. That was quite a lot code I added 😲


xXSeraphimXx:
I'm very sorry for that non-transparency. I'm a bit unconform with github more used to BitBucket. I'll close this and make a new push request with less complications.

PR #32, opened 2016-09-07
Added template class scoped_handle and create functions with those, which is default enabled.

xXSeraphimXx:
Hope it fits. If there are any questions, feel free to ask!


mtavenrath:
We've added a UniqueHandle to Vulkan-Hpp. Closing.

PR #42, opened 2016-10-18
Update documentation to include latest features

mtavenrath:


README.md
```diff
@@ -1,11 +1,57 @@
...
+* Clang >= 3.3
+
+# namespace vk
+To avoid name collisions with the Vulkan C API the C++ bindings reside in the vk namespace. The following rules apply to the new naming
+* All functions, enums and structs have the Vk prefix removed.
+  * ```vkCreateImage``` can be accessed as ```vk::CreateImage```
+  * ```VkImageTiling``` can be accessed as ```vk::ImageTiling```
+  * ```VkImageCreateInfo``` can be accessed as ```vk::ImageCreateInfo```
+* Enums are mapped to scoped enums to provide compile time type safety. The names have been changed to 'e' + CamelCase with the VK_ prefix and type infix removed. In case the enum type is an extension the extension suffix has been removed from the enum values.
+ In all other cases the extension suffix will has not been removed.
```
nsubtil: > extension suffix will has not been removed

Did you mean "extension suffix has been removed"?

README.md
```diff
@@ -1,11 +1,57 @@
...
+  * ```VK_COLOR_SPACE_SRGB_NONLINEAR_KHR``` is now ```vk::ColorSpaceKHR::eSrgbNonlinear```.
+  *  ```VK_STRUCTURE_TYPE_PRESENT_INFO_KHR``` is now ```vk::StructureType::ePresentInfoKHR```.
+*  Flag bits are handled like scoped enums with the addition that the _BIT suffix has also been removed.
+
+# Handles
+Vulkan-Hpp declares a class for all handles to ensure full type safety and to add support for member functions on handles. A member function has been added to a handle class for each
+ function which accepts the corresponding handle as first parameter. Instead of ```vkBindBufferMemory(device, ...)``` one can write ```device.bindBufferMemory(...)``` or ```vk::bindBufferMemory(device, ...)```.
+
+# C/C++ Interop for Handles
+On 64-bit platforms Vulkan-Hpp supports implicit conversions between C++ Vulkan handles and C Vulkan handles. On 32-bit platforms all non-dispatchable handles are defined as ```uint64_t``` and thus not typesafe.
```
nsubtil: > ... and thus not typesafe

I would say something like "... defined as uint64_t, thus preventing type-conversion checks at compile time which would catch assignments between incompatible handle types."

README.md
```diff
@@ -1,11 +1,57 @@
...
+Vulkan-Hpp declares a class for all handles to ensure full type safety and to add support for member functions on handles. A member function has been added to a handle class for each
+ function which accepts the corresponding handle as first parameter. Instead of ```vkBindBufferMemory(device, ...)``` one can write ```device.bindBufferMemory(...)``` or ```vk::bindBufferMemory(device, ...)```.
+
+# C/C++ Interop for Handles
+On 64-bit platforms Vulkan-Hpp supports implicit conversions between C++ Vulkan handles and C Vulkan handles. On 32-bit platforms all non-dispatchable handles are defined as ```uint64_t``` and thus not typesafe. 
+Due to that Vulkan-Hpp does not enable implicit conversion for 32-bit platforms by default and it is recommended to use a static_cast for the conversion like this: ```VkDevice = static_cast<VkDevice>(cppDevice)```
+to prevent converting some arbitrary int to a handle or vice versa by accident. If you're developing your code on a 64-bit platform, but want compile your code for a 32-bit platform without adding the explicit casts
+you can define ```VULKAN_HPP_TYPESAFE_CONVERSION``` to 1 in your build system or before including ```vulkan.hpp```.
```
nsubtil: It would be worth clarifying whether setting VULKAN_HPP_TYPESAFE_CONVERSION would affect 64-bit builds at all.

README.md
```diff
@@ -24,192 +70,200 @@ Have a look at the following piece of code which creates a VkImage:
...
-* uninitialized data fields (e.g. missing initialization of ```ci.mipLevels```) 
-* use of invalid bits for ```ci.flags``` (no type-safety for bits)
-* use of incorrect enums for fields (no type-safety for enums)
-
-These initializations will most likely show up as random runtime errors, which usually are nasty and time-consuming to debug.
-Our auto-generated, C++ 11-conform layer uses commonly known C++ features like implicit initialization through constructors
-to avoid incorrect or missing  initializations and introduces type-safety with scoped enums to turn explicit initialization
-errors into compile errors. Following is a list of features and conventions introduced by our Vulkan C++ layer:
+There are two typical issues Vulkan developers encounter when filling out a CreateInfo struct field by field
+* One or more fields are missing thus holds either 0 or random values.
```
nsubtil: > One or more fields are missing thus holds either 0 or random values

Would rewrite this as "One or more fields are left uninitialized"

README.md
```diff
@@ -1,11 +1,57 @@
...
+ with the improvement that it is impossible to set bits not specified by the corresponding enum by accident. Here are a few examples for the bitmask handling:
 
-Vulkan is a C API and as such inherits all common pitfalls of using a general C programming library. The motivation of a low-level Vulkan C++ API is to avoid these common pitfalls by applying commonly known C++ features while keeping the overall structure of a Vulkan program and preserving the full freedom it provides as low-level graphics API. An additional guideline we followed was not to introduce additional runtime overhead by providing a header-only library with inline functions.
+```c++
+vk::ImageUsage iu1; // initialize a bitmask with no bit set
+vk::ImageUsage iu2 = {}; // initialize a bitmask with no bit set
+vk::ImageUsage iu3 = vk::ImageUsage::eColorAttachment; // initialize with a single value
+vk::ImageUsage iu4 = vk::ImageUsage::eColorAttachment | vk::ImageUsage::eStorage; // or two bits to get a bitmask
+PipelineShaderStageCreateInfo ci( {} /* pass a flag without any bits set, ...);
```
nsubtil: It would be desirable to make 0 work as the "no flags set" value as well. Does it?

README.md
```diff
@@ -24,192 +70,200 @@ Have a look at the following piece of code which creates a VkImage:
...
+vec.push_back(scissorRect1);
+vec.push_back(scissorRect2);
+c.setScissor(0, vec);
 ```
 
-By adding the scoped enums and ```vk::Flags<T>``` the C++ API provides type safety for all enums and flags which is a
-big improvement. This leaves the remaining issue that the compiler might not detect uninitialized fields in structs. As a solution
-we have added constructors to all structs which accept all values defined by the corresponding struct. 
+# Passing Structs to Functions
+Vulkan-Hpp generates references for pointers to structs. This conversion allows pass temporary structs to functions which can result in shorter code. In case the input is optional and thus accepting a null pointer the parameter type will be
```
nsubtil: > This conversion allows pass temporary structs ...

This conversion allows _passing_ temporary structs...

nsubtil: > In case the input is optional and thus accepting a null pointer ...

In cases where the input is optional and can be a null pointer ...

README.md
```diff
@@ -24,192 +70,200 @@ Have a look at the following piece of code which creates a VkImage:
...
-Optional parameters are being replaced by ```Optional<T>``` which accept a type of ```const T```, ```T```, or ```const std::string```. ```nullptr``` can be used to initialize an empty ```Optional<T>```.
-
-Here are a few code examples:
-```c++
-  try {
-    VkInstance nativeInstance = nullptr; // Fetch the instance from a favorite toolkit
-
-    // create a vk::Instance handle from a native handle
-    vk::Instance i(nativeInstance);
+The second snipped shows how to use the API using return value transformation, but without exceptions. It’s already a little bit shorter than the original code:
```
nsubtil: > The second snipped ...

snippet

README.md
```diff
@@ -24,192 +70,200 @@ Have a look at the following piece of code which creates a VkImage:
...
+  handle error code;
+  cleanup?
+  return?
+}
+```
 
-    // Get a std::vector as result of an enumeration call.
-    std::vector<vk::PhysicalDevice> physicalDevices = i.enumeratePhysicalDevices();
-    vk::FormatProperties formatProperties = physicalDevices[0].getFormatProperties(vk::Format::eR8G8B8A8Unorm);
+A nicer way to unpack the result is provided by the structured bindings of C++17. They will allow us to get the result with a line line of code:
```
nsubtil: > ... get the result with a line line of code

with a single line of code?

README.md
mtavenrath: I mean 'extension suffix has not been removed'

README.md
mtavenrath: Up to a second ago I didn't have a solution for this problem. 

Now I'm wondering if it'd be acceptable to provide a constructor which accepts a nullptr_t. This way one can pass a constant 0 to flags. The downside is one can also pass a nullptr to create a flags object with no bits set. Is this too confusing?

README.md
mtavenrath: yes

PR #58, opened 2017-01-24
Add allocator argument to enhanced functions that return a vector

wesleygriffin:
C++17 will support polymorphic allocators where std::vector can be handed an allocator on construction. I've modified the function generation code to take a defaulted allocator argument that will be passed along to the std::vector. This enables code like the following using
 [Howard Hinnant's stack allocator](https://howardhinnant.github.io/stack_alloc.html).

```
vk::PhysicalDevice physical;
// ...

using extensions_vector = std::vector<vk::ExtensionProperties, short_alloc<T, 8192, alignof(T)>>;
using allocator = extensions_vector::allocator_type;
allocator::arena_type arena;
extensions_vector extensions{arena};
std::tie(result, extensions) = physical.enumerateDeviceExtensionProperties(nullptr, allocator{arena});
```

### DIFF
--- a/VulkanHppGenerator.cpp
+++ b/VulkanHppGenerator.cpp
@@ -2078,8 +2078,14 @@ void writeFunctionBody(std::ofstream & ofs, std::string const& indentation, std:
             }
           }
           assert(!size.empty());
-          ofs << "( " << size << " )";
-        } else if (commandData.twoStep)
+          ofs << "( " << size;
+          if (returnType.find("Allocator") != std::string::npos)
+          {
+            ofs << ", " << commandData.arguments[returnIndex].pureType << "(), alloc";
+          }
+          ofs << " )";
+        }
+        else if (commandData.twoStep)
         {
             ofs << "( alloc )";
         }
@@ -3257,7 +3263,7 @@ int main( int argc, char **argv )
         << arrayProxyHeader;
 
     // first of all, write out vk::Result and the exception handling stuff
-    std::list<DependencyData>::const_iterator it = std::find_if(vkData.dependencies.begin(), vkData.dependencies.end(), [](DependencyData const& dp) { return dp.name == "Result"; });
+    std::list<DependencyData>::iterator it = std::find_if(vkData.dependencies.begin(), vkData.dependencies.end(), [](DependencyData const& dp) { return dp.name == "Result"; });
     assert(it != vkData.dependencies.end());
     writeTypeEnum(ofs, *it, vkData.enums.find(it->name)->second);
     writeEnumsToString(ofs, *it, vkData.enums.find(it->name)->second);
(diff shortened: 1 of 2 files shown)


wesleygriffin:
For the CLA, I am a US government employee, therefore this contribution is not subject to copyright protection and is in the public domain as per [NIST Licensing](https://www.nist.gov/director/licensing).

Also, I am not sure I can sign the CLA, I would have to have some internal talks. Since my contribution is in the public domain by US code, I don't think the CLA could even apply, because I don't have a copyright to assign to Khronos.

Thanks,
Wes

ioquatix:
Good idea.

wesleygriffin:
Is there any interest from the maintainers in merging this?

Like I said in the PR, I don’t believe the CLA applies because my work is automatically in the public domain.

But if signing the CLA is holding this up, I can start talking to my management about it.

Thanks,
Wes

mtavenrath:
@wesleygriffin We really appreciate your PR and we'd like to merge it. Though we have the CLA in place and we have to evaluate if we can accept a PR without the CLA signed.

There's one easy option to get this feature in quite quick. We could treat your PR description as feature request and implement it ourselves. Is this something you'd be happy with for the case you can't to sign the CLA and we cannot accept pull requests without the CLA signed?

wesleygriffin:
Re-implementing yourselves is fine with me, I don't need my PR directly merged.

I will start an internal discussion about signing the CLA, but as I said, my contributions are explicitly in the public domain without any copyright protection.

Thanks,
Wes

________________________________________

ioquatix:
@wesleygriffin Requiring a CLA to be signed is probably just to protect everyone involved, yourself, Khronos, and anyone who chooses to use this library.

However, if what you've done is truly in the public domain, then it shouldn't be a problem. This (legal) issue feels borderline pedantic, but I can certainly see why Khronos would have something like this in place.

wesleygriffin:
Don't get me wrong - I absolutely agree that the CLA is important and useful. I'm simply in a unique position where I work for a US Government Agency (NIST) and therefore anything I create is automatically in the public domain. And on top of that, there are strict internal rules about what I'm allowed to sign so I would need to have someone look over the CLA to determine if I am allowed to sign it.

But I'm more than happy to have this be treated as a feature request and not have the PR merged. That way I don't need to maintain my own fork to use my custom allocator.

Thanks,
Wes

________________________________________

mtavenrath:
@wesleygriffin No worries and thank you for your patience. I really appreciate your contribution and we're treating it as feature request so that neither the CLA, nor your position prevents this useful feature. We're going to add it soon.

wesleygriffin:
Not a problem! I’ve also started working internally the possibility of signing the CLA, but I wouldn’t expect anything to happen on that for several months at best.

Thanks,
Wes

mtavenrath:
@wesleygriffin I've added allocator objects as last arguments to all functions which do return a vector. Closing.

PR #78, opened 2017-02-13
UniqueHandle - Fixed compile error in 32-bit

spfuetzner:
Because in 32-bit Vulkan-handles are integers, the assignment from
nullptr is not valid and doesn't compile. VK_NULL_HANDLE on the other
hand abstracts the differences between 32- and 64-bit environments.

### DIFF
--- a/VulkanHppGenerator.cpp
+++ b/VulkanHppGenerator.cpp
@@ -515,7 +515,7 @@ std::string const uniqueHandleHeader = { R"(
     Type release()
     {
       Type value = m_value;
-      m_value = nullptr;
+      m_value = VK_NULL_HANDLE;
       return value;
     }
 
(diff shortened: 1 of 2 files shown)


asuessenbach:
This change does not compile in 32 bit, as there's no operator=() that takes an int. Note that m_value is not a vkType, but a vk::Type!

spfuetzner:
Ah, I see the problem now. The fix only works, if VULKAN_HPP_TYPESAFE_CONVERSION is defined. There is still the problem that `m_value = nullptr;` doesn't compile because vk::Type has no constructor taking a nullptr_t or any other pointer. I'll retry it with another fix.

asuessenbach:
Please have a look at pull-request #77. That resolves this issue.

Issue #79, opened 2017-02-13
Vulkan Header 1.0.40: Vulkan HPP ignores 'supported=disabled' tag in vk.xml


mark-lunarg:
Building with the 1.0.40 header and vk.xml file from Vulkan-Docs results in undefined errors for `VK_EXT_SMPTE2086_metadata` extension members.  This extension is tagged as '`disabled`' in vk.xml but is picked up regardless, causing compilation failures.

mark-lunarg:
@asuessenbach, thanks for the quick response.  Works perfectly!

PR #80, opened 2017-02-14
Update to Vulkan 1.0.40; added handling of types and commands from di…

asuessenbach:
…sabled extensions


asuessenbach:
Fixes issue #79.

Issue #97, opened 2017-04-10
Vulkan-Hpp specific exception classes


RippeR37:
What do you think about adding custom exception/error classes derived from one of the `std::exception` derived classes? Currently without special filtering developer can't really know if given exception (like `std::system_error` was thrown by Vulkan-Hpp or other piece of code and with that developers would be able to use natural exceptions "filtering" like:
```
try {
    // some Vulkan-Hpp code that can throw
    // with some normal code that can throw
} catch(const vk::error& vk_error) {
    // error from vulkan-related code
} catch(const std::exception& exception) {
    // error related to normal code / STL / etc.
}
```
Since exceptions already add some small overhead (mostly when fired), maybe we could also add hierarchy of exceptions representing some scenarios based on VkResult return value like
```
vk::error
    - vk::out_of_memory_error
        - vk::out_of_device_memory_error
        - vk::out_of_host_memory_error
    - vk::initialization_failed_error
    - vk::device_lost_error
    - ...
```
That way some errors might be handled right away and others might for example terminate application.
Worst case you could still catch them by `std::exception` or another STD-base-class.

asuessenbach:
I think, this is a good idea. I will introduce them in the next couple of days.

RippeR37:
I'm glad you liked it. I was thinking on how this should be implemented because standard exception classes doesn't inherit virtually, so we can't simply make:
```
vk::Error : std::exception
vk::SystemError : Error, std::system_error
vk::OutOfDeviceMemoryError : SystemError
```
and I've came up with couple ideas but none were perfect. Above implementation tries to:
- achieve backwards compatibility with existing code (exceptions inherit from `std::logic_error` or `std::system_error`)
- not double-inherit anything
- allow for grouping of Vulkan-Hpp exceptions and catch only them
- use existing STL exceptions like `std::logic_error` and `std::system_error` so that systems that work with them can be used here too
- expose some API to get exception message in base Vulkan-Hpp exception class

It's not perfect but without virtual inheritance between STL exceptions I think we can't do better.

P.S. I haven't updated `vulkan.hpp` there because for some reasons rebuilding it on my system changes each line, even without any changes...

asuessenbach:
Regarding the changes on every line: this is probably due to a different end-of-line setting on your end. We're using CR/LF (Windows-style).

asuessenbach:
Resolved with pull-request #99.

PR #99, opened 2017-04-14
Introduce custom exceptions for Vulkan-Hpp (#97)

RippeR37:
Base exception architecture:

* vk::Error
** vk::LogicError : std::logic_error
** vk::SystemError : std::system_error
*** vk::$ErrorResultEnumName$Error : SystemError

This is to have backward compatibility with existing code which was catching std::logic_error or std::system_error.

### DIFF
--- a/VulkanHppGenerator.cpp
+++ b/VulkanHppGenerator.cpp
@@ -62,6 +62,57 @@ const std::string exceptionHeader(
   "\n"
   );
 
+std::string const exceptionClassesHeader = (
+  "#if defined(_MSC_VER) && (_MSC_VER == 1800)\n"
+  "# define noexcept _NOEXCEPT\n"
+  "#endif\n"
+  "\n"
+  "  class Error\n"
+  "  {\n"
+  "    public:\n"
+  "    virtual ~Error() = default;\n"
+  "\n"
+  "    virtual const char* what() const noexcept = 0;\n"
+  "  };\n"
+  "\n"
+  "  class LogicError : public Error, public std::logic_error\n"
+  "  {\n"
+  "    public:\n"
+  "    explicit LogicError( const std::string& what )\n"
+  "      : Error(), std::logic_error(what) {}\n"
+  "    explicit LogicError( char const * what )\n"
+  "      : Error(), std::logic_error(what) {}\n"
+  "    virtual ~LogicError() = default;\n"
+  "\n"
+  "    virtual const char* what() const noexcept { return std::logic_error::what(); }\n"
+  "  };\n"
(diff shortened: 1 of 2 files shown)


VulkanHppGenerator.cpp
```diff
@@ -62,6 +62,53 @@ const std::string exceptionHeader(
...
 
+std::string const exceptionClassesHeader = (
+  "#if defined(_MSC_VER) && (_MSC_VER == 1800)\n"
+  "# define noexcept _NOEXCEPT\n"
+  "#endif\n"
+  "\n"
+  "  class Error\n"
+  "  {\n"
+  "    public:\n"
+  "    virtual const char* what() const noexcept = 0;\n"
```
asuessenbach: I think, there's no need to have this function here, as both std::logic_error and std::system_error already provide that.
In order to make vk::Error a polymorphic type, it should be enough to have, for example, a virtual destructor here.

RippeR37: I added this, because - since we can't inherit from both `std::exception` and `std::system_error` - when user would want to catch all Vulkan exceptions at once, he couldn't do anything with it. Imagine:
```
try {
    // ...
} catch (/* something specific */) {
    // ...
} catch (const vk::Error& vk_error) {
    // What to do with `vk_error`? Can't even print what was the problem...
    // not much more useful then `catch (...)`
}
```
So I think that it would be nice to have at least `what()` available there. If you disagree with this argument I will remove this.

As for virtual destructor - nice catch, I should have seen this... :+1:

RippeR37:
@asuessenbach 
Not to pollute history I've amended last commit and force-pushed changes (but I made sure to re-edit so new metadata would appear) - I hope it's fine.

Changes I made:
* Added virtual destructors for `vk::Error`, `vk::SystemError` and `vk::LogicError`,
* Removed trailing space from `throwResultException()` definition,
* Added `vulkan.hpp` generated from current code.

P.S. I seriously have no idea why you chose to use CRLF for `vulkan.hpp` and LF for `VulkanHppGenerator.cpp` - this is really confusing.

asuessenbach:
@RippeR37 would you mind to resolve the conflicts or your pull-request that occurred with the latest merge on the main branch?

RippeR37:
@asuessenbach Since this cleanup was so big, I wanted to make sure that everything is okey so I've synced-up with current master and rebased my changes by hand on it. Basically everything is same, one function was deleted by this cleanup so my changes don't belong there anymore and I've added forward declaration of used functions to match what already is in code.

You can check it out, for me it looks okey.

PR #110, opened 2017-06-01
Update to Vulkan 1.0.50

mtavenrath:
* The naming scheme from the Vulkan spec would Ycbcr for YCBCR in enums, but is using YCbCr for structs which are linked to those enums. Introduce a replacement vector to address this issue.


asuessenbach:
This is outdaged with the update to Vulkan 1.0.51. Closing.

Issue #122, opened 2017-09-05
Better unique handles to simplify updating existing code.


ioquatix:
The current system for unique handles is useful but cumbersome especially when migrating existing code which doesn't use them. I'm not sure how useful these ideas are but I think it would be good to have a discussion about it.

It would be good to discuss the following improvements:

- Prefer `std::unique_ptr` as base implementation.
- Type dependent allocation/initialization. Rather than `vk::DeviceDeleter`, expose `vk::Device::Deleter`.
- Use a template for the majority of deleter implementations. You know you can pass function pointers as template parameters.
- Use the type dependent lookup to automatically create the deleter appropriately rather than having to manually invoke specific methods.

Ideally, it would be possible just to change the type signature, and have everything continue to work. In order to do that, you'd need a base wrapper which could be automatically converted to a unique pointer.

```c++
vk::Device device;
vk::Unique<vk::Device> unique_device;

device = _physical_device.createDevice(device_info, alloc());
unique_device = device; // take ownership
```

or probably more reasonably:

```c++
vk::Unique<vk::Device> device = _physical_device.createDevice(device_info, alloc());
```

Otherwise, a lot of changes to the codebase are required to change `createX` to `createXUnique`.

ioquatix:
I'm aware that the design means that the code above is not possible (that the deleter needs more information). But, I'm just giving examples, and perhaps we can discuss (a) whether this is something worth pursuing and (b) how it could be achieved.

ioquatix:
Perhaps one option is to make a `vk::Value<T>` which contains all the information to generate either a `vk::T` or `vk::UniquePointer<T>` instance. `createX` would return `vk::Value<X>` and you could either assign this to a raw value or a handle which deletes it. The compiler would take care of the rest.

ioquatix:
it would be pretty awesome if the semantics of `vk::Value<T>` were similar to `vk::UniquePointer`, in terms of being a container and the use of `->` vs `.`. That would ease code migration since you could move things all to `vk::Value<T>` as a first step and then to `vk::UniquePointer<T>` next, one thing at a time.

mtavenrath:
Those are all good ideas. One thing we've to be careful about is breaking compatibility with existing applications, so all changes have to be compile-time compatible. 

> Prefer std::unique_ptr as base implementation.

[...]

PR #123, opened 2017-09-05
Add UniqueHandle non-const accessors, ensure all by reference.

ioquatix:
- Add `Type * operator->()`
- Fix `Type get() const` to `const Type & get() const`
- Add `Type & get()`

Fixes https://github.com/KhronosGroup/Vulkan-Hpp/issues/122#issuecomment-327198391

### DIFF
--- a/vulkan/vulkan.hpp
+++ b/vulkan/vulkan.hpp
@@ -361,13 +361,28 @@ namespace vk
     {
       return &m_value;
     }
-
+    
+    Type * operator->()
+    {
+      return &m_value;
+    }
+    
     Type const& operator*() const
     {
       return m_value;
     }
 
-    Type get() const
+    Type & operator*()
+    {
+      return m_value;
+    }
+
+    const Type & get() const
+    {
+      return m_value;
+    }
+    
+    Type & get()
     {
       return m_value;
     }
--- a/VulkanHppGenerator.cpp
+++ b/VulkanHppGenerator.cpp
@@ -381,8 +381,8 @@ const std::string structureChainHeader = R"(
 
     StructureChain& operator=(StructureChain const &rhs)
     {
-      linkAndCopy(rhs);
-      return this;
+      linkAndCopy<StructureElements...>(rhs);
+      return *this;
     }
 
     template<typename ClassType> ClassType& get() { return static_cast<ClassType&>(*this);}
--- a/VulkanHppGenerator.cpp
+++ b/VulkanHppGenerator.cpp
@@ -607,13 +607,28 @@ const std::string uniqueHandleHeader = R"(
     {
       return &m_value;
     }
-
+    
+    Type * operator->()
+    {
+      return &m_value;
+    }
+    
     Type const& operator*() const
     {
       return m_value;
     }
 
-    Type get() const
+    Type & operator*()
+    {
+      return m_value;
+    }
+
+    const Type & get() const
+    {
+      return m_value;
+    }
+    
+    Type & get()
     {
       return m_value;
     }


mtavenrath:
@ioquatix Can you add the generated vulkan.hpp to this PR?

ioquatix:
Sure, I wasn't sure if it was appropriate to do that or not.

ioquatix:
I ran the ./VulkanHppGenerator but the entire file was different.. line endings?

mtavenrath:
Line endings is a likely difference.

ioquatix:
Is this not something that's normalised by the generator?

ioquatix:
Bump :)

asuessenbach:
The generator simply puts std::endl at each line end.
Have  you changed the line ending handling of your git config (core.autocrlf)?

ioquatix:
`std::endl` is platform specific?

ioquatix:
Okay, so I just checked and the `vulkan.hpp` files is generated with `\r\n` line endings, probably from Windows, while on Unix it only generates `\n`. This is how `std::endl` works.

So, back to you - what would you like to do?

asuessenbach:
Could you please run some unix2dos converter on the generated vulkan.hpp to just get the "real" changes for that file?
We'll address that issue later in a separate PR.

ioquatix:
Okay done.

ioquatix:
It looks like some other PR has been merged in the mean time and I've reverted the change. Perhaps it would make more sense for the generation of vulkan.hpp to be automated otherwise this kind of clobbering might become more frequent. I'll rebase.

vulkan/vulkan.hpp
```diff
@@ -454,8 +469,8 @@ namespace vk
 
     StructureChain& operator=(StructureChain const &rhs)
     {
```
mtavenrath: This change is not expected. Did you generate with header with the latest version of Vulkan-Hpp including your changes?

ioquatix:
Yep.

vulkan/vulkan.hpp
ioquatix: Yes, I just did it using the exact code in this PR.

ioquatix:
Here is the diff from running the header generation tool on the current master:

```
diff --git a/vulkan/vulkan.hpp b/vulkan/vulkan.hpp
index a276624..8aaa8e4 100644
--- a/vulkan/vulkan.hpp
+++ b/vulkan/vulkan.hpp
@@ -454,8 +454,8 @@ namespace vk
 
     StructureChain& operator=(StructureChain const &rhs)
     {
-      linkAndCopy<StructureElements...>(rhs);
-      return *this;
+      linkAndCopy(rhs);
+      return this;
     }
 
     template<typename ClassType> ClassType& get() { return static_cast<ClassType&>(*this);}
```

ioquatix:
It matches up with the code here: https://github.com/KhronosGroup/Vulkan-Hpp/blob/bca6564dac806ea8d30bad792066a3ba963fdbf1/VulkanHppGenerator.cpp#L377-L386

ioquatix:
Why don't I roll back the changes to vulkan.hpp and you can sort it out?

mtavenrath:
I just checked master and it looks like it's missing a change. It'd be nice if you could just change the corresponding lines in VulkanHppGenerator.cpp, but I'm also okay with you rolling back vulkan.hpp and I'll take care of it.

ioquatix:
At this point I think it would just be easier if you took care of it.

ioquatix:
If you fix the master branch I will rebase and regenerate `vulkan.hpp`

realvictorprm:
Small side node regarding the line endings. There should be a git configuration on windows which runs the conversion of windows to Unix endings automatically.

ioquatix:
@realvictorprm Are you suggesting the authors of this code change from Windows line endings to Unix? Because Unix is my preference too.

ioquatix:
Okay I fixed the issue and provided an initial commit to add the missing code (I assume that's what you wanted).

realvictorprm:
@ioquatix during installing _Git for Windows_ the user is asked whether git should autotranslate windows line endings to unix line endings, so yes 😃.

I'm a bit amused that Vulkan.Hpp afterwards got the deleter feature. About 1 1/2 year ago I tried to propose my change with my old account but it was declined. Also I wasn't aware of that I shouldn't edit VulkanHpp directly 😅 ...

PR #125, opened 2017-09-12
Changes necessary to build 1.0.60 header

mikes-lunarg:
I ran into some asserts doing the 1.0.60 header update in https://github.com/KhronosGroup/Vulkan-LoaderAndValidationLayers.  These are the workarounds I used to generate a header.

### DIFF
--- a/VulkanHppGenerator.cpp
+++ b/VulkanHppGenerator.cpp
@@ -1676,7 +1676,7 @@ void readExtensionsExtension(tinyxml2::XMLElement * element, VkData & vkData)
   assert(vkData.tags.find(tag) != vkData.tags.end());
 
   tinyxml2::XMLElement * child = element->FirstChildElement();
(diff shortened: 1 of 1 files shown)


asuessenbach:
Thanks for pointing us to this new issue with 1.0.60. But it took a bit more to handle all those new elements in the vk.xml.
But after extending the generator to correctly handle all that, I encountered an error in the vk.xml, probably completely independent of your findings: there's a command (vkCreateValidationCacheEXT) returning a VkResult, but without any successcodes listed. That information is mandatory for the generator, and therefore, we can't generate a correct header on that version.
An issue on gitlab.khronos is already filed (https://gitlab.khronos.org/vulkan/vulkan/issues/995).

mtavenrath:
@mikes-lunarg We've changed Vulkan-Hpp to work with Vulkan 1.0.61. This change still requires a manual change in vk.xml until the next header will be released.

PR #137, opened 2017-10-10
Add note about exceptions and UniqueHandle

christetreault:
I was trying out vulkan.hpp this morning, and was unable to use UniqueHandle with exceptions disabled. Looking through the code a bit turned up this

```
#if defined(VULKAN_HPP_NO_EXCEPTIONS) && !defined(VULKAN_HPP_NO_SMART_HANDLE)
#  define VULKAN_HPP_NO_SMART_HANDLE
#endif
```

It seems that the use of UniqueHandle requires exceptions. I assume there is a good reason for this, so I've added a note to the readme.md stating that UniqueHandle requires exceptions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -304,7 +304,7 @@ Vulkan-Hpp provides a ```vk::UniqueHandle<Type, Deleter>``` interface. For each 
 For each function which constructs a Vulkan handle of type ```vk::Type``` Vulkan-Hpp provides a second version which returns a ```vk::UnqiueType```. E.g. for ```vk::Device::createBuffer``` there is ```vk::Device::createBufferUnique``` and for ```vk::allocateCommandBuffers```
 there is ```vk::allocateCommandBuffersUnique```.
 
-Note that using ```vk::UniqueHandle``` comes at a cost since most deleters have to store the ```vk::AllocationCallbacks``` and parent handle used for construction because they are required for automatic destruction.
+Note that using ```vk::UniqueHandle``` comes at a cost since most deleters have to store the ```vk::AllocationCallbacks``` and parent handle used for construction because they are required for automatic destruction. The use of ```vk::UniqueHandle``` also requires that exceptions are enabled.
 
 # Custom allocators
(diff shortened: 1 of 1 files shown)


mtavenrath:
We're going to make the UniqueHandle available in case exceptions are disabled. The note won't be required anymore.

mtavenrath:
Closing as the restriction has been removed.

ro0o0o0okie:
Now with the  VULKAN_HPP_NO_EXCEPTIONS  defined, I got the following series of compiling errors:
```shell
vulkan.hpp:28241:83: error: no matching function for call to ‘vk::UniqueHandle<vk::DeviceMemory, vk::DeviceMemoryDeleter>::UniqueHandle(vk::ResultValueType<vk::DeviceMemory>::type, vk::DeviceMemoryDeleter&)’
     return UniqueDeviceMemory( allocateMemory( allocateInfo, allocator ), deleter );

vulkan.hpp: In member function ‘vk::UniqueFence vk::Device::createFenceUnique(const vk::FenceCreateInfo&, vk::Optional<const vk::AllocationCallbacks>) const’:
/home/rayn/Dev/xender/external/include/vulkan/vulkan.hpp:28396:71: error: no matching function for call to ‘vk::UniqueHandle<vk::Fence, vk::FenceDeleter>::UniqueHandle(vk::ResultValueType<vk::Fence>::type, vk::FenceDeleter&)’
     return UniqueFence( createFence( createInfo, allocator ), deleter );

...
```
Looks like the  VULKAN_HPP_NO_EXCEPTIONS in struct ResultValueType has something to do with it?

I'm using vulkan header version 1.0.65, in which the line 887 to 894 is:
```shell
  template <typename T>
  struct ResultValueType
  {
#ifdef VULKAN_HPP_NO_EXCEPTIONS
    typedef ResultValue<T>  type;
#else
    typedef T              type;
#endif
  };
```

mark-lunarg:
+1 for @ro0o0o0okie.

mtavenrath:
@ro0o0o0okie I cannot reproduce the issue with the latest version of Vulkan-Hpp. Can you verify if the issue still exists and if so, tell me which compiler you're using and reopen the issue?

PR #144, opened 2017-11-17
Device::acquireNextImageKHR and other functions return Result<...> in enhanced mode

JulianvDoorn:
Earier today I was coding stuff with the vulkan.hpp wrapper and noticed device::acquireNextImageKHR(Swapchain, Timeout, Semaphore, Fence) returns Result<uint32_t> which forbid me to compile my code.
I noticed this and went though the header file and found out that there seems to be no reason for Device::acquireNextImageKHR, Device::acquireNextImage2KHR and Device::getSwapchainCounterEXT to diverge from the usual pattern to return the created object.

How the call is:
```c++
uint32_t imageIndex = device.acquireNextImageKHR(swapchain, 0, semaphore, fence).value;
```

How the call should be, I think:
```c++
uint32_t imageIndex = device.acquireNextImageKHR(swapchain, 0, semaphore, fence);
```

### DIFF
--- a/vulkan/vulkan.hpp
+++ b/vulkan/vulkan.hpp
@@ -27566,7 +27566,7 @@ namespace VULKAN_HPP_NAMESPACE
 
     Result acquireNextImage2KHX( const AcquireNextImageInfoKHX* pAcquireInfo, uint32_t* pImageIndex ) const;
 #ifndef VULKAN_HPP_DISABLE_ENHANCED_MODE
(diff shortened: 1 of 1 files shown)


mtavenrath:
@JulianvDoorn The functions you're mentioning do have more than one successful result and Vulkan-Hpp may not throw an exception upon a successful return. E.g. vkAcquireNextImageKHR can report one those those values: "VK_SUCCESS,VK_TIMEOUT,VK_NOT_READY,VK_SUBOPTIMAL_KHR". If we drop the result the application wouldn't be able to detect a timeout etc.

PR #158, opened 2017-12-01
Update to Vulkan 1.0.66

mtavenrath:
- Add support for the new alias tag

### DIFF
--- a/VulkanHppGenerator.cpp
+++ b/VulkanHppGenerator.cpp
@@ -754,6 +754,7 @@ struct DependencyData
 {
   enum class Category
   {
+    ALIAS,
     COMMAND,
     ENUM,
     FLAGS,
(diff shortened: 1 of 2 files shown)


VulkanHppGenerator.cpp
```diff
@@ -865,6 +877,7 @@ struct VendorIDData
 
 struct VkData
 {
+  std::map<std::string, AliasData>              aliases; // Enum Aliases
```
dumganhar: Why not to use `std::unordered_map<std::string, xxx>` since it has better performance than `std::map<std::string, xxx>`?

mtavenrath: Both, std::map and std::unordered_map, may have benefits here. std::map is faster with regards to insertion complexity (depends on allocation performance), std::unordered_map is faster with regards to lookup complexity (depends on speed up hash-computation for the string). So choosing the one or the other won't change much.

PR #218, opened 2018-05-15
Added reference to VK², counterpart of Vulkan-Hpp for Kotlin

elect86:


mtavenrath:
This is better added to https://github.com/KhronosGroup/Vulkan-Ecosystem which collects projects in the Vulkan Ecosystem. Vulkan-Hpp should reference only C++ projects using Vulkan-Hpp.

PR #223, opened 2018-06-06
List out default ctor parameters in separate lines

nsubtil:
It's often useful to browse through vulkan.hpp to check what the default values
for various info structs are, but having them on a single line makes it hard to
read.

This change puts each ctor argument on a different line, which makes the
default values more visible.

### DIFF
--- a/VulkanHppGenerator.cpp
+++ b/VulkanHppGenerator.cpp
@@ -3996,11 +3996,14 @@ void VulkanHppGenerator::writeResultEnum(std::ostream & os)
 void VulkanHppGenerator::writeStructConstructor(std::ostream & os, std::string const& name, StructData const& structData, std::map<std::string, std::string> const& defaultValues)
 {
   // the constructor with all the elements as arguments, with defaults
-  os << "    " << name << "( ";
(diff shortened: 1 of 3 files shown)


jherico:
I'm kind of tempted to say that rather than attempt to do more formatting programatically, perhaps the pipeline for building the vulkan.hpp should include a clang-format pass, and the repo should include a .clang-format file for the desired formatting.  The generation code could then be simplified to eliminate the need to output anything that isn't a C++ token or the minimal amount of required whitespace.

asuessenbach:
Hi Nuno,

we like that change, makes reading the constructors substantially easier.
I'd like to ask for two little changes in your pull-request, though:
1. Part of your change is three times the very same code sequence. That calls for a little helper function with some indentationLength (or so) argument.
2. Instead of removing lines 4097-4100, you could call that little helper right there. You would just add that same indentationLength argument to writeStructConstructorArgument().

With those two changes, we'd be happy to merge that.

nsubtil:
Updated PR with the changes requested.

mtavenrath:
I fixed the generation issue in MR #225 and submitted it. Closing this one as duplicate.

PR #227, opened 2018-07-04
Add .get() method to ResultValue

PoignardAzur:
This method would allow someone who has defined `VULKAN_HPP_NO_EXCEPTIONS` to still use exceptions when necessary in their workflow.

Alternative names:

* tryGet
* at

Note that I haven't tested this PR yet (I haven't installed Vulkan on my current computer), but the code is simple enough that I don't expect there to be bugs on testing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -202,7 +202,7 @@ vk::MemoryDedicatedRequirementsKHR &dedMemReqs = result.get<vk::MemoryDedicatedR
 
 ### Return values, Error Codes & Exceptions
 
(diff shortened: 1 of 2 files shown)


asuessenbach:
Thanks for providing a pull request. We appreciate that.
But we don't think, that having such a get-method in ResultValue is worth to have, in general. We especially consider having an exception when exceptions are explicitly turned off not a great idea at all.

PoignardAzur:
Can you detail your reasoning?

Allowing users to not use exceptions in general, but use them in specific cases gives them more granularity in their code. In that regard, a `get()` method would fulfill the same role as the `try` operator in `swift or the `?` operator in Rust.

asuessenbach:
As said before, when I explicitly switch off exceptions, I would not expect to get exceptions any more. In such a case, if I really want to add exceptions under some special circumstances, I'd have to add them on my own.

jherico:
@PoignardAzur bear in mind that some dev organization _disable exception handling_ at the compiler level as a matter of policy.  So the presumption as a project is that if you support disabling exceptions, as Vulkan.hpp allows you to do, then there should be _no_ code that would throw an exception as it would fail to compile in such a circumstance.

PR #231, opened 2018-07-12
Correct return type of create*Unique functions

asuessenbach:


pierremoreau:
Thank you for the quick fix, that solves the issue and the code sample compiles and links properly!

I was wondering, would it be better to define `UniqueDebugReportCallbackEXT` (and similar) as a templated alias instead? With the current change, the alias is no longer used and could therefore be removed.

Issue #233, opened 2018-07-20
Newly added Dispatch member on ObjectDestroy breaks structs containing vk::Unique* objects


genpfault:
fbd17bb272 seems to break structs/classes containing `vk::Unique*` members (minimized from the current [Vookoo](https://github.com/andy-thomason/Vookoo) master):

    #include <vulkan/vulkan.hpp>

    struct Foo
    {
        vk::UniqueShaderModule module;
    };

    int main()
    {
        Foo foo;
        vk::Device device;
        vk::ShaderModuleCreateInfo ci;
        foo.module = device.createShaderModuleUnique(ci);
        return 0;
    }

Errors:

    2>  main.cpp
    2>c:\test\vulkan-hpp\vulkan/vulkan.hpp(392): error C2280: 'vk::ObjectDestroy<vk::Device,Dispatch> &vk::ObjectDestroy<vk::Device,Dispatch>::operator =(const vk::ObjectDestroy<vk::Device,Dispatch> &)': attempting to reference a deleted function
    2>          with
    2>          [
    2>              Dispatch=vk::DispatchLoaderStatic
    2>          ]
    2>  c:\test\vulkan-hpp\vulkan/vulkan.hpp(2366): note: compiler has generated 'vk::ObjectDestroy<vk::Device,Dispatch>::operator =' here
    2>          with
    2>          [
    2>              Dispatch=vk::DispatchLoaderStatic
    2>          ]
    2>  c:\test\vulkan-hpp\vulkan/vulkan.hpp(390): note: while compiling class template member function 'vk::UniqueHandle<vk::ShaderModule,vk::DispatchLoaderStatic> &vk::UniqueHandle<vk::ShaderModule,vk::DispatchLoaderStatic>::operator =(vk::UniqueHandle<vk::ShaderModule,vk::DispatchLoaderStatic> &&)'
    2>  c:\test\main.cpp(13): note: see reference to function template instantiation 'vk::UniqueHandle<vk::ShaderModule,vk::DispatchLoaderStatic> &vk::UniqueHandle<vk::ShaderModule,vk::DispatchLoaderStatic>::operator =(vk::UniqueHandle<vk::ShaderModule,vk::DispatchLoaderStatic> &&)' being compiled
    2>  c:\test\main.cpp(5): note: see reference to class template instantiation 'vk::UniqueHandle<vk::ShaderModule,vk::DispatchLoaderStatic>' being compiled

[...]

PR #236, opened 2018-07-24
Changed m_dispatch to pointer type to fix copy/move of UniqueHandles

lowkey42:
Resolves issue #233

ObjectDestroy, ObjectFree and PoolFree currently store a const reference to the Dispatch objects which inhibits the compiler from generating copy/move operators and a move constructor.

### DIFF
--- a/VulkanHppGenerator.cpp
+++ b/VulkanHppGenerator.cpp
@@ -742,7 +742,7 @@ const std::string deleterClassString = R"(
       ObjectDestroy( OwnerType owner = OwnerType(), Optional<const AllocationCallbacks> allocator = nullptr, Dispatch const &dispatch = Dispatch() )
         : m_owner( owner )
         , m_allocator( allocator )
-        , m_dispatch( dispatch )
+        , m_dispatch( &dispatch )
       {}
 
       OwnerType getOwner() const { return m_owner; }
@@ -752,13 +752,13 @@ const std::string deleterClassString = R"(
       template <typename T>
       void destroy(T t)
       {
-        m_owner.destroy( t, m_allocator, m_dispatch );
+        m_owner.destroy( t, m_allocator, *m_dispatch );
       }
 
     private:
       OwnerType m_owner;
       Optional<const AllocationCallbacks> m_allocator;
-      Dispatch const& m_dispatch;
+      Dispatch const* m_dispatch;
   };
 
   class NoParent;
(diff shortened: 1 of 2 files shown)


mtavenrath:
@lowkey42: Thank you for fixing issue #233. Can you please update the PR with the generated vulkan.hpp?

lowkey42:
After a bit of confusion about the line endings I have added the generated vulkan.hpp. 
But there are still some differences in it that I don't quite understand (e.g. missing vkGetPhysicalDevicePortabilitySubsetDataKHR) although my changes are based on the latest commit (7d264c7c10e193027c74729993c3298ad5b96907) and I'm using the same Vulkan-Docs (6377124f371f9d8b1d9761b1116fac647222a13d).

VulkanHppGenerator.cpp
```diff
@@ -742,7 +742,7 @@ const std::string deleterClassString = R"(
       ObjectDestroy( OwnerType owner = OwnerType(), Optional<const AllocationCallbacks> allocator = nullptr, Dispatch const &dispatch = Dispatch() )
         : m_owner( owner )
         , m_allocator( allocator )
-        , m_dispatch( dispatch )
+        , m_dispatch( &dispatch )
```
pierremoreau: This seems dangerous: aren’t you going to end up with a dangling pointer if the constructor is called without providing a `Dispatch` object, as you would be taking the address of a temporary? Or if the `dispatch` object goes out-of-scope before `ObjectDestroy` does? (For example `dispatch` is a local variable in a class method, and `ObjectDestroy`, indirectly via `UniqueInstance` for example, is an attribute of that same class.)

The existing code has the same issue.

lowkey42: Yes, I noticed that too, but couldn't find an obvious/easy fix, since vk::DispatchLoaderDynamic would be too big to reasonably copy.
The best option I can currently think of would be using a static instance as a default parameter value. e.g.:

```cpp
namespace detail {
	template<class Dispatch>
	const Dispatch& defaultDispatchInstance() {
		static const Dispatch instance;
		return instance;
	}
}

// ...

ObjectFree( OwnerType owner = OwnerType(), Optional<const AllocationCallbacks> allocator = nullptr, Dispatch const &dispatch = detail::defaultDispatchInstance<Dispatch>() )
```

pierremoreau: There might still be some issues for DispatchLoaderDynamic, as you might want to give it a special instance and device.

Another approach, though it’s a bit heavy, would be to change the API to take a shared_ptr to the dispatch object, instead of the reference; that way you are sure it won’t run away while you’re still holding on it.

PR #244, opened 2018-08-24
extra constructor for uniquehandles that take the owner variable dire…

dokipen3d:
…ctly

(had to recreate the pull request because my git username wasn't set correctly when I made the commit so the CLA couldn't be signed)

Hiya. I was looking for a way to create a unique handle and have the deleter be created by handing the object that does the deletion directly. I gave this a go and it seemed to work in the simple case of surface creation and destruction. I think it is a lot cleaner than creating a deleter and passing it in.

ex.

```
VkSurfaceKHR surface;
glfwCreateWindowSurface(instance, window.get(), nullptr, &surface);
vk::UniqueSurfaceKHR(surface, instance);
```

I went and modified the generator to add the required bits and bobs.

Let me know what you think!

### DIFF
--- a/VulkanHppGenerator.cpp
+++ b/VulkanHppGenerator.cpp
@@ -634,10 +634,16 @@ const std::string uniqueHandleHeader = R"(
   class UniqueHandle : public UniqueHandleTraits<Type,Dispatch>::deleter
   {
   private:
+    using Owner = typename UniqueHandleTraits<Type,Dispatch>::owner;
     using Deleter = typename UniqueHandleTraits<Type,Dispatch>::deleter;
   public:
     explicit UniqueHandle( Type const& value = Type(), Deleter const& deleter = Deleter() )
       : Deleter( deleter)
+      , m_value( value )
+    {}
+
+    explicit UniqueHandle( Type const& value = Type(), Owner const& owner = Owner() )
+      : Deleter( owner)
       , m_value( value )
     {}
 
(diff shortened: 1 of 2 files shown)


mtavenrath:
@dokipen3d This is a useful enhancement and we plan to merge it. To resolve the merge conflict in vulkan.hpp it's necessary to regenerate the header after merging the lastest changes. Would you be willing to do this?

dokipen3d:
hi @mtavenrath I can if you want, however, I recently discovered that you can actually pass in an instance to the handle as a second parameter already and it implicitly converts to the deleter. I'll merge the changes any and we can decide where to go from there.

dokipen3d:
So I made the changes (was a little bit fiddly as the conflict was the whole file).

I tested the new header with my hello triangle program (commented out the vk heeader version static assert temporarily).

I had to add delegating constructors to the unique handle class to allow for default construction of uniquehandles (in a std::vector for example). Otherwise with the default parameters, the constructor was ambigious.

To be honest it was working fine without the owner type. But if you still want to merge I'm more than happy if you'l'l take it.

asuessenbach:
Thanks a lot for your contribution. We appreciate such help!
But as you mentioned that this change is not needed at all and it doesn't even ease any usage, we decided to not merge it.
We hope, you're not too disappointed. Stay tuned!

dokipen3d:
Not too disappointed. Just fun to be part of the process and learn a bit along the way. Cheers!

PR #273, opened 2018-11-24
Use std::move when calling createResultValue with NO_EXCEPTIONS

michaelcowan:
Use move semantics when calling `createResultValue` to allow compiling when `T` is a `std::vector<UniqueHandle<>>`.


asuessenbach:
Good catch!
Thanks for your help.

PR #291, opened 2019-02-13
Fix to_string bracket formatting functions on *Flags.

Wunkolo:
This just a fix in relation to 9a4f863d6e8bbaa9a029d985781104e0ffc4158d to have bitflag formatting be consistent with previous versions of `vulkan.hpp`
Right now:
Empty bitmask returns: `{}`
Populated bitmask returns: ` Graphics | Compute | Transfer | SparseBinding`

How it used to be:
Empty bitmask returns: `{}`
Populated bitmask returns: ` {Graphics | Compute | Transfer | SparseBinding}`


mtavenrath:
asussenbach fixed the issue slightly different. Thank you for your input!

PR #292, opened 2019-02-24
Add StructureChain tuple-based accessor

Wunkolo:
At the moment, pulling out individual components of a structure chain involves multiple calls to `get` and creating references to each individual element in the chain:
```cpp
// Makes a structure-chain type
auto DevicePropertyChain
	= SomePhysicalDevice.getProperties2<
		vk::PhysicalDeviceProperties2,
		vk::PhysicalDeviceIDProperties
	>();

// Now I have to pull out individual references
auto& DeviceProperties = DevicePropertyChain.get<vk::PhysicalDeviceProperties2>();
auto& DeviceIDProperties = DevicePropertyChain.get<vk::PhysicalDeviceIDProperties>();
```

A tuple-based accessor is added that returns a tuple of the requested reference types. Also allows for [structured-binding](https://en.cppreference.com/w/cpp/language/structured_binding) based declarations.
```cpp
// Makes a structure-chain type
auto DevicePropertyChain
	= SomePhysicalDevice.getProperties2<
		vk::PhysicalDeviceProperties2,
		vk::PhysicalDeviceIDProperties
	>();

// Structured binding creates referenced types
auto [DeviceProperties, DeviceIDProperties] = DevicePropertyChain.get<vk::PhysicalDeviceProperties2,vk::PhysicalDeviceIDProperties>();
```

### DIFF
--- a/VulkanHppGenerator.cpp
+++ b/VulkanHppGenerator.cpp
@@ -4554,6 +4554,15 @@ int main( int argc, char **argv )
 
     template<typename ClassType> ClassType& get() { return static_cast<ClassType&>(*this);}
 
+    template<typename ClassTypeA, typename ClassTypeB, typename ...ClassTypes>
+    std::tuple<ClassTypeA, ClassTypeB, ClassTypes...> get()
+    {
+        return std::tuple_cat(
+            std::make_tuple(get<ClassTypeA>(),get<ClassTypeB>()),
+            std::make_tuple(get<ClassTypes>()...)
+        );
+    }
+
   private:
     template<typename List, typename X>
     void link()


mtavenrath:
Structured binding is a C++17 feature. This change would force code bases compiled on compilers without C++17 support to use tuples which will break compatibility.

I filed issue to discuss how enable  C++14/C++17 features. Once we got an agreement here this PR can be merged with a guard and a second get function get_tuple with the interface you've implemented.

Wunkolo:
> Structured binding is a C++17 feature.
> This change would force code bases compiled on compilers without C++17 support to use tuples which will break compatibility.

Structured bindings are a C++17 feature yeah, but the code in this PR is all C++**11** features! C++17 just adds the syntax sugar around declaring a value from a `tuple`.

This PR uses `std::tuple_cat` and `std::make_tuple` to handle calls to `get` with two or more template arguments and returns a tuple of the requested types.
```cpp
    template<typename ClassTypeA, typename ClassTypeB, typename ...ClassTypes>
    std::tuple<ClassTypeA, ClassTypeB, ClassTypes...> get()
    {
        return std::tuple_cat(
            std::make_tuple(get<ClassTypeA>(),get<ClassTypeB>()),
            std::make_tuple(get<ClassTypes>()...)
        );
    }
```

The structured binding feature I showed is just one of the benefits of having a tuple-based return value. This would also work with an `std::tie` based assignment as well.

asuessenbach:
You're absolutely right, that's just C++11.
So thanks a lot for your contribution, we very much appreciate that!

Issue #297, opened 2019-03-08
add option use system-wide tinyxml2


sl1pkn07:
Hi

is possible add the option for use system tinyxml2 instead of download the sources?

tinyxml2 have pkgconfig and cmake modules for use in cmake

greetings

sl1pkn07:
~~~patch
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 8f11117..05bfa7b 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -52,19 +52,27 @@ set(SOURCES
   VulkanHppGenerator.cpp
 )
 
-set(TINYXML2_SOURCES
-  tinyxml2/tinyxml2.cpp
-)
-
-set(TINYXML2_HEADERS
-  tinyxml2/tinyxml2.h
-)
+find_package(tinyxml2 REQUIRED)
+if(tinyxml2_FOUND)
+  set(TINYXML2_HEADERS
+    ${tinyxml2_INCLUDE_DIRS}
+  )
+else()
+  set(TINYXML2_HEADERS
+    tinyxml2/tinyxml2.h
+  )
+  set(TINYXML2_SOURCES
+      tinyxml2/tinyxml2.cpp
+  )
+endif()
 
 source_group(headers FILES ${HEADERS})
 source_group(sources FILES ${SOURCES})
 
 source_group(TinyXML2\\headers FILES ${TINYXML2_HEADERS})
-source_group(TinyXML2\\sources FILES ${TINYXML2_SOURCES})
+if(NOT tinyxml2_FOUND)
+  set(TINYXML2_SOURCES
+      tinyxml2/tinyxml2.cpp
+  )
+endif()
 
 source_group(headers FILES ${HEADERS})
 source_group(sources FILES ${SOURCES})
 
 source_group(TinyXML2\\headers FILES ${TINYXML2_HEADERS})
-source_group(TinyXML2\\sources FILES ${TINYXML2_SOURCES})
+if(NOT tinyxml2_FOUND)
+  source_group(TinyXML2\\sources FILES ${TINYXML2_SOURCES})
+endif()
 
 add_executable(VulkanHppGenerator
   ${HEADERS}
@@ -75,7 +83,11 @@ add_executable(VulkanHppGenerator
 
 set_property(TARGET VulkanHppGenerator PROPERTY CXX_STANDARD 11)
 
[...]

PR #299, opened 2019-03-11
Add find_package-based tinyxml2 linkage

Wunkolo:
Will select between either the user-system's pre-existing tinyxml2 library, or utilize
the embedded git library within the repo.
When using the repo-tinyxml2, it will instead use `add_library` to link against the tinyxml2 target. 
Saves the need for a recursive git-clone in favor of linking against the user's already-compiled tinyxml2 library, making it build-ready straight from a `git clone` if the user already has tinyxml2 installed.
Implements suggestions based on #297

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -44,38 +44,24 @@ file(TO_NATIVE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/vulkan/vulkan.hpp vulkan_hpp)
 string(REPLACE "\\" "\\\\" vulkan_hpp ${vulkan_hpp})
 add_definitions(-DVULKAN_HPP_FILE="${vulkan_hpp}")
 
-set(HEADERS
-  VulkanHppGenerator.hpp
-)
-
-set(SOURCES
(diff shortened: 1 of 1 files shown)


mtavenrath:
@Wunkolo The change does not provide a FindTinyXML2.cmake script and cmake does not have support for tinyxml2 by default. To ensure this change works nicely those would have to be added making the change more complex than necessary.

What is the reason to use the system library instead of the one provided as submodule? Given the added complexity I'm not sure if it makes sense to do such a change.

asuessenbach:
This has been resolved by PR #417.

sl1pkn07:
> @Wunkolo The change does not provide a FindTinyXML2.cmake script and cmake does not have support for tinyxml2 by default. To ensure this change works nicely those would have to be added making the change more complex than necessary.
> 

no need FindTinyXML2.cmake because tinyxml2 have include the cmake module for use with cmake

> What is the reason to use the system library instead of the one provided as submodule? Given the added complexity I'm not sure if it makes sense to do such a change.

avoid use static libs

PR #318, opened 2019-04-06
Check for errors before initializing vectors of UniqueHandles

### LINKED ISSUE #317: Device::allocateDescriptorSetsUnique crashes on errors
`Device::allocateDescriptorSetsUnique` doesn't check for errors returned by `vkAllocateDescriptorSets` before pushing elements onto the local `std::vector<UniqueDescriptorSet>`. If an error occurs, it will be checked later (in createResultValue) and converted into an exception. Stack unwinding will then cause this `std::vector` to be destroyed, hence calling `PoolFree::destroy` on uninitialized objects.

From a quick glance it seems this issue also affects `allocateCommandBuffersUnique`.

My suggested fix is to replace

    for ( size_t i=0 ; i<allocateInfo.descriptorSetCount ; i++ )

with

    for ( size_t i=0 ; result == Result::eSuccess && i<allocateInfo.descriptorSetCount ; i++ )

neobrain:
Fixes #317.

Turns out this fixes a couple more instances than I originally anticipated.

### DIFF
--- a/VulkanHppGenerator.cpp
+++ b/VulkanHppGenerator.cpp
@@ -2901,7 +2901,7 @@ ${i}  ${type}* buffer = reinterpret_cast<${type}*>( reinterpret_cast<char*>( ${t
 ${i}  Result result = static_cast<Result>(d.vk${command}( m_device, ${arguments}, reinterpret_cast<Vk${type}*>( buffer ) ) );
 
 ${i}  ${Deleter}<${DeleterTemplate},Dispatch> deleter( *this, ${deleterArg}, d );
-${i}  for ( size_t i=0 ; i<${vectorSize} ; i++ )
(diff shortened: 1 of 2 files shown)


neobrain:
I'm not sure this small change justifies me spending time researching the legal implications of me signing the suggested CLA. I would suggest merging this based on non-copyrightability, though.

asuessenbach:
Thanks a lot for spotting this bug and suggest how to fix it. I've triggered a pull request that resolves that issue as well and will replace yours. I assume, as you don't want to go through the Contributor License Agreement, you're fine with that approach?

mtavenrath:
@asuessenbach implemented a slightly different fix for this issue.

neobrain:
@asuessenbach Yeah, that's fine of course. Thank you!

I'll make sure to review the CLA if I do came up with another patch. Seemed somewhat heavy for this simple change, but I know it's there for good reason :smile:

PR #322, opened 2019-04-16
Don't create destroy functions for alias commands

thananop1987:
........


asuessenbach:
Could you please resolve the conflicts of your pull request?
What do you want to do here, that is no already part of the VulkanHppGenerator?

PR #326, opened 2019-05-08
Add type alias for underlying type of vk::UniqueHandle

JarrettSJohnson:
In the same spirit of `std::unique_ptr::element_type`--want to be able to easily retrieve the type of the class managed by the UniqueHandle.

### DIFF
--- a/VulkanHppGenerator.cpp
+++ b/VulkanHppGenerator.cpp
@@ -4602,6 +4602,9 @@ int main( int argc, char **argv )
   private:
     using Deleter = typename UniqueHandleTraits<Type,Dispatch>::deleter;
   public:
+
+    using element_type = Type;
+
     explicit UniqueHandle( Type const& value = Type(), Deleter const& deleter = Deleter() )
       : Deleter( deleter)
       , m_value( value )
@@ -4684,8 +4687,6 @@ int main( int argc, char **argv )
       std::swap(m_value, rhs.m_value);
       std::swap(static_cast<Deleter&>(*this), static_cast<Deleter&>(rhs));
     }
-
-    using element_type = Type;
 
   private:
     Type    m_value;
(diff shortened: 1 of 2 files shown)


asuessenbach:
That's a great idea! Thanks for contributing to this project.

For completeness, could you add a generated vulkan.hpp to this PR?

And, minor thing: I'd prefer to have that type alias at the beginning of the UniqueHandle class, right above the constructors.

JarrettSJohnson:
Thanks for considering the PR. I'm sure there's probably many use-cases for it. Here's just one that I currently use:

```cpp
template <typename VK_HANDLE_TYPE,
          typename VK_TYPE = typename VK_HANDLE_TYPE::element_type>
std::vector<VK_TYPE>
UniqueToRaw(const std::vector<VK_HANDLE_TYPE>& handles)
{
  std::vector<VK_TYPE> newBuffer(handles.size());
  std::transform(handles.begin(), handles.end(), newBuffer.begin(),
     [](const auto& handle) { return handle.get(); });
  return newBuffer;
}

/*
 * ...
 */
std::vector<vk::UniqueCommandBuffer>& commandBuffer{};
//...
auto commands = UniqueToRaw(commandBuffer);
//...
vk::SubmitInfo submitInfo{};
submitInfo.commandBufferCount = static_cast<std::uint32_t>(commands.size());
submitInfo.pCommandBuffers = commands.data();
```
I'm not sure if this is the optimal way to do it, but couldn't find a better way.

asuessenbach:
Your copy operation looks good.
Would you mind if we'd use a slight variation of it?
```
template <typename UniqueType>
VULKAN_HPP_INLINE std::vector<typename UniqueType::element_type>
uniqueToRaw(std::vector<UniqueType> const& handles)
{
  std::vector<UniqueType::element_type> newBuffer(handles.size());
  std::transform(handles.begin(), handles.end(), newBuffer.begin(),
    [](auto const& handle) { return handle.get(); });
  return newBuffer;
}
```

JarrettSJohnson:
Go right ahead. :)

PR #328, opened 2019-05-09
Support compilation with VK_NO_PROTOTYPES defined

res2k:
Some third-party libraries includes `vulkan.h` with `VK_NO_PROTOTYPES` defined (e.g. Qt).
Currently, to use Vulkan-Hpp with these libraries you need to include the Vulkan and certain Qt headers in the correct order.
This patch avoids this hassle by supporting compilation with VK_NO_PROTOTYPES defined.

This works by adding checks for `VK_NO_PROTOTYPES` and disables code that relies on the prototypes:
`DispatchLoaderStatic` and certain initializations of `DispatchLoader`.

This also means that you _need_ to manually provide a Dispatcher (`DispatchLoaderDynamic` or somethinh custom) for all Vulkan-Hpp calls; but presumably you're using some kind of dynamic dispatching anyway if you haven't got the prototypes.

### DIFF
--- a/VulkanHppGenerator.cpp
+++ b/VulkanHppGenerator.cpp
@@ -2354,6 +2354,7 @@ void VulkanHppGenerator::writeDispatchLoaderDynamic(std::ostream &os)
   public:
     DispatchLoaderDynamic() = default;
 
+#if !defined(VK_NO_PROTOTYPES)
     // This interface is designed to be used for per-device function pointers in combination with a linked vulkan library.
     DispatchLoaderDynamic(vk::Instance const& instance, vk::Device const& device = {})
     {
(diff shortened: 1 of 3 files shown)


mtavenrath:
@res2k Thank you for your contribution. We plan to merge your change, but have to ask you to add the newly generated vulkan.hpp to this PR.

res2k:
> add the newly generated vulkan.hpp to this PR.

Sure, is on branch.

Carandiru0:
This screws up the usage of Volk for me. 
I have simply exposed the DispatchLoaderStatic by removing the new #ifdefs that surrounded it. And deleted the NeedsExplicitDispatchLoader typedef to get it working again
Using latest version of Volk and have predefined VK_NO_PROTOTYPES

DispatchLoaderStatic, in this case, maps to the functions defined in Volk nicely, I do not link with the vulkan.lib static library.

. This regression should be reverted or make the changes as stated above.

Attached changed vulkan.hpp that has me compiling again. Thanks.

[vulkan.zip](https://github.com/KhronosGroup/Vulkan-Hpp/files/3591290/vulkan.zip)

res2k:
Can you be more specific than "screws up"?

res2k:
Okay, I took a look at what Volk does:
1. `#define`s `VK_NO_PROTOTYPES`
2. Includes `vulkan.h`
3. Defines it's own Vulkan "functions" (pointers actually)

So the simplest solution is to `#undef`ine `VK_NO_PROTOTYPES` after Volk made it's own function definitions. Arguably, that would also be a more consistent state of affairs - there _are_ prototype-compatible definitions available, but currently the macro says there aren't.

Anyway, such an `#undef VK_NO_PROTOTYPES` could either happen in Volk itself (you'd need to file a PR there) or in your own code (after including `volk.h`, before `vulkan.hpp`).

PR #330, opened 2019-05-13
Add handling of "comment" to disabled extensions.

### LINKED ISSUE #329: Build failure with 1.1.108 spec update
There's another build issue with VulkanHppGenerator:

> % (cd build && ./VulkanHppGenerator)
> Loading vk.xml from /home/tree/git/Vulkan-Hpp/Vulkan-Docs/xml/vk.xml
> Writing vulkan.hpp to /home/tree/git/Vulkan-Hpp/vulkan/vulkan.hpp
> warning: Unknown element in spec on line: 8093 comment!
> VulkanHppGenerator: /home/tree/git/Vulkan-Hpp/VulkanHppGenerator.cpp:1463: void VulkanHppGenerator::readExtensionDisabledRequire(const tinyxml2::XMLElement*): Assertion `value == "enum"' failed.
> Abort

No clue what's causing this - @mtavenrath @asuessenbach could you take a look at it? We won't be able to update Vulkan-Headers via https://github.com/KhronosGroup/Vulkan-Headers/pull/57 until this is resolved.

asuessenbach:
Fixes #329.

### DIFF
--- a/VulkanHppGenerator.cpp
+++ b/VulkanHppGenerator.cpp
@@ -1426,7 +1426,7 @@ void VulkanHppGenerator::readExtensionDisabledRequire(tinyxml2::XMLElement const
 {
   checkAttributes(getAttributes(element), element->GetLineNum(), {}, {});
   std::vector<tinyxml2::XMLElement const*> children = getChildElements(element);
-  checkElements(children, { "command", "enum", "type" });
+  checkElements(children, { "command", "comment", "enum", "type" });
 
   for (auto child : children)
   {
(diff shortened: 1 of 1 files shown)


VulkanHppGenerator.cpp
```diff
@@ -1426,7 +1426,7 @@ void VulkanHppGenerator::readExtensionDisabledRequire(tinyxml2::XMLElement const
 {
   checkAttributes(getAttributes(element), element->GetLineNum(), {}, {});
   std::vector<tinyxml2::XMLElement const*> children = getChildElements(element);
-  checkElements(children, { "command", "enum", "type" });
+  checkElements(children, { "command", "comment", "enum", "type" });
```
oddhack: ISTM that any addition to the schema is likely to trigger this kind of problem - it appears that you've encoded a snapshot of the schema into the generator code.

asuessenbach: That's right, it's a snapshot of what I've encountered. I don't know how to handle stuff I haven't encountered, yet. So I try to make sure I'm getting notified when something unexpected appears in vk.xml that might need some additional handling.

PR #332, opened 2019-05-24
Use VULKAN_HPP_ASSERT instead of assert

Valakor:
Update `DispatchLoaderDynamic::init(...)` to use the `VULKAN_HPP_ASSERT` macro instead of a plain `assert`.


asuessenbach:
This looks good, thanks for your contribution.
Though, due to some changes unrelated to yours, the test build failed on your PR. That issue should be fixed by now, but your PR is not scheduled for re-building.
To trigger a re-build of your PR, could you please push it again?

PR #337, opened 2019-06-06
Corrected inline if, which should be evaluated before writing output

nyronium:
This PR fixes an expression of the form
`os << someOptional ? x : y`, which should be
`os << (someOptional ? x : y)`.

Otherwise, it discards the result of the inline if due to the higher precedence of the `<<` operator.

While this change could technically change the resulting `vulkan.hpp`, it does in fact not change the current one, as the the faulty path is apparently not taken for the latest xml spec.


asuessenbach:
Thanks a lot for fixing this issue!

PR #345, opened 2019-06-23
Fix #342 (gcc warning for memcpy())

### LINKED ISSUE #342: GCC 9.1.1 warnings
GCC 9.1.1 generates many warnings of the form:

`vulkan.hpp:48313:61:` warning: ‘void* memcpy(void*, const void*, size_t)’ writing to an object of type ‘struct vk::XcbSurfaceCreateInfoKHR’ with no trivial copy-assignment; use copy-assignment or copy-initialization instead [-Wclass-memaccess]
48313 |       memcpy( this, &rhs, sizeof( XcbSurfaceCreateInfoKHR ) );`

res2k:
Avoid the warning by casting destination to void* when memcpy()ing from C structs.

### DIFF
--- a/VulkanHppGenerator.cpp
+++ b/VulkanHppGenerator.cpp
@@ -3765,12 +3765,12 @@ ${initializers}    {${copyOps}}
 ${subStructConstructor}
     ${name}( Vk${name} const & rhs )
     {
-      memcpy( this, &rhs, sizeof( ${name} ) );
+      memcpy( (void*)this, &rhs, sizeof( ${name} ) );
     }
 
     ${name}& operator=( Vk${name} const & rhs )
     {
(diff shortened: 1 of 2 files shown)


res2k:
Thinking about it, a possible alternative to `memcpy()` would be:
`*reinterpret_cast<VkSomething*> (this) = rhs;`

This would be beneficial for structures with 'holes' due to alignment;
`memcpy()` would touch these holes as well (and access potentially
uninitialized data), but an assign doesn't. This is relevant for error
checking tools like valgrind.

On the other hand, maybe that could introduce some aliasing issue I'm missing.

mtavenrath:
We decided to implement the *reinterpret_cast<VkSomething*> (this) = rhs; solution.

PR #346, opened 2019-06-23
Update travis config

res2k:
* Update the Ubuntu dist, to avoid deprecation warning from Travis
* Build with a newer gcc version as well
* Enable more warnings for the mini-test compilation

The last two changes taken together expose #342 on the CI build and may expose future introductions of new warnings.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,4 +45,4 @@ script:
   - echo "    vk::Result result = vk::createInstance(&inst_info, nullptr, &inst);" >> /tmp/test.cpp
   - echo "}" >> /tmp/test.cpp
   - cat /tmp/test.cpp
-  - ${CXX} -std=c++11 -Ivulkan -IVulkan-Headers/include -c /tmp/test.cpp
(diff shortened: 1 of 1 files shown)


MuAlphaOmegaEpsilon:
`dist: xenial` is the most recent distro offered by Travis CI, and already has GCC 5.4.0 embedded into it, so there would be no need to install it back for each run as it would happen in this case. There's a reason why you chose `trusty` instead?

MuAlphaOmegaEpsilon:
Plus I noticed that the current CI doesn't use GCC 8: having to install GCC at each run slows down quite a lot the build process for CI.
For this reason I usually try to stick as much as possible with the compiler already offered by Travis.
I don't know if CI latency is considered a value here though.

res2k:
> There's a reason why you chose `trusty` instead?

As the previous configuration explicitly set `dist: precise`, I was thinking that an older distro might have been chosen to provide some degree of testing with "old" tools. That was my main reasoning behind using `trusty` instead of `xenial`.

Issue #343, opened 2019-06-18
(1.1.108.0): MSVC 32-bit compilation, error C2664


past-due:
```
C:\VulkanSDK\1.1.108.0\Include\vulkan/vulkan.hpp(56908): error C2664: 
'void vk::DispatchLoaderDynamic::init(VkInstance,PFN_vkGetInstanceProcAddr,VkDevice,PFN_vkGetDeviceProcAddr)': 
cannot convert argument 1 from 'const vk::Instance' to 'VkInstance
```

This appears to have been introduced by #300?

(This is MSVC 2017.)

Doozer:
I get the exact same error, but only when i compile for 32 bit. When i compile the same code in a 64 bit setup it does work without a problem. I am using MSVC 2019 commuinty edition.

asuessenbach:
Resolved with PR #344.

PR #347, opened 2019-06-23
Add an appveyor config

res2k:
MSVC is a major platform, so having some automated builds on it can certainly be useful.

For example, this exposes #343.


mtavenrath:
Merging now that appveyor is finally active in this repository. Thanks to @res2k for this change.

Issue #334, opened 2019-06-01
constexpr Flags


neobrain:
In addition to the constexpr-ification in #195 , having the Flags helper class be `constexpr` would be very beneficial. For instance, I've written helper templates to restrict certain functionality to a predefined set of "known" flag combinations. Currently, since Flags isn't constexpr, this kind of metaprogramming requires the use of the C-style Vulkan enumerations.

This is a trivial change for most of the `Flags` API, including the global `operator|` definitions. The exception are the `operator~`s, since they require the use of FlagTraits before the FlagTraits are actually specialized. I'm not sure how hard fixing that would be, but for the time being just constexpr'ing all other operators and the constructors would be fantastic.

What do you think?

ReinUsesLisp:
In reference to #358, I'm also being forced to define expressions that should be constexpr as const, for instance:
```cpp
namespace {

const/*expr*/ auto BufferUsage =
    vk::BufferUsageFlagBits::eVertexBuffer | vk::BufferUsageFlagBits::eIndexBuffer |
    vk::BufferUsageFlagBits::eUniformBuffer | vk::BufferUsageFlagBits::eStorageBuffer;

} // Anonymous namespace
```
It would be nice to have these as constexpr since AFAIK it's currently being calculated at runtime.

asuessenbach:
Thanks for bringing this up again, the comment in #358 somehow slipped through.

PR #358, opened 2019-07-09
Add constexpr to vk::Flags<> functions.

asuessenbach:
Fixes issue #334.


neobrain:
Awesome, cheers!

As far as I can see, the global `operator|`s generated from `bitmaskOperatorsTemplate` can also be marked `VULKAN_HPP_CONSTEXPR`. (Not sure about `operator~` - I think I ran into build issues when I tried that one.)

PR #365, opened 2019-07-25
Fix typo.

orbea:
I think this is supposed to be `__clang__` and not `__clang___` as seen in elsewhere in the repo and in online documentation.


jack40125:
#365

PR #369, opened 2019-08-11
Only emit union constructors and setters if !returnedonly

jekstrand:
This is required to prevent breakage from an upcoming extension.


jekstrand:
More specifically, this is to fix returnedonly unions in VK_KHR_pipeline_executable_properties.

PR #381, opened 2019-09-16
Add trivial vulkan.hpp compile test.

oddhack:
I end up recreating this every time I do a Vulkan-Headers update, to at
least ensure the new vulkan.hpp compiles, so it would be convenient to
just have it in the repo.

### DIFF
--- a/tests/Vulkan-Headers/hpptest.cpp
+++ b/tests/Vulkan-Headers/hpptest.cpp
@@ -1,0 +1,30 @@
+/*
+ * Copyright (c) 2019 The Khronos Group Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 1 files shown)


asuessenbach:
And using just any of the test projects (like samples/01_InitInstance or tests/DeviceFunctions) would not serve that purpose?

oddhack:
I don't know anything about the project other than the boilerplate required to build the headers, and frankly don't want to spend any time learning. If there's a trivial recipe I can use to build a trivial sample that has no platform dependencies, that would suffice.

asuessenbach:
Are you using some pre-build VulkanHppGenerator to generate your vulkan.hpp? Or are you building it on your own?
In the latter case, just check the cmake options SAMPLES_BUILD and/or TESTS_BUILD, and SAMPLES_BUILD_WITH_LOCAL_VULKAN_HPP and/or TESTS_BUILD_WITH_LOCAL_VULKAN_HPP, respectively. Then you've got a number of more or less meaningfull projects you can build. For example samples/01_InitInstance, looks very similar to your test case.

PR #392, opened 2019-09-25
Change (most of) the samples to use the DispatchLoaderDynamic by default

asuessenbach:
The utility function vk::su::createInstance either creates a default DispatchLoaderDynamic, or prepares usage of DebugUtilsMessenger with a DispatchLoaderStatic as the default.

### DIFF
--- a/samples/EnableValidationWithCallback/CMakeLists.txt
+++ b/samples/EnableValidationWithCallback/CMakeLists.txt
@@ -26,6 +26,8 @@ set(SOURCES
 source_group(headers FILES ${HEADERS})
 source_group(sources FILES ${SOURCES})
 
+add_definitions(-DVULKAN_HPP_DISPATCH_LOADER_DYNAMIC=0)
+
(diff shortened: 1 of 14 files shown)


samples/EnableValidationWithCallback/CMakeLists.txt
```diff
@@ -26,6 +26,8 @@ set(SOURCES
 source_group(headers FILES ${HEADERS})
 source_group(sources FILES ${SOURCES})
 
+add_definitions(-DVULKAN_HPP_DISPATCH_LOADER_DYNAMIC=0)
```
mtavenrath: >add_definitions(-DVULKAN_HPP_DISPATCH_LOADER_DYNAMIC=0)
Those definitions are not required.

asuessenbach: They are needed, as the samples/CMakeLists.txt adds that definition with a 1. There are a few samples that don't need the dynamic DispatchLoader.

PR #405, opened 2019-10-09
Disable CTAD on ArrayProxy's constructors

mocabe:
When you use `vk::ArrayProxy` with C++17 Class Template Argument Deduction(CTAD), template parameter `T` always deduced into argument type of constructor because `ArrayProxy(T & ptr)` always produces best match than other constructors for `std::array` or `std::vector`.  
```cpp
std::array<int, 42> array;
vk::ArrayProxy prox = array;  // vk::ArrayProxy<std::array<int, 42>>
prox.size(); // 1!
```
This is very counter-intuitive and potentially becomes a source of nasty bugs.  
This PR disables CTAD on `ArrayProxy(T &)` by changing it to `ArrayProxy(typename std::remove_reference<T>::type &)` so compilers no longer be able to deduce `T` from argument.  
I'll probably do the same on `ArrayProxy(std::initializer_list<T> const&)` too.

In the future, we can add deduction guides to enable CTAD with C++17 support.

### DIFF
--- a/VulkanHppGenerator.cpp
+++ b/VulkanHppGenerator.cpp
@@ -4383,7 +4383,7 @@ int main( int argc, char **argv )
       , m_ptr(nullptr)
     {}
 
-    ArrayProxy(T & ptr) VULKAN_HPP_NOEXCEPT
+    ArrayProxy(typename std::remove_reference<T>::type & ptr) VULKAN_HPP_NOEXCEPT
       : m_count(1)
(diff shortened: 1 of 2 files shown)


mtavenrath:
@mocabe I haven't found any case where this PR will break existing code, so your MR is approved. Can you please rebase your PR so that we can merge it?

mocabe:
@mtavenrath Thanks, rebased.

PR #417, opened 2019-10-29
Allow setting tinyxml2 source directory

dj2:
This CL adds a build flag to allow the tinyxml2 source directory to
exist outside the Vulkan-HPP source tree.


CMakeLists.txt
```diff
@@ -52,12 +52,16 @@ set(SOURCES
   VulkanHppGenerator.cpp
 )
 
+if (NOT DEFINED VULKAN_HPP_TINYXML2_SRC_DIR)
+  set(VULKAN_HPP_TINYXML2_SRC_DIR ${CMAKE_CURRENT_SOURCE_DIR}/tinyxml2)
```
mtavenrath: Please add "" around the path to ensure that paths with whitespaces won't fail.

PR #422, opened 2019-11-04
Implement support for unicode Windows projects using `vk::DynamicLoader`.

Warpten:
I wasn't really sure what to do with this. Windows is just annoying to work with, and I just hit this on a test project at work.

The other solution is to just force calling `LoadLibraryA`. Both works, but I don't know which one is better, if any.

This macro is defined by `winnt.h`, so I don't see it getting removed anytime soon.


mtavenrath:
According to https://docs.microsoft.com/en-us/windows/win32/learnwin32/working-with-strings on should always prefer the unicode version over the ansi version if possible. Besides this I'm fine with the TEXT macro which ensures that the encoding is correct if there will ever be the need for some new encoding.

Warpten:
Not sure with AppVeyor is failing.

asuessenbach:
Don't worry, we're working on a fix for the AppVeyor stuff.

PR #424, opened 2019-11-05
Add CMakeList target to run generator

dj2:
This CL adds the VULKAN_HPP_RUN_GENERATOR option which will run the
vulkan hpp generator from CMake if enabled.


dj2:
The appveyor failure seems unrelated to my CL?

asuessenbach:
Don't worry, we're working on a fix for the AppVeyor stuff.

PR #429, opened 2019-11-06
Add option to install vulkan.hpp header

dj2:


asuessenbach:
Where does that VULKAN_HPP_INSTALL come from?

dj2:
It would either be set on the command line with `-DVULKAN_HPP_INSTALL=on` or it can be set in the embedders CMakefile. I turned it off by default because I assumed folks wouldn't want to have this auto-install by default.

dj2:
ping

PR #465, opened 2019-12-28
Fixed vertexBindingDescription scope.

nvaara:
vk::PipelineVertexInputStateCreateInfo::pVertexBindingDescriptions used to run out of scope before being used in the Device::createGraphicsPipelineUnique resulting in undefined behavior.


asuessenbach:
Good catch!
Thanks a lot for this little fix.

PR #466, opened 2019-12-30
Update EnableValidationWithCallback.cpp

Ybalrid:
Quick patch that fixes the `EnableValidationWithCallback` sample:
 -  There's no `str()` method here, should use `c_str()`;  
 -  *but* also, you can print the std::string directly to `std::cout`.

Patches remove `.str()` from printing line when WIN32 is not defined.


asuessenbach:
Good catch!
Thanks a lot for this little fix.

PR #485, opened 2020-01-15
Fix regression in ObjectDestroy default argument

W4RH4WK:
It seems like the default `dispatch` argument should be `VULKAN_HPP_DEFAULT_DISPATCHER` rather than default constructing a new dispatcher.

Default constructing a new dispatcher probably works fine when using static dispatch, but fails with dynamic dispatch since the dynamic dispatcher uses a `VkInstance` (and a `VkDevice`) which the newly default constructed dispatcher lacks.


asuessenbach:
Thanks a lot for spotting this issue. And for your contribution.

Issue #463, opened 2019-12-20
GLFW is there but no Linux samples available


filipwasil:
Hello,

Is there specific reason why Linux build is "Unhandled platform" ? Seems like an easy thing to do. GLFW is already there.

mtavenrath:
The only reason for missing Linux samples is the lack of time. PRs which add glfw samples are welcome.

mocabe:
Please check out my PR (#491) and see if it works on your machines :)

PR #491, opened 2020-01-18
Building samples for Linux

mocabe:
This PR tries to support building samples on Linux with minimal changes.

related issues:
#463   _"Seems like an easy thing to do."_

checklists:
- [x] Replace window handling with GLFW
- [x] CMake support for building samples on Linux
- [x] Fixing broken samples
- [x] Testing on both Windows and Linux
  - Just on my machines, though.

known issues:
- [x] Missing deleter for `vk::UniqueSurfaceKHR`
  - Should work now.
- [x] SurfaceCapabilities sample doesn't compile since `SurfaceCapabilitiesFullScreenExclusiveEX` is only supported on Win32.
  - Disabled on non-WIN32 platform
- [x] Raytracing sample is currently broken
  - Fixed, ~~but now it doesn't compile with latest header due to `vk::Flags`'s breaking changes~~
- [x] Some tests missing initialization for dynamic dispatcher, resulting segfault.
  - [x] EnableValidationWithCallback.cpp
- [x] PhysicalDeviceQueueFamilyProperties.cpp is broken
  - [x] Already fixed on master (#432)
- [x] Currently it only works with XCB surfaces. Not sure we can support more platforms.


mocabe:
A bit rough but it looks working on my machines now.
Please test and provide feedback.

@asuessenbach @mtavenrath

asuessenbach:
Thanks a lot for your contribution. We really appreciate it.

mocabe:
> At least needs to check for Linux to use xcb: (CMAKE_SYSTEM_NAME MATCHES "Linux")

Yes. Added check.
Also, now you should be able to build samples with MinGW toolchain.

samples/PushDescriptors/CMakeLists.txt
```diff
@@ -40,6 +32,5 @@ add_executable(PushDescriptors
 )
 
 set_target_properties(PushDescriptors PROPERTIES FOLDER "Samples")
-target_compile_definitions(PushDescriptors PUBLIC VULKAN_HPP_DISPATCH_LOADER_DYNAMIC=1)
-target_include_directories(PushDescriptors PUBLIC ${CMAKE_SOURCE_DIR}/glslang)
-target_link_libraries(PushDescriptors PUBLIC glslang SPIRV)
+target_link_libraries(PushDescriptors PRIVATE utils)
+target_link_libraries(PushDescriptors PRIVATE "${Vulkan_LIBRARIES}")
```
asuessenbach: Why are the Vulkan_LIBRARIES needed here?

samples/RayTracing/CameraManipulator.hpp
```diff
@@ -29,7 +29,7 @@ namespace vk
         enum class Mode { Examine, Fly, Walk, Trackball };
         enum class MouseButton { None, Left, Middle, Right };
         enum class ModifierFlagBits { Shift = 1, Ctrl = 2, Alt = 4 };
-        using ModifierFlags = vk::Flags<ModifierFlagBits, uint32_t, ModifierFlagBits::Shift>;
+        using ModifierFlags = vk::Flags<ModifierFlagBits, uint32_t>;
```
asuessenbach: vk::Flags need the default initialization value as the third template parameter now.

samples/SurfaceCapabilities/SurfaceCapabilities.cpp
```diff
@@ -116,7 +116,7 @@ int main(int /*argc*/, char ** /*argv*/)
         {
           vk::SurfaceProtectedCapabilitiesKHR surfaceProtectedCapabilities = surfaceCapabilities2.get<vk::SurfaceProtectedCapabilitiesKHR>();
           std::cout << "\tSurfaceProtectedCapabilitiesKHR:\n";
-          std::cout << "\t\t" << "setSupportsProtected  = " << static_cast<bool>(surfaceProtectedCapabilities.setSupportsProtected) << "\n";
+          std::cout << "\t\t" << "supportsProtected  = " << static_cast<bool>(surfaceProtectedCapabilities.supportsProtected) << "\n";
```
asuessenbach: Good catch !

samples/RayTracing/CameraManipulator.hpp
mocabe: But doesn't that break compilation for old Vulkan SDK users?
I'd like to ask reasons for breaking change, rather than doing like this:
```cpp
  template <typename BitType, typename MaskType = VkFlags, BitType defaultValue = BitType()>
  class Flags
```

samples/PushDescriptors/CMakeLists.txt
mocabe: Seems mistake. I'll look into it.

samples/utils/utils.hpp
```diff
@@ -16,6 +16,10 @@
 //
 
 #include "vulkan/vulkan.hpp"
+
+#define GLFW_INCLUDE_NONE
+#include <GLFW/glfw3.h>
```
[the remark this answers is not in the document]
mocabe: `glfw` target should already have `PUBLIC` include directory property, which propagates samples through `utils` target, if I'm correct.

samples/utils/utils.hpp
```diff
@@ -29,6 +29,57 @@ namespace vk
...
+      copyToDevice<T>(device, memory, &data, 1);
+    }
+
+    template<class T>
+    VULKAN_HPP_INLINE constexpr const T& clamp(const T& v, const T& lo, const T& hi)
+    {
+      return v < lo ? lo : hi < v ? hi : v;
+    }
+
+    void setImageLayout(vk::UniqueCommandBuffer const &commandBuffer, vk::Image image, vk::Format format, vk::ImageLayout oldImageLayout, vk::ImageLayout newImageLayout);
```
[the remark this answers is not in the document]
mocabe: That is possible, but will require adding declarations for classes defined in the header since some functions use them as parameters.

samples/SurfaceCapabilities/SurfaceCapabilities.cpp
mocabe: I rather surprised that MSVC accepted that code...

samples/utils/utils.hpp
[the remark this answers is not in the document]
asuessenbach: Ah, you're right. It was just a couple of projects that didn't got your changes on utils... because they're not yet in.
Sorry about that.

samples/utils/utils.hpp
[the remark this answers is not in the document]
asuessenbach: ok, then I think it's not worth the work.

samples/RayTracing/CameraManipulator.hpp
asuessenbach: That will break, but only for those who introduce their own Flags.
But you're right, and it's easy to prevent the breaking, so I'll add the default for the defaultValue.

samples/CMakeLists.txt
```diff
@@ -23,7 +23,7 @@ option (SAMPLES_BUILD_WITH_LOCAL_VULKAN_HPP OFF)
 if (CMAKE_SYSTEM_NAME MATCHES "Windows")
   add_definitions(-DNOMINMAX -DVK_USE_PLATFORM_WIN32_KHR)
 else()
-  message(FATAL_ERROR, "unhandled platform !")
+  add_definitions(-DVK_USE_PLATFORM_XCB_KHR)
```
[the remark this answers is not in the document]
mocabe: Please check 7f81075

samples/utils/CMakeLists.txt
```diff
@@ -0,0 +1,26 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+add_library(utils 
+  math.cpp
+  shaders.cpp
+  utils.cpp
+)
+
```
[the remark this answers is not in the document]
mocabe: Added source group for utils in 865d93f

samples/PushDescriptors/CMakeLists.txt
mocabe: Looks like following line causes linker error without `Vulkan_LIBRARIES`
```   
 commandBuffer->pushDescriptorSetKHR(vk::PipelineBindPoint::eGraphics, *pipelineLayout, 0, { 2, writeDescriptorSets }, vk::DispatchLoaderDynamic(*instance, *device));
```


It calls `vk::DispatchLoaderDynamic(*instance, *device)` which ends up accessing `::vkGetInstanceProcAddr` and `::vkGetDeviceProcAddr `.

mocabe:
I think most issues are fixed now, so PR is complete.

samples/PushDescriptors/CMakeLists.txt
asuessenbach: Seems, there's an error in that call: one should not use that temporary vk::DispatchLoaderDynamic. That has somehow survived my changes on using the DispatchLoaderDynamic as the default one.
Could you please make sure that you still need Vulkan_LIBRARIES, if you remove that function argument?

mocabe: Sure. Fixed in e0a026a.

PR #495, opened 2020-01-28
Change return of StructureChain::get() from tuple of values to tuple of references

### LINKED ISSUE #494: StructureChain: almost unusable and inconsistent overloading of get
Variadic template version of `vk::StructureChain::get` return a tuple of values, whereas single parameter version return reference. It is expected, that variadic template version of `get` return a tuple of references.

asuessenbach:
Resolves #494

### DIFF
--- a/VulkanHppGenerator.cpp
+++ b/VulkanHppGenerator.cpp
@@ -5049,12 +5049,9 @@ int main(int argc, char **argv)
     template<typename ClassType> ClassType& get() VULKAN_HPP_NOEXCEPT { return static_cast<ClassType&>(*this);}
 
     template<typename ClassTypeA, typename ClassTypeB, typename ...ClassTypes>
-    std::tuple<ClassTypeA, ClassTypeB, ClassTypes...> get()
+    std::tuple<ClassTypeA&, ClassTypeB&, ClassTypes&...> get()
     {
-        return std::tuple_cat(
-            std::make_tuple(get<ClassTypeA>(),get<ClassTypeB>()),
-            std::make_tuple(get<ClassTypes>()...)
-        );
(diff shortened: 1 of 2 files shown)


VulkanHppGenerator.cpp
```diff
@@ -5049,12 +5049,9 @@ int main(int argc, char **argv)
     template<typename ClassType> ClassType& get() VULKAN_HPP_NOEXCEPT { return static_cast<ClassType&>(*this);}
 
     template<typename ClassTypeA, typename ClassTypeB, typename ...ClassTypes>
```
mocabe: Shouldn't `ClassTypeB` also be removed from template parameter?

asuessenbach: Thanks for catching this!
But we have to keep ClassTypeB (and use it, of course) to help the compiler to distinguish the different flavors of get().

VulkanHppGenerator.cpp
```diff
@@ -5049,12 +5049,9 @@ int main(int argc, char **argv)
...
-    std::tuple<ClassTypeA, ClassTypeB, ClassTypes...> get()
+    std::tuple<ClassTypeA&, ClassTypeB&, ClassTypes&...> get()
     {
-        return std::tuple_cat(
-            std::make_tuple(get<ClassTypeA>(),get<ClassTypeB>()),
-            std::make_tuple(get<ClassTypes>()...)
-        );
+      return std::forward_as_tuple(get<ClassTypeA>(), get<ClassTypeB>(), get<ClassTypes>()...);
```
tomilov: I think `std::tie` is more suitable for this. [Example](https://wandbox.org/permlink/UpakCDPb5DJmxvMS).

PR #498, opened 2020-01-30
use VULKAN_HPP_ASSERT instead of raw assert in generated code

tomilov:
I also suggest to define `VULKAN_HPP_ASSERT` and use it in both `VulkanHppGenerator.cpp` and `VulkanHppGenerator.hpp`, regardless of whether it is generated code or not. It would be less error prone.


asuessenbach:
Thanks for spotting that issue.
Would you mind addig the generated vulkan.hpp to this PR as well?

PR #499, opened 2020-01-30
Add missing initialization of dynamic dispatcher in NoExceptions test

tomilov:


tests/NoExceptions/NoExceptions.cpp
```diff
@@ -48,6 +54,7 @@ int main(int /*argc*/, char ** /*argv*/)
   float queuePriority = 0.0f;
   vk::DeviceQueueCreateInfo deviceQueueCreateInfo(vk::DeviceQueueCreateFlags(), static_cast<uint32_t>(graphicsQueueFamilyIndex), 1, &queuePriority);
   vk::UniqueDevice device = physicalDevices[0].createDeviceUnique(vk::DeviceCreateInfo(vk::DeviceCreateFlags(), 1, &deviceQueueCreateInfo)).value;
+  VULKAN_HPP_DEFAULT_DISPATCHER.init(*device);
```
tomilov: Really not needed, because below "device" functions already resolved during `VULKAN_HPP_DEFAULT_DISPATCHER.init(*instance);`.

asuessenbach:
Thanks for spotting this crash.
But we'd prefer to have this test correctly use the DispatchLoaderStatic. I'll file a PR accordingly.

tests/NoExceptions/NoExceptions.cpp
asuessenbach: You're right, that's not really needed. But it doesnt hurt either.

tomilov:
It is my first experience with pull requests on github. I do something wrong (seems I have to be wiser and use branches). Now I try to recover this pull request.

mocabe:
Don't really recommend using your `master` for PR. Would be better to use separate branch as you said.

PR #500, opened 2020-01-30
Add/Reshape copy constructors for structs and unions.

### LINKED ISSUE #493: vk::SurfaceFormatKHR and vk::SurfaceFormat2KHR do not have constexpr constructors
Like the title says. Structs `vk::SurfaceFormatKHR`, defined at line #57365, and `vk::SurfaceFormat2KHR`, defined at line #57412, do not have constexpr constructors.

asuessenbach:
Allows usage of constexpr on those constructors.

Resolves #493


vulkan/vulkan.hpp
```diff
@@ -2868,13 +2868,20 @@ namespace VULKAN_HPP_NAMESPACE
         dst[I-1] = src[I-1];
         ConstExpression1DArrayCopy<T, N, I - 1>::copy(dst, src);
       }
+
+      VULKAN_HPP_CONSTEXPR_14 static void copy(T dst[N], const T src[N]) VULKAN_HPP_NOEXCEPT
+      {
+        dst[I - 1] = src[I - 1];
+        ConstExpression1DArrayCopy<T, N, I - 1>::copy(dst, src);
```
tomilov: Can't you use for loop? Don't think reverse copy is the best way to copy on every platform. Also multiple template instantiations slowdown compilation time.

VulkanHppGenerator.cpp
```diff
@@ -2614,6 +2614,24 @@ void VulkanHppGenerator::appendStruct(std::string & str, std::pair<std::string,
...
+{
+  // we need an assignment operator if there is constant sType in this struct
+  if ((nonConstSTypeStructs.find(structData.first) == nonConstSTypeStructs.end()) && !structData.second.members.empty() && (structData.second.members.front().name == "sType"))
+  {
+    assert((2 <= structData.second.members.size()) && (structData.second.members[1].name == "pNext"));
+
+    static const std::string stringTemplate = R"(
+${prefix}${structName} & operator=( ${structName} const & rhs ) VULKAN_HPP_NOEXCEPT
+${prefix}{
+${prefix}  memcpy( &pNext, &rhs.pNext, sizeof( ${structName} ) - offsetof( ${structName}, pNext ) );
```
tomilov: Do you really want to skip `sizeof(void *)` bytes at the end of every struct with the `operator =`? Seems test `DeviceFunctions` crashed not only because `VK_NV_RAY_TRACING_EXTENSION_NAME` is missing among `vk::DeviceCreateInfo::ppEnabledExtensionNames`, but also due to this wrong copy constructor.
I would prefer to save `pNext` then `memcpy` whole `rhs` into `*this` then restore original `pNext` of `*this`.

vulkan/vulkan.hpp
asuessenbach: Unfortunately, there's no `for constexpr` (yet). So, to have that copy as a constexpr (with C++14), we have to got that route.

VulkanHppGenerator.cpp
tomilov: Even better is to generate member-wise copy assignment operator. To be consistent with how arranged copy constructor.

asuessenbach: This copy assignment operator copies everything but the constant member sType at the beginning of the structure. I don't see how there should be something missing at the end.

tomilov: I wrongly thought, that `pNext` is the first member, sorry.

vulkan/vulkan.hpp
tomilov: You don't need `for constexpr`. You just can use `for` in `constexpr` (special)function.

tomilov: If can't, you'll get compile error.

mocabe: This kind of recursive loop unrolling works well with small sized arrays, but copying large buffer with this will probably blow up code size and decrease performance. Should be discussed on different issue or PR I guess.

PR #529, opened 2020-03-02
Add const versions of StructureChain::get

mbarriault:
Simple addition, does what it says on the tin.


mbarriault:
Somehow author email got changed to wrong version, closing to re-submit.

PR #548, opened 2020-03-19
Enable samples on AppVeyor builds

mocabe:
This PR enables samples on AppVeyor builds.
Following configurations are included:
- Visual Studio images
  + `Visual Studio 2015`
  + `Visual Studio 2017`
  + `Visual Studio 2019`
- Arch options:
  + `Win32`
  + `x64`  

All builds are compiled with `Debug`.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,40 +1,32 @@
-os:
+image:
+  - Visual Studio 2015 
   - Visual Studio 2017
-
-build:
-  verbosity: detailed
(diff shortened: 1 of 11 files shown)


mocabe:
> Why are those explicit casts needed?

`VULKAN_HPP_TYPESAFE_CONVERSION` is only defined in `x64`, so for `Win32` builds it needs to be explicit to cast from/to C handle types.

PR #549, opened 2020-03-21
Add proper typedefs to vk::Flags

mocabe:
Current `vk::Flags` has `MaskType` typedef in public interface, which is not ideal in terms of naming convention.  
This PR renames `MaskType` to `maks_type`, and also adds `bit_type` for convenience.

### DIFF
--- a/VulkanHppGenerator.cpp
+++ b/VulkanHppGenerator.cpp
@@ -5650,7 +5650,10 @@ int main(int argc, char **argv)
   class Flags
   {
   public:
-    using MaskType = typename std::underlying_type<BitType>::type;
+    // flag bit type
+    using bit_type  = BitType;
+    // underlying type
+    using mask_type = typename std::underlying_type<BitType>::type;
 
     // constructors
     VULKAN_HPP_CONSTEXPR Flags() VULKAN_HPP_NOEXCEPT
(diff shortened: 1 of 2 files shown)


asuessenbach:
What naming convention are you referring to here? Why should we change from MaskType to mask_type?

mocabe:
I'm not reffering any of particular naming convention here. I just wanted to say typedefs in classes are usually `snake_case`, like `UniqueHandle::element_type`.   
Both addition of `mask_type` and `bit_type` is still optional; we can just hide `MaskType` into `private` not to be used from users, but I thought it would be better to provide proper typedefs for users in case they wanted to use `vk::Flags` for generic way.

asuessenbach:
Well, I think, it's the other way round: UniqueHandle::element_type is the only snake_case typedef in vulkan.hpp.

mocabe:
> Well, I think, it's the other way round: UniqueHandle::element_type is the only snake_case typedef in vulkan.hpp.

You're right. Internal wise, there're some non `UpperCamelCase` typenames in some places though.  
I had some oversight on naming stuff, so I'll close this. 
I'm not sure why that `element_type` is there, maybe for `std::pointer_traits` or something?

PR #550, opened 2020-03-23
Introduce ArrayWrapper?D, replacing ConstExpression?DArrayCopy

### LINKED ISSUE #546: Perfectly trivially copyable types are not trivially copyable
I'm on commit `74556a1` from `Vulkan-Headers`.

Recent headers started adding a defined constructor for trivial Vulkan types:
```cpp
  struct DescriptorImageInfo
  {
    VULKAN_HPP_CONSTEXPR DescriptorImageInfo( DescriptorImageInfo const& rhs ) VULKAN_HPP_NOEXCEPT
      : sampler( rhs.sampler )
      , imageView( rhs.imageView )
      , imageLayout( rhs.imageLayout )
    {}
```
This immediately makes the type non-trivially copyable, making it impossible to put on unions and making its `std::memcpy` calls undefined behaviour (something that is done in a copy assignment method from the same class).

These constructors don't have to be defined, C++ takes care of them when no destructor is defined, as it can be seen [here](https://godbolt.org/z/-3mUMp).

asuessenbach:
+ no explicit copy constructor of structs needed any more
+ removed assignment operator of structs that don't hold a const sType member (for those, we still need the assignment operator!)
+ simplified operator==() of structs (pre C++20)

Resolves #546


mocabe:
Note: most of operators are marked as `VULKAN_HPP_CONSTEXPR` but won't be actually usable from `cosntexpr` context since: 
1. use of C functions like `strncmp`
2. use of `std::array`, which is not very `constexpr` friendly until C++17/20

asuessenbach:
You're right. Removed the constexpr from all those relational operators on ArrayWrapper1D<char,N>

PR #555, opened 2020-03-27
Replace vector capacity logic with size logic.

anon-ord:
Vector allocations were using std::vector::reserve() followed by
references to std::vector::data() and offsets from
std::vector::data().

### DIFF
--- a/VulkanHppGenerator.cpp
+++ b/VulkanHppGenerator.cpp
@@ -1978,8 +1978,7 @@ void VulkanHppGenerator::appendFunctionBodyEnhancedVectorOfUniqueHandles(std::st
 {
   std::string const stringTemplate =
     R"(${i}  static_assert( sizeof( ${type} ) <= sizeof( UniqueHandle<${type}, Dispatch> ), "${type} is greater than UniqueHandle<${type}, Dispatch>!" );
-${i}  std::vector<UniqueHandle<${type}, Dispatch>, Allocator> ${typeVariable}s${allocator};
-${i}  ${typeVariable}s.reserve( ${vectorSize} );
+${i}  std::vector<UniqueHandle<${type}, Dispatch>, Allocator> ${typeVariable}s(${vectorSize}${allocator});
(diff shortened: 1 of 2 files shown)


asuessenbach:
Have you encountered any issues with that code?
Actually, we reserve some memory, use that memory as a scratch area, and then push_back stuff into the vector.

anon-ord:
Hi Andreas,

Yeah, we have encountered problems with that code.  It fails our asan
testing.  This is an example of the logic we're concerned about:

 commandBuffers.reserve( allocateInfo.commandBufferCount );

 CommandBuffer* buffer = reinterpret_cast<CommandBuffer*>(
reinterpret_cast<char*>( commandBuffers.data() ) +
allocateInfo.commandBufferCount * ( sizeof( UniqueHandle<CommandBuffer,
Dispatch> ) - sizeof( CommandBuffer ) ) );

 Result result = static_cast<Result>(d.vkAllocateCommandBuffers( m_device,
reinterpret_cast<const VkCommandBufferAllocateInfo*>( &allocateInfo ),
reinterpret_cast<VkCommandBuffer*>( buffer ) ) );


This makes sense to us because reserve doesn't modify the size of the
container, only its capacity and the hpp code relies on the .data() method
(plus offsets) which has strict size requirements.

I've seen this language in a few places, a quick search yields this on
https://en.cppreference.com/w/cpp/container/vector/data :

"Returns pointer to the underlying array serving as element storage. The
pointer is such that range [data(); data() + size()) is always a valid
range, even if the container is empty (data() is not dereferenceable in
that case)."

and https://en.cppreference.com/w/cpp/container/vector/reserve:

"reserve() does not change the size of the vector."

The vkAllocateCommandBuffers() call in the code example above requires that
the memory is allocated (i.e. (data() + offset is valid)) and reserve() doesn't
guarantee that.

Sizing guarantees the size is valid so we're suggesting this change to
remedy.

Cheers,

John

asuessenbach:
Hi John,

thanks for your detailed explanation. Especially this little note in https://en.cppreference.com/w/cpp/container/vector/data made it evident to change the handling:
> If size() is 0, data() may or may not return a null pointer.

But as your proposed change is not quite right (you resize the vector, and then push_back on it), I've created my own PR #560 to resolve that issue.

I hope, you're fine with that.

Stay safe,

    Andreas

anon-ord:
That's all good Andreas.  Thanks.

PR #566, opened 2020-04-16
Add type traits for translating between stdint types and vk::IndexType

jackoalan:
Adds vk::index_type and vk::cpp_index_type templates for translating between stdint.h types and vk::IndexType. This may be useful for template users needing to call bindIndexBuffer without having to write their own traits.

### DIFF
--- a/VulkanHppGenerator.cpp
+++ b/VulkanHppGenerator.cpp
@@ -845,6 +845,70 @@ void VulkanHppGenerator::appendBaseTypes(std::string & str) const
   }
 }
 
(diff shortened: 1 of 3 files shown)


asuessenbach:
Hi Jack,

thanks a lot for your PR. We think, it's worth to have such type traits.
But we preferred to keep them with namings consistent to the other stuff we have, so we decided to file an own PR that does most of what you're suggesting, plus some more (-> #576).
We hope, you're fine with closing your PR then.

Kind regards,

   Andreas

jackoalan:
This looks good! I actually prefer the adjusted convention.

Cheers!

PR #573, opened 2020-04-20
Adjust structure and union constructors and setter functions to correctly handle sType of a member of that structure/union.

### LINKED ISSUE #571: AccelerationStructureGeometryDataKHR doesn't assign a valid sType to it's union members
When constructing or assigning any of the union values to an AccelerationStructureGeometryDataKHR, the sType field of the member doesn't get set.  The three constructors to AccelerationStructureGeometryDataKHR each do an assignment to the union member, but the operator= overload for AccelerationStructureGeometry(Triangles/Aabbs/Instances)DataKHR don't copy the sType (since it's marked const, I suppose).  However, because it's a union, the sType is shared between all three of them, and is never actually set because the real type of the union isn't known before the assignment.

This can be technically solved by using member initialization in the case of the AccelerationStructureGeometryDataKHR constructors, but it doesn't really help in the case of assigning the geometry data after construction.  Maybe the geometry data assignment overloads should copy the sType too?

This is on header version 137.

asuessenbach:
Resolves #571.

### DIFF
--- a/VulkanHppGenerator.cpp
+++ b/VulkanHppGenerator.cpp
@@ -3007,36 +3007,42 @@ void VulkanHppGenerator::appendStructs(std::string & str) const
   }
 }
 
-void VulkanHppGenerator::appendStructSetter(std::string & str, std::string const& structureName, MemberData const& memberData) const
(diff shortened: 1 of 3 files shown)


jeremy-lunarg:
I think travis is stuck on this job.

PR #590, opened 2020-05-01
Provide rvalue cast operator for ResultValue

### LINKED ISSUE #589: Inconsistency in rvalue handling of ResultValue vs ResultValueType
In recent vulkan-hpp versions, when using the "enhanced mode", some objects are created as ResultValue<> whereas others use ResultValueType<>::type. This leads to the following inconsistent behavior when dealing with rvalues:

```
#include <vulkan/vulkan.hpp>

template <typename T>
struct Resource {
    Resource(T&& raw) : raw{raw} {}
    T raw;
};

int main()
{
    auto device = vk::Device{};

    // Succeeds (createImage returns ResultValueType::type)
    Resource<vk::Image>{device.createImage({})};
    // Fails with "cannot bind rvalue reference of type ‘vk::Pipeline&&’
    // to lvalue of type ‘vk::Pipeline’"
    // (createGraphicsPipeline returns ResultValue).
    Resource<vk::Pipeline>{device.createGraphicsPipeline({}, {})};
}
```
One can explicitly use the ResultValue::value member to work around this:
```
Resource<vk::Pipeline>{device.createGraphicsPipeline({}, {}).value};
```

but this is far from ideal, since we then have to know about the implementation details of each create*() method (and such details may change).

afrantzis:
Allows objects returned as ResultValue (instead of ResultValueType::type) to be moved without requiring additional casting or explicit use of ResultValue::value.

Resolves #589

### DIFF
--- a/VulkanHppGenerator.cpp
+++ b/VulkanHppGenerator.cpp
@@ -7859,6 +7859,11 @@ namespace std
     {
       return value;
     }
+
+    operator T&& () && VULKAN_HPP_NOEXCEPT
+    {
+      return std::move( value );
+    }
 #endif
   };
 
--- a/tests/ResultValueRValue/CMakeLists.txt
+++ b/tests/ResultValueRValue/CMakeLists.txt
@@ -1,0 +1,38 @@
+# Copyright(c) 2020, Collabora Ltd. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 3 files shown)


asuessenbach:
This looks like a reasonable addition. Thanks for contributing to this project!
Would you mind, though, to add the generated vulkan.hpp as well?

afrantzis:
@asuessenbach Hi! I have updated this MR to include the change to the generated file. I also changed the formatting in the generator string slightly to better match what's mostly used elsewhere (not that it matters much since everything is passed through clang-format).

Do you think it would be worth it to also add a small compilation test case for this behavior? If so, what's the best place for it (tests?, samples?).

Thanks!

asuessenbach:
A little compilation test would be great to have!
But I would not call it required.
If you'd like to add such a test, please introduce a new project in the Tests directory.

afrantzis:
> If you'd like to add such a test, please introduce a new project in the Tests directory.

Done!

asuessenbach:

Issue #600, opened 2020-05-06
Vulkan.hpp pollutes namespace by windows.h


pc-john:
Hi all,

after upgrading Vulkan SDK, I cannot compile some projects. It turned out that vulkan.hpp now pollutes c++ namespace by huge number of macros, such as min, max, near, far, and tons of others. I was scared when I find out that it now pulls windows.h in by default. Recently, it was included only when VK_USE_PLATFORM_WIN32_KHR was defined. And I made sure to define it in very specific cpp files only and never in headers to minimize its impact.

Now, with DynamicLoader,... yes, there are workarounds (VULKAN_HPP_ENABLE_DYNAMIC_LOADER_TOOL set to 0). But in my opinion, such horrible thing as windows.h should not be included by default. It not only makes compilation much longer but also pollutes the namespace by so many macros and garbage. And you have to fight through, especially in big projects... Endless renaming of variables, functions, etc. to make them not clash with windows.h garbage.

I made some investigation. The whole problem is just because of HMODULE type and three functions - LoadLibrary, FreeLibrary and GetProcAddress. It seems to me that these might be handled without including windows.h. Would you be open to discuss such solution? Including windows.h is evil in my eyes unless completely necessary.

John

asuessenbach:
As you mentioned, we include windows.h only if _WIN32 is defined and `VULKAN_HPP_ENABLE_DYNAMIC_LOADER_TOOL == 1`
Guarding that include by VK_USE_PLATFORM_WIN32_KHR would not work with cross-platform frameworks like glfw.
If you have a reliable method to prevent inclusion of windows. h, we'd be happy to hear about it.

pc-john:
As you wrote, VULKAN_HPP_ENABLE_DYNAMIC_LOADER_TOOL == 1 is enabling it on Windows. However, even not defining the macro is enabling it as vulkan.hpp will define it itself and set it to 1. Anyway, that might be the best solution if we manage to get rid of windows.h .

I investigated two approaches and succeeded only on the second one. The first attempt was directed to redefining all the required stuff myself instead of windows.h . Particularly, LoadLibrary(), FreeLibrary(), GetProcAddress(), HMODULE and TEXT macro. We can put the replacements of these to vulkan.hpp, but unfortunately, it would fail if an user will include also windows.h, like:
#include <vulkan/vulkan.hpp>
#include <windows.h>
[...]

PR #635, opened 2020-06-17
[HPP] Library Loading Signature and Casting Changes for Clang Windows Resolves #629

### LINKED ISSUE #629: Vulkan HPP Compilation with Clang on Windows
Vulkan HPP does not compile when using clang on Windows since clang interprets Windows defined constructs such as LPCSTR as distinctly different than char const *:

    extern "C" __declspec( dllimport ) void * __stdcall LoadLibraryA( char const * lpLibFileName );
    extern "C" __declspec( dllimport ) int __stdcall FreeLibrary( void * hLibModule );
    extern "C" __declspec( dllimport ) void * __stdcall GetProcAddress( void * hModule, char const * lpProcName );

Ergo, clang believes the library loading functions do not share the same signatures as those defined in the Window's header libloaderapi.h.  Hence, the fix to permit the Vulkan HPP to compile on Windows with clang is to simply change the signature and use reinterpret cast whenever these functions are called.:

    extern "C" __declspec( dllimport ) HMODULE __stdcall LoadLibraryA( LPCSTR lpLibFileName );
    extern "C" __declspec( dllimport ) BOOL __stdcall FreeLibrary( HMODULE hLibModule );
    extern "C" __declspec( dllimport ) FARPROC __stdcall GetProcAddress( HMODULE hModule, LPCSTR lpProcName );

I have a patch for this, but I'm not certain if this is the repo to upload it to since it seems that the hpp file is also found in the Vulkan-Headers repo.

Fierykev:
The alterations to the code are intended to permit Clang to be able to compile the HPP file on Windows.  All alterations do not change the nature of the underlying code, but rather specify function signatures and inputs in a manner that Clang will accept.


asuessenbach:
Hi @Fierykev , thanks for providing a PR on this repo.
But note, though, that the file vulkan.hpp is generated by the VulkanHppGenerator.cpp. Thus, any changes you want to have in the vulkan.hpp would have to be done in the generator.

Besides that, the changes you're proposing would need to include windows.h. Please see #600 why we chose to not do so any more.

Fierykev:
Understood.  Unfortunately, clang will not recognize the current signatures in the HPP as the same as those defined by windows in libloader.h since typedefs are used in the windows definition of the functions.  Fortunately, windows.h is included when VK_USE_PLATFORM_WIN32_KHR is defined (vulkan.h).  Since all of the current tests and samples for Win32 define VK_USE_PLATFORM_WIN32_KHR as a preprocessor definition, there should not be any issue with the repo compiling.  The only possible problem would occur when VK_USE_PLATFORM_WIN32_KHR is not defined which may not be problematic if the standard is that the platform define is always required.  If it is not always required, then it would be necessary for the user to include windows.h or libloader.h prior to including vulkan.h.  The final possible solution would be to only have this alteration enabled for when clang and windows preprocessor defintions are present so that the underlying issues are resolved for the compiler on windows and behavior is preserved for other compilers.

To your other point regarding the HPP generator file, I would be more than happy to investigate integrating my changes into it as long as the function signature changes are acceptable.  If those changes do not comply with the goal of the HPP file, the cast alterations I'm proposing would not be useful since they are only needed to resolve the clang compilation issues.

PR #639, opened 2020-06-18
Replace hardcoded namesapce 'vk::' to 'VULKAN_HPP_NAMESPACE'.

byzin:
Fix build errors due to hardcoded namespace `vk::` when user defined `VULKAN_HPP_NAMESPACE` is used.


byzin:
I'll create another branch and make those changes in VulkanHppGenerator.cpp.

PR #640, opened 2020-06-18
Fix the #if condition of ResultValue<UniqueHandle>.

byzin:
Fix a build error when `VULKAN_HPP_NO_SMART_HANDLE` is defined.


asuessenbach:
Hi @byzin, your change seems to be ok.
Note, though, that vulkan.hpp is a generated file. Please make those changes in VulkanHppGenerator.cpp.

byzin:
Oh thank you for the advice. I'll create another branch.

PR #664, opened 2020-07-08
Guard uses of ArrayPoxyNoTemporaries

### LINKED ISSUE #663: setGeomtries method uses ArrayProxyNoTemporaries but is not guarded by  !defined(VULKAN_HPP_DISABLE_ENHANCED_MODE)
I get a compilation error because a few method definitions use an enhanced-mode template, but the method declarations are not guarded by the necessary #ifdef.

The first example is in AccelerationStructureInfoNV:

```
    AccelerationStructureInfoNV & setGeometries(
      VULKAN_HPP_NAMESPACE::ArrayProxyNoTemporaries<const VULKAN_HPP_NAMESPACE::GeometryNV> const & geometries_ )
      VULKAN_HPP_NOEXCEPT
```
at line 28668 of vulkan/vulkan.hpp

dneto0:
Fixes #663

### DIFF
--- a/VulkanHppGenerator.cpp
+++ b/VulkanHppGenerator.cpp
@@ -3830,12 +3830,14 @@ void VulkanHppGenerator::appendStructSetter( std::string &                   str
       }
 
       static const std::string setArrayTemplate = R"(
+#if !defined(VULKAN_HPP_DISABLE_ENHANCED_MODE)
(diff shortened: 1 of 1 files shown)


dneto0:
Most of the diff is due to clang-format differences. I used clang-format from LLVM top-of-tree.

asuessenbach:
Due to the pure formatting changes, I created a PR without those changes. I hope, you're fine with that.

PR #665, opened 2020-07-09
Disable implicit move constructors for wrapper structs

### LINKED ISSUE #659: Implicit conversion of ResultValue<Pipeline> to Pipeline not working
vk::Pipeline _graphicsPipeline;

...

_graphicsPipeline = _device.createGraphicsPipeline(nullptr, graphicsPipelineCreateInfo);

for some reason this fails with 'operator =' is ambiguous.

vulkan version 1.2.141.2

mocabe:
This PR disables implicitly generated move constructors for handles and structs.
Should fix #659.

### DIFF
--- a/VulkanHppGenerator.cpp
+++ b/VulkanHppGenerator.cpp
@@ -3192,6 +3192,8 @@ ${enter}  class ${className}
       : m_${memberName}(VK_NULL_HANDLE)
     {}
 
+    VULKAN_HPP_CONSTEXPR ${className}( const ${className}& ) VULKAN_HPP_NOEXCEPT = default;
+
     VULKAN_HPP_CONSTEXPR ${className}( std::nullptr_t ) VULKAN_HPP_NOEXCEPT
       : m_${memberName}(VK_NULL_HANDLE)
     {}
--- a/VulkanHppGenerator.cpp
+++ b/VulkanHppGenerator.cpp
@@ -3192,7 +3192,7 @@ ${enter}  class ${className}
       : m_${memberName}(VK_NULL_HANDLE)
     {}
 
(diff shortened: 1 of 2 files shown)


mocabe:
@tksuoran Thank you for pointing that out!

asuessenbach:
I think, as we currently can't find a complete solution to the implicit-cast issue on ResultValue<Type> (see #680), this PR can be closed out.

mocabe:
@asuessenbach I agree that implicit conversions are source of headache, but I think removing them at this point is not ideal solution either. Can't we provide some mitigations and let them being deprecated for some time for users?  
Also, I believe this PR can fix some of these issues at least, do you have any opinion on this?

mtavenrath:
@mocabe The biggest issue I see is that different compilers behave different. On clang/gcc all issues (except maybe for the std::vector return value) have been solved. On msvc we haven't been able to solve a const T& vs. T&& ambiguity. Do you think that it's a good idea to let code compile on linux where people most likely use gcc which fails on windows with msvc later on?

mocabe:
@mtavenrath I don't think there's reasonable fix for this even when all compilers works correctly, so removing implicit conversions is the right way I guess. However, I'm concerned about that removing it right away breaks a lot of existing codes, as we see many bug reports which indicates there're considerable amount of them written based on these conversions. Not fixing the bug might be okay, but I think we can be a little bit more sensible about breaking the API itself.

Here's what I think the most gentle solution:
* (In next release) Fix some of common ambiguous errors, if not all of them. Deprecate implicit conversions (`[[deprecated]]` can be used).
* (In future release) Remove them entirely.

asuessenbach:
@mocabe I think, that's a good approach. I'll mark the implicit cast-operators in #680 as deprecated, for now.

asuessenbach:
Closing this out, as it is resolved by #680 now.

PR #675, opened 2020-07-15
Rename CMake variable for Vulkan registry

tksuoran:
This is change is part of my effort to make Vulkan github projects work better as CMake subprojects using git submodules.

`VulkanRegistry_DIR` and `VulkanHeaders_INCLUDE_DIRS` is what `Vulkan-Tools/cmake/FindVulkanHeaders.cmake` uses.

```
# This module is intended to be used by projects that build Vulkan
# "system" components such as the loader and layers.
```

As a side note, it seems like Vulkan-Hpp should also use `FindVulkanHeaders.cmake` like Vulkan-Tools does. I do this in my CMake parent project, which is what makes these variables `VulkanRegistry_DIR` and `VulkanHeaders_INCLUDE_DIRS` work here for me.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -55,20 +55,20 @@ endif(MSVC)
 
 set(CMAKE_CXX_STANDARD 11)
 
-if (NOT DEFINED VULKAN_HPP_VULKAN_HEADERS_SRC_DIR)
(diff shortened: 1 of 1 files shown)


asuessenbach:
Thanks a lot for you pull-request.
I'ld like to keep VULKAN_HPP_VULKAN_HEADERS_SRC_DIR and VULKAN_HPP_PATH as a backup though.
Would you please adjust your PR to handle that?

asuessenbach:
I've done a PR with back-up right now, rendering this one obsolete.
Nevertheless, thanks a lot for your help!

PR #678, opened 2020-07-16
Provide implicit conversions from ResultValue to value type.

asuessenbach:


asuessenbach:
@everyone: as we can't manage to get that implicit cast from ResultValue<Type> to Type working (especially with vectors of stuff and UniqueHandles), we think we stop this approach here.
That is, if you get a ResultValue<Type> from a vulkan-hpp function, you have to handle that explicitly.
If you have any thoughts on this, please let us know!

PR #679, opened 2020-07-17
Make struct copy constructors explicit

tksuoran:
This removes compiler warnings about implicit copy constructors being deprecated.


asuessenbach:
I think, with the removal of the implicit cast operators of ResultValue (see #680), this PR gets outdated?

asuessenbach:
Closing this, as it is resolved by #680 now.

PR #722, opened 2020-08-24
Introduce mapping between DebugReportObjectType enum value and the corresponding cpp handle type.

### LINKED ISSUE #713: Get DebugReportObjectTypeEXT from Vulkan Object
In my code-base I have a template function to give me the corresponding `DebugReportObjectTypeEXT` of a given Vulkan object. It would be great if the Vulkan-Hpp could provide this.

My ad hoc implementation looks like this:

```c++
template<typename T>
constexpr vk::DebugReportObjectTypeEXT getDebugReportObjectTypeEXT()
{
    if constexpr(false) {
    }

#define VULKAN_TYPE(_type) \
    else if constexpr(std::is_same_v<T, vk::_type>) { return vk::DebugReportObjectTypeEXT::e##_type; }
#include "raygun/utils/vulkan_types.def"

    else {
        return vk::DebugReportObjectTypeEXT::eUnknown;
    }
}
```

Where `vulkan_types.def` contains a list of Vulkan types.

```
VULKAN_TYPE(Instance)
VULKAN_TYPE(PhysicalDevice)
VULKAN_TYPE(Device)
...
```

Side note: This is also used to implement a type trait, which could also be a nice addition. Perhaps implemented independent of each other.

```c++
template<typename T>
struct is_vulkan_type {
    static constexpr bool value = getDebugReportObjectTypeEXT<T>() != vk::DebugReportObjectTypeEXT::eUnknown;
};

template<typename T>
constexpr bool is_vulkan_type_v = is_vulkan_type<T>::value;
```

asuessenbach:
+ new type trait isVulkanHandleType<>

Resolves #713

### DIFF
--- a/VulkanHppGenerator.cpp
+++ b/VulkanHppGenerator.cpp
@@ -261,7 +261,10 @@ void checkAttributes( int                                                  line,
   }
 }
 
-void checkedAssignment( std::string & str, std::string const& leave, std::string const& commandName, std::string aliasName )
+void checkedAssignment( std::string &       str,
+                        std::string const & leave,
+                        std::string const & commandName,
+                        std::string         aliasName )
 {
   str.erase( str.length() - leave.length() );
   str += "      if ( !" + commandName + " ) " + commandName + " = " + aliasName + ";\n" + leave;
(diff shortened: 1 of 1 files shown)


W4RH4WK:
Thank you very much.

PR #723, opened 2020-08-24
Fixup usage of CMake variables.

dj2:
This CL fixes the VULKAN_HPP_VULKAN_HEADERS_SRC_DIR and VULKAN_HPP_PATH
usages to use the values of the variables instead of the text strings.


asuessenbach:

PR #791, opened 2020-10-21
Fix T & ArrayWrapper1D::operator[](int index)

xantares:
With MinGW I get the following error, because constexpr clashes with const:
vulkan.hpp:544:30: error: ‘constexpr T& vk::ArrayWrapper1D<T, N>::operator const’
cannot be overloaded with ‘constexpr const T& vk::ArrayWrapper1D<T, N>::operator const’
544 | VULKAN_HPP_CONSTEXPR T & operator[](int index) VULKAN_HPP_NOEXCEPT

### DIFF
--- a/VulkanHppGenerator.cpp
+++ b/VulkanHppGenerator.cpp
@@ -10250,7 +10250,7 @@ int main( int argc, char ** argv )
       return std::array<T, N>::operator[](index);
     }
 
-    VULKAN_HPP_CONSTEXPR T & operator[](int index) VULKAN_HPP_NOEXCEPT
+    T & operator[](int index) VULKAN_HPP_NOEXCEPT
     {
       return std::array<T, N>::operator[](index);
     }


asuessenbach:
@xantares Usually, we need to have the generated vulkan.hpp in the PR as well, to really see the consequences of the change.
But with this really simple change, I think it's ok to just take it.

Thanks a lot for contributing!

xantares:
@asuessenbach why is there a specific case for _WIN64 in the first place here ?

asuessenbach:
Those two array subscript operators have been introduced von win32 only, because we've got some errors on win32 only.
Try, for example, samples/PhysicalDeviceGroups/PhysicalDeviceGroups.cpp, line 57 with those accessors removed:
`error C2666: 'std::array<vk::PhysicalDevice,32>::operator []': 4 overloads have similar conversions`

Issue #844, opened 2020-12-17
Compilation errors of Instance*, ArrayProxy & DesignatedInitializers samples


khadoran:
Checked out version: 4ec5b4a
cmake version: 3.17.2
gcc version: 9.3.1

As compilation treats warnings as errors, I encounter the following problems compiling the samples:

Some samples (Instance* and ArrayProxy) define EngineName and AppName but don't use them (triggering -Werror=unused-variable)

--

ArrayProxy does not compile for std::initializer_list generating these errors (multiple encountered, only one shown as an example):

In file included from Vulkan-Hpp/tests/ArrayProxy/ArrayProxy.cpp:18:
Vulkan-Hpp/vulkan/vulkan.hpp: In instantiation of 'vk::ArrayProxy<T>::ArrayProxy(const std::initializer_list<_Tp>&) [with T = const int]':
Vulkan-Hpp/tests/ArrayProxy/ArrayProxy.cpp:119:18:   required from here
Vulkan-Hpp/vulkan/vulkan.hpp:257:29: error: initializing 'vk::ArrayProxy<const int>::m_ptr' from 'std::initializer_list<const int>::begin' does not extend the lifetime of the underlying array [-Werror=init-list-lifetime]
  257 |       , m_ptr( list.begin() )
      |                             ^

--

In DesginatedInitializers, variables ai1, ai2, ai3, ai4, ai5, ai6 are not used, resulting in errors and -Werror=deprecated-copy is triggered multiple times in vulkan/vulkan.hpp (lines 17591, 17595, 17600, 26288). This is the g++ output for the first line:

In file included from Vulkan-Hpp/tests/DesignatedInitializers/DesignatedInitializers.cpp:20:
Vulkan-Hpp/vulkan/vulkan.hpp: In constructor 'vk::AccelerationStructureGeometryDataKHR::AccelerationStructureGeometryDataKHR(vk::AccelerationStructureGeometryTrianglesDataKHR)':
Vulkan-Hpp/vulkan/vulkan.hpp:17591:31: error: implicitly-declared 'vk::AccelerationStructureGeometryTrianglesDataKHR::AccelerationStructureGeometryTrianglesDataKHR(const vk::AccelerationStructureGeometryTrianglesDataKHR&)' is deprecated [-Werror=deprecated-copy]
17591 |       : triangles( triangles_ )
      |                               ^
[...]

PR #845, opened 2020-12-30
Fix compiler warnings on gcc/clang

mtavenrath:
Fix issue #844

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -113,7 +113,10 @@ add_executable(VulkanHppGenerator
 set_property(TARGET VulkanHppGenerator PROPERTY CXX_STANDARD 14)
 
 if(MSVC)
-  target_compile_options(VulkanHppGenerator PRIVATE /W4 /WX /permissive-)
+  target_compile_options(VulkanHppGenerator PRIVATE /W4 /WX)
+  if (MSVC_VER GREATER_EQUAL 1910)
+   target_compile_options(VulkanHppGenerator PRIVATE /permissive-)
+  endif()
 else(MSVC)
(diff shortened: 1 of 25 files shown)


tests/ArrayProxy/CMakeLists.txt
```diff
@@ -34,4 +34,7 @@ if (NOT TESTS_BUILD_ONLY_DYNAMIC)
 
 	set_target_properties(ArrayProxy PROPERTIES FOLDER "Tests")
 	target_link_libraries(ArrayProxy "${Vulkan_LIBRARIES}")
+	if (UNIX)
```
asuessenbach, line 37: In the samples, handling such warnings in done in the cpp files, instead of the CMakeLists.txt, like so (from 02_EnumerateDevices.cpp):

#if defined( _MSC_VER )
// no need to ignore any warnings with MSVC
#elif defined( __clang__ )
#  pragma clang diagnostic ignored "-Wunused-variable"
#elif defined( __GNUC__ )
#  pragma GCC diagnostic ignored "-Wunused-but-set-variable"
#else
// unknow compiler... just ignore the warnings for yourselves ;)
#endif

Would you mind adjusting your fix accordingly?

tests/DesignatedInitializers/CMakeLists.txt
```diff
@@ -36,4 +36,7 @@ if (NOT TESTS_BUILD_ONLY_DYNAMIC)
 
 	set_target_properties(DesignatedInitializers PROPERTIES FOLDER "Tests")
 	target_link_libraries(DesignatedInitializers "${Vulkan_LIBRARIES}")
+	if (UNIX)
```
asuessenbach, line 39: See comment above.

PR #890, opened 2021-02-24
Add support for QNX.

mgorchak-blackberry:
Hi,

This PR adds support for QNX platform.

Thanks!


asuessenbach:
@mgorchak-blackberry 
Adding `_screen_window` to the set of specialPointerTypes means, `vkGetPhysicalDeviceScreenPresentationSupportQNX` is not supposed to fill that struct, but just uses it. Is that what you intend to do?
If so, is there any reason to not make `window` a const pointer in that function?

mgorchak-blackberry:
> If so, is there any reason to not make `window` a const pointer in that function?

The window object could be altered by calling QNX Screen composition manager by this API call. I cannot guarantee that it will not be changed. Composition manager could pre-cache some settings to reduce amount of calls to composition manager.

PR #891, opened 2021-02-24
Semanti issue (proposal) (paragraph 53)

kant:


asuessenbach:
The sentence you're changing is correct as it is.

PR #1004, opened 2021-06-30
Add void* support for basetype

pdaniell-nv:
This is to allow this:
```xml
        <type category="basetype">typedef <type>void</type>* <name>VkOpaqueType</name>;</type>
```


asuessenbach:
This PR just extends the check, but does not handle the additional prefix/postfix information for the basetype.
#1005 implements that as well.

PR #1031, opened 2021-07-28
Update to VK_HEADER_VERSION 186.

asuessenbach:


samoylenkodmitry:
Hello @asuessenbach!
Starting from this commit, I can no longer build ALVR for linux: https://github.com/alvr-org/ALVR/issues/768
The error message is `‘const struct alvr::VkContext::dispatch’ has no member named ‘getVkHeaderVersion’.
I found out that started from version 186 and assertion line in `vulkan_funcs.hpp`:
``` VULKAN_HPP_ASSERT( d.getVkHeaderVersion() == VK_HEADER_VERSION );```

Can you please guide what the possible underlying problem is?

**Comment by asuessenbach** [comment]
`getVkHeaderVersion` is part of the `vk::DispatcherLoaderBase`, and it's available in debug builds only (guarded by `#if !defined( NDEBUG )`, as usual). And it's used in vulkan.hpp (and its sub-headers) only with the assertion macro `VULKAN_HPP_ASSERT`.
So, it seems, it's an issue with your build environment.
Could you please try what you get, when you remove the guard around `getVkHeaderVersion` in vulkan.hpp?

**Comment by ichlubna** [comment]
> Could you please try what you get, when you remove the guard around `getVkHeaderVersion` in vulkan.hpp?

Edited `usr/include/vulkan/vulkan.hpp` like this:
```
//#if !defined( NDEBUG )
  public:
    size_t getVkHeaderVersion() const
    {
      return vkHeaderVersion;
    }
  private:
    size_t vkHeaderVersion = VK_HEADER_VERSION;
//#endif
```
Recompiled my project and the problem is still there. Yes, you're right though, when I switch to release I can run it but you know, validation layers and stuff depend on the debug build.

**Comment by ichlubna** [comment]
> Do you use vulkan.hpp in different compilation units? And are they all using the very same VK_HEADER_VERSION?

Well It is included elsewhere via a common header file but using it only in one unit. Not sure what you mean about the VK_HEADER_VERSION, I just use the installed headers as they are. It's 187 in vulkan_core.h and the static assertion in vulkan.hpp is also set to 187. Hmm maybe I should have wrote this in the 187 update thread instead, sorry ^_^". I might just wait or try next updates to see if the problem is still present.

**Comment by asuessenbach** [comment]
Can you break where that assertion fires and look at the value of DispatchLoaderBase::vkHeaderVersion? Is it different from your VK_HEADER_VERSION?

**Comment by ichlubna** [comment]
> Can you break where that assertion fires and look at the value of DispatchLoaderBase::vkHeaderVersion? Is it different from your VK_HEADER_VERSION?

Aaah that's weird. I got it printed in the getVkHeaderVersion before the return and getting 187 but right before the crash the vkHeaderVersion holds some numbers like 94890342912928 or 94471860460144. Overflow or memory corruption maybe?

**Comment by asuessenbach** [comment]
That is, a number of times, that assertion is quiet, and then it fires due to such values?
Can you set a Data Breakpoint on that member?

**Comment by ichlubna** [comment]
> That is, a number of times, that assertion is quiet, and then it fires due to such values?
That's right!
> Can you set a Data Breakpoint on that member?
Hmm I don't know the problematic instance I think. It's apparently happening every time I destroy an unique handle, in my case when destroying an unique shader module. Even when I leave that variable alive it happens again when I destroy for example unique command buffer.

Example backtraces:
```
#4  0x000055555559243d in vk::Device::free<vk::DispatchLoaderStatic> (d=..., commandBuffers=..., commandPool=..., this=0x7fffffffe2c0)
    at /usr/include/vulkan/vulkan_funcs.hpp:4431
#5  vk::PoolFree<vk::Device, vk::CommandPool, vk::DispatchLoaderStatic>::destroy<vk::CommandBuffer> (this=0x7fffffffe2c0, t=...)
    at /usr/include/vulkan/vulkan.hpp:5814
#6  0x000055555558a2f9 in vk::UniqueHandle<vk::CommandBuffer, vk::DispatchLoaderStatic>::~UniqueHandle (this=0x7fffffffe2c0, 
    __in_chrg=<optimized out>) at /usr/include/vulkan/vulkan.hpp:1286
#7  0x000055555556fc01 in GpuVulkan::transitionImageLayout (this=0x555555682cc0, image=..., format=vk::Format::eD32Sfloat, 
    oldLayout=vk::ImageLayout::eUndefined, newLayout=vk::ImageLayout::eDepthStencilAttachmentOptimal)

```

```
#4  0x000055555558df7f in vk::Device::destroy<vk::DispatchLoaderStatic> (d=..., allocator=..., shaderModule=..., this=0x7fffffffe260)
    at /usr/include/vulkan/vulkan_funcs.hpp:2577
#5  vk::ObjectDestroy<vk::Device, vk::DispatchLoaderStatic>::destroy<vk::ShaderModule> (this=0x7fffffffe260, t=...)
    at /usr/include/vulkan/vulkan.hpp:5678
#6  0x0000555555587c93 in vk::UniqueHandle<vk::ShaderModule, vk::DispatchLoaderStatic>::~UniqueHandle (this=0x7fffffffe260, 
    __in_chrg=<optimized out>) at /usr/include/vulkan/vulkan.hpp:1286
```

**Comment by asuessenbach** [comment]
> Hmm I don't know the problematic instance I think.

How many `vk::DispatchLoaderStatic` do you have? I would expect there is exactly one??

Besides that: is that a regression? Does it not happen with an older version of vulkan.hpp?

**Comment by ichlubna** [comment]
> How many `vk::DispatchLoaderStatic` do you have? I would expect there is exactly one??
 
To be honest, I don't understand where it comes from since I don't use this class directly. Only [here](https://github.com/ichlubna/lfPlayer/blob/main/src/gpuVulkan.cpp#L163).

> Besides that: is that a regression? Does it not happen with an older version of vulkan.hpp?
 
Yes, started after an update yesterday (from log: vulkan-headers (1:1.2.185-1 -> 1:1.2.188-1)).

**Comment by asuessenbach** [comment]
Well, no matter how many instances of `vk::DispatchLoaderStatic` you have, could you (temporarily!) add a constructor and a destructor (that do nothing) to `vk::DispatchLoaderBase` in vulkan.hpp, and set a breakpoint there? That might give you some insight what happens here.
Sidenote: you don't need to explicitly include vulkan_enums.hpp and vulkan_structs.hpp, as they get included by vulkan.hpp.

**Comment by ichlubna** [comment]
> add a constructor and a destructor

The value seems to be correct when I take a look during construction or destruction. The value must be later accidentally overwritten. Hard to find where though.

> Sidenote

Thanks, not sure why I got it there. :D

**Comment by asuessenbach** [comment]
And now you can set data breakpoints, and verify that you don't access it after destruction.

**Comment by ichlubna** [comment]
Hmmm yea I think it is really the problem with accessing the data after destruction. It happens when I send the resources of unique handles to functions such as shader module to pipeline creation and destroy the handle afterwards. Is it really wrong though? For example [here](https://github.com/ichlubna/lfPlayer/blob/main/src/gpuVulkan.cpp#L453-L461) the crash happens at the end of the iteration when the shader module handle is destroyed. I know that this tutorial is not flawless but it's [here](https://github.com/Overv/VulkanTutorial/blob/master/code/15_hello_triangle.cpp#L536) as well. Another thing is that it's just the assertion, everything works fine and no validation errors are printed, and also this problem didn't occur before. I can confirm now after downgrading that it started from 186.
I am not sure if keeping the handles around is the right way to solve this.

**Comment by asuessenbach** [comment]
I think I know what happens!
For a test, would you mind replacing some part in your vulkan.hpp?

Please replace this section:
```
#if !defined( VULKAN_HPP_DEFAULT_DISPATCHER )
#  if VULKAN_HPP_DISPATCH_LOADER_DYNAMIC == 1
#    define VULKAN_HPP_DEFAULT_DISPATCHER ::VULKAN_HPP_NAMESPACE::defaultDispatchLoaderDynamic
#    define VULKAN_HPP_DEFAULT_DISPATCH_LOADER_DYNAMIC_STORAGE                     \
      namespace VULKAN_HPP_NAMESPACE                                               \
      {                                                                            \
        VULKAN_HPP_STORAGE_API DispatchLoaderDynamic defaultDispatchLoaderDynamic; \
      }
  extern VULKAN_HPP_STORAGE_API DispatchLoaderDynamic defaultDispatchLoaderDynamic;
#  else
#    define VULKAN_HPP_DEFAULT_DISPATCHER ::VULKAN_HPP_NAMESPACE::DispatchLoaderStatic()
#    define VULKAN_HPP_DEFAULT_DISPATCH_LOADER_DYNAMIC_STORAGE
#  endif
#endif

#if !defined( VULKAN_HPP_DEFAULT_DISPATCHER_TYPE )
#  if VULKAN_HPP_DISPATCH_LOADER_DYNAMIC == 1
#    define VULKAN_HPP_DEFAULT_DISPATCHER_TYPE ::VULKAN_HPP_NAMESPACE::DispatchLoaderDynamic
#  else
#    define VULKAN_HPP_DEFAULT_DISPATCHER_TYPE ::VULKAN_HPP_NAMESPACE::DispatchLoaderStatic
#  endif
#endif

#if defined( VULKAN_HPP_NO_DEFAULT_DISPATCHER )
#  define VULKAN_HPP_DEFAULT_ARGUMENT_ASSIGNMENT
#  define VULKAN_HPP_DEFAULT_ARGUMENT_NULLPTR_ASSIGNMENT
#  define VULKAN_HPP_DEFAULT_DISPATCHER_ASSIGNMENT
#else
#  define VULKAN_HPP_DEFAULT_ARGUMENT_ASSIGNMENT         = {}
#  define VULKAN_HPP_DEFAULT_ARGUMENT_NULLPTR_ASSIGNMENT = nullptr
#  define VULKAN_HPP_DEFAULT_DISPATCHER_ASSIGNMENT       = VULKAN_HPP_DEFAULT_DISPATCHER
#endif
```
with this section:
```
#if defined( VULKAN_HPP_NO_DEFAULT_DISPATCHER )
#  define VULKAN_HPP_DEFAULT_ARGUMENT_ASSIGNMENT
#  define VULKAN_HPP_DEFAULT_ARGUMENT_NULLPTR_ASSIGNMENT
#  define VULKAN_HPP_DEFAULT_DISPATCHER_ASSIGNMENT
#else
#  if !defined( VULKAN_HPP_DEFAULT_DISPATCHER_TYPE )
#    if VULKAN_HPP_DISPATCH_LOADER_DYNAMIC == 1
#      define VULKAN_HPP_DEFAULT_DISPATCHER_TYPE ::VULKAN_HPP_NAMESPACE::DispatchLoaderDynamic
#    else
#      define VULKAN_HPP_DEFAULT_DISPATCHER_TYPE ::VULKAN_HPP_NAMESPACE::DispatchLoaderStatic
#    endif
#  endif

#  if !defined( VULKAN_HPP_DEFAULT_DISPATCHER )
#    define VULKAN_HPP_DEFAULT_DISPATCHER ::VULKAN_HPP_NAMESPACE::defaultDispatchLoader
#    define VULKAN_HPP_DEFAULT_DISPATCH_LOADER_STORAGE                                    \
       namespace VULKAN_HPP_NAMESPACE                                                     \
       {                                                                                  \
         VULKAN_HPP_STORAGE_API VULKAN_HPP_DEFAULT_DISPATCHER_TYPE defaultDispatchLoader; \
       }
#    if VULKAN_HPP_DISPATCH_LOADER_DYNAMIC == 1
#      define VULKAN_HPP_DEFAULT_DISPATCH_LOADER_DYNAMIC_STORAGE VULKAN_HPP_DEFAULT_DISPATCH_LOADER_STORAGE
#    endif
  extern VULKAN_HPP_STORAGE_API VULKAN_HPP_DEFAULT_DISPATCHER_TYPE defaultDispatchLoader;
#  endif

#  define VULKAN_HPP_DEFAULT_ARGUMENT_ASSIGNMENT         = {}
#  define VULKAN_HPP_DEFAULT_ARGUMENT_NULLPTR_ASSIGNMENT = nullptr
#  define VULKAN_HPP_DEFAULT_DISPATCHER_ASSIGNMENT       = VULKAN_HPP_DEFAULT_DISPATCHER
#endif
```
And, in addition to that, have this in exactly one of your cpp files:
`VULKAN_HPP_DEFAULT_DISPATCH_LOADER_STORAGE`

asuessenbach:
@ichlubna A fix for this issue is pending: #1066.
See #1065 for the description of a simple temporary solution.

ichlubna:
> See #1065 for the description of a simple temporary solution.

Many thanks! The fix works!

PR #1049, opened 2021-08-11
Fix release build of samples: suppress warnings about unused variables

theHamsta:
Due to `-Werror` I can only build the samples in debug mode. In release mode (clang-14), the warning about unused variables cause the build to fail.

Is there any preference how a warning about a unused variable should be suppressed?

### DIFF
--- a/RAII_Samples/11_InitShaders/11_InitShaders.cpp
+++ b/RAII_Samples/11_InitShaders/11_InitShaders.cpp
@@ -45,14 +45,20 @@ int main( int /*argc*/, char ** /*argv*/ )
     glslang::InitializeProcess();
 
     std::vector<unsigned int> vertexShaderSPV;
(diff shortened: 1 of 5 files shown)


asuessenbach:
Thanks for spotting these issues!

> Is there any preference how a warning about a unused variable should be suppressed?

I would prefer to not introduce some `markUsed`-function. Instead please use the NDEBUG define, like so:
```
#if !defined( NDEBUG )
    bool ok =
#endif
      vk::su::GLSLtoSPV( vk::ShaderStageFlagBits::eVertex, vertexShaderText_PC_C, vertexShaderSPV );
```
Moreover, your change in CreateDebugUtilsMessenger.cpp fixes the wrong issue: the call to 
`    instance.destroyDebugUtilsMessengerEXT( debugUtilsMessenger );
` should not be done with debug builds only!

Would you mind to adjust your PR accordingly?

Issue #1123, opened 2021-11-04
More constexpr/literal types?


Saalvage:
I'm personally a huge fan of `constexpr` and I think this (awesome!) library would quite benefit from turning most data-aggregation structs (such as `ClearDepthStencilValue`) into literal types so that they can be used in conjunction with `constexpr` elsewhere!

asuessenbach:
What structs are you looking for here?
For example the constructors of `ClearDepthStencilValue` are already marked as `VULKAN_HPP_CONSTEXPR`.

Saalvage:
Sorry, I was being really dumb when picking the example.

So I think there are two general cases where `constexpr` can be utilized more (at least that I came across!):
1. Union copy operations. For some reason those are implemented via `memcpy`, which is not valid in a `constexpr`. Either omitting or explicitely defaulting them should result in the same behavior while causing them to be `constexpr` compatible. (I could totally be missing something though.)
2. Basically all setters. This would allow them to be used in other `constexpr` functions. Should be fine to just add the `constexpr` specifier to all of them.

If this sounds reasonable I could try and craft up a PR.

asuessenbach:
@Saalvage: Sounds good! Please go and craft up a PR.
I'm always happy to get the Vulkan-Hpp improved.

PR #1132, opened 2021-11-12
More support for constexpr

Saalvage:
This deals with #1123.

Before I start I should state that I had a bit of trouble with clang-format, I am using the one that comes with LLVM 13.0.0, which I got straight from the [GitHub releases page](https://github.com/llvm/llvm-project/releases/tag/llvmorg-13.0.0)) for x64 Windows. CMake correctly recognized it when building and created the appropriate .clang-format file, but there appear to be some changes as compared to the format present in the current files which I am unable to explain. I've tried installing different version (including 12.0.0 and 11.0.0) but everything only ever got more different! If this is in fact some issue with my setup I'm really sorry. If it is too bothersome I could go ahead and manually correct the rest to the best of my ability.

(More or less) major changes are:
- **Remove explicit copy constructor and assignment operator from unions.**
The automatically generated default should be semantically the same, while supporting `constexpr`.
- **Setters use `constexpr` in C++>=14**
Allows them to be used in other `constexpr` functions.
- **Union constructors are `constexpr` in C++>=14**
A majority of union constructors could be `constexpr` in C++11 as well. The only unions that absolutely can't are those that contain only members that themself aren't constructible in a `constexpr`. But determining both which unions those are as well as dealing with unions that contain *some* non `constexpr`-constructible members seems more bothersome than it's worth. It would allow a few more structs to be `constexpr`-constructible in C++11 as well, but overall it still seems more trouble than it's worth. If anyone disagrees I can take another look.

I've tried my best to adhere to the standards used in the library, if anything slipped through the cracks please let me know and I'll fix it asap! Also if there's anything causing questions feel free to ask and I'll also try to get back to it asap as well!


asuessenbach:
@Saalvage Your change looks really good! Thanks for your contribution.
I just would like to not have the .gitignore file as part of this repository. I think, everyone can have one of its own.
That is, if you would remove that part, I'd be more than happy to approve and merge your PR.

Saalvage:
Sure thing! There we go.

PR #1160, opened 2021-12-07
Make struct cast operators explicit.

asuessenbach:


MennoVink:
@asuessenbach Could you explain the reasoning behind this change? I've come to appreciate that this was working the way it was but updating to the latest headers i now need to place casts throughout my code.
These implicit conversions working is also actually a feature that is listed on the repo's homepage, so maybe these conversions being explicit should become optional?

PR #1164, opened 2021-12-08
Add CI on MacOS.

asuessenbach:


theHamsta:
The hurdles to build on MacOS are:

- You have to use `g++` and `clang++` instead of `gcc` or `clang`. Otherwise it will compile just fine but will not have the C++-Standard library linked as this has to be done explicitly when
- Vulkan-Hpp does not build with `libc++` (compile errors with `std::hash`, will file a bug in minute)
- glslang does not build on MacOS due to some weird linker warning flag that does not seem to exist for ldd or whatever is used as a default on MacOS (can be easily) fixed)
- The samples have an assertion on choosing `VK_USE_PLATFORM_???` which is not handled for MacOS. Probably it should be configurable for all OSes with a default of `VK_USE_PLATFORM_MACOS_MVK` for MacOS
- `clang-format` (and maybe `molten-vk`, probably not necessary for just building) can be install via `brew install clang-format`

CI: https://github.com/theHamsta/Vulkan-Hpp/runs/4730055528?check_suite_focus=true
WIP branch: https://github.com/theHamsta/Vulkan-Hpp/tree/macos

ylz-at:
Any update? Could you please add clang for macOS CI as well?

Issue #1169, opened 2021-12-22
Latest vulkan_hash.hpp triggers a static assert about hashing applicationInfo.pApplicationName


amaiorano:
When building against the latest Vulkan-Headers, our g++ build fails with:

```
In module '//third_party/stl:stl':
ERROR: third_party/stl/cxx17/functional:88:5 static_assert failed due to requirement 'sizeof (t) < 0' "hash<const char*> would hash the pointer, not the contents of the string, which is almost certainly not what you want. To explicitly hash the pointers, use const void*."
    static_assert(sizeof(t) < 0,
    ^             ~~~~~~~~~~~~~
third_party/vulkan_headers/include/vulkan/vulkan_hash.hpp:997:7 in instantiation of function template specialization 'std::hash<const char *>::hash<int>' requested here
      VULKAN_HPP_HASH_COMBINE( seed, applicationInfo.pApplicationName );
      ^
third_party/vulkan_headers/include/vulkan/vulkan_hash.hpp:504:15 expanded from macro 'VULKAN_HPP_HASH_COMBINE'
      seed ^= std::hash<std::decay<decltype( value )>::type>{}( value ) + 0x9e3779b9 + ( seed << 6 ) + ( seed >> 2 )
              ^
1 error generated.
```

I believe the static assert is probably right here. When hashing pApplicationName -- and same for pEngineName -- we probably want to hash the string, and not the pointer value:

```
  template <>
  struct hash<VULKAN_HPP_NAMESPACE::ApplicationInfo>
  {
    std::size_t operator()( VULKAN_HPP_NAMESPACE::ApplicationInfo const & applicationInfo ) const VULKAN_HPP_NOEXCEPT
    {
      std::size_t seed = 0;
      VULKAN_HPP_HASH_COMBINE( seed, applicationInfo.sType );
      VULKAN_HPP_HASH_COMBINE( seed, applicationInfo.pNext );
      VULKAN_HPP_HASH_COMBINE( seed, applicationInfo.pApplicationName ); // <--- here
      VULKAN_HPP_HASH_COMBINE( seed, applicationInfo.applicationVersion );
      VULKAN_HPP_HASH_COMBINE( seed, applicationInfo.pEngineName ); // <--- and here
      VULKAN_HPP_HASH_COMBINE( seed, applicationInfo.engineVersion );
      VULKAN_HPP_HASH_COMBINE( seed, applicationInfo.apiVersion );
      return seed;
    }
  };
```

[...]

PR #1170, opened 2021-12-23
Fix std::hash specializations for null-terminated strings

amaiorano:
Instead of hashing the pointer, hash the string contents.

Fixes: https://github.com/KhronosGroup/Vulkan-Hpp/issues/1169

### DIFF
--- a/VulkanHppGenerator.cpp
+++ b/VulkanHppGenerator.cpp
@@ -11954,6 +11954,29 @@ std::string VulkanHppGenerator::generateStructHashSum( std::string const &      
       }
       hashSum += "    }\n";
     }
+    else if ( member.type.type == "char" && !member.len.empty() )
+    {
+      assert( member.len.size() < 3 );
+      if ( member.len.size() == 1 )
+      {
+        assert( member.len[0] == "null-terminated" );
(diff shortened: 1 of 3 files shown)


asuessenbach:
Your change looks great! Thanks for your contribution.

asuessenbach:
No idea, what's going wrong here with the CI.
Would you mind to re-trigger it by pushing your changes again?

amaiorano:
I rebased on latest master. Let's see if the CI runs now.

asuessenbach:
One thing that came to my mind right now: With your change, std::hash and operator==() would behave differently (on structures containing null-terminated strings, of course).
In your test case, appInfo1 and appInfo2 would give the same hash but would not be equal. Which is allowed, but probably not expected.
That is, I think, operator==() should be extended to also compare null-terminated string (and array of null-terminated string) members of structures.

amaiorano:
> That is, I think, operator==() should be extended to also compare null-terminated string (and array of null-terminated string) members of structures.

I agree, that would make the most sense. I think we'd want operator==() to be used for value-equality, not reference-equality -- for the latter, you could always compare the addresses of the objects. Is this something you can take care of? Or perhaps open a GitHub issue for?

Issue #1185, opened 2022-01-27
'missing braces around initializer' warnings in vulkan_enums.hpp


sugoi1:
See this PR for sample fix:
https://github.com/KhronosGroup/Vulkan-Headers/pull/257

asuessenbach:
Resolved by #1186.

PR #1186, opened 2022-01-28
Fix 'missing braces around initializer' warnings in vulkan_enums.hpp

sugoi1:
This is trying to fix:
https://github.com/KhronosGroup/Vulkan-Hpp/issues/1185

@mikes-lunarg PTAL


sugoi1:
@asuessenbach PTAL

asuessenbach:
Thanks a lot for your contribution! And the fix, of course.

PR #1222, opened 2022-03-02
fix(ci): filter out publications of Vulkan SC on header update

theHamsta:
The publication of Vulkan SC (https://www.khronos.org/news/press/khronos-releases-vulkan-safety-critical-1.0-specification-to-deliver-safety-critical-graphics-compute)
confused the update job since it was the most recent tag.

Preliminary solution: only accept tags that match `v1\.`.


theHamsta:
Will test this change in a separate branch that unconditionally triggers the CI jobs to its effect.

PR #1233, opened 2022-03-08
ci: only create a header update PR if tag does not yet exist

theHamsta:
Run when tag does not yet exist: https://github.com/theHamsta/Vulkan-Hpp/runs/5461820621?check_suite_focus=true (it actually should exist but I haven't run `git fetch --all --tags` in that revision to actually have any tags locally)
Run when tag already exists: https://github.com/theHamsta/Vulkan-Hpp/runs/5462719659?check_suite_focus=true


theHamsta:
On my way to lunch, i git the thought that a `git stash` might be necessary as clean might not undo modifications

PR #1243, opened 2022-03-15
ci: fix bash command to retrieve latest Vulkan-Headers commit

theHamsta:
The problem: grep was filtering on the commit hashes not the tag names.

One should probably never use bash: https://github.com/ValveSoftware/steam-for-linux/issues/3671


theHamsta:
works on my fork https://github.com/theHamsta/Vulkan-Hpp/pull/18

PR #1288, opened 2022-04-19
Add ArrayProxy constructors with support for raw array with size

phantom10111:
Currently `ArrayProxy` doesn't support raw arrays which is annoying. For instance, I have code where I include contents of SPIR-V  shaders (generated with `glslc -mfmt=num`) into `uint32_t[]` array. I can't use `std::array` because I don't know the size in advance. However, since raw arrays are unsupported by `ArrayProxy`, then I have to manually call `setCodeSize` for every shader.

This pull requests adds support for raw arrays, when their size is statically known.

### DIFF
--- a/README.md
+++ b/README.md
@@ -185,6 +185,10 @@ c.setScissor(0, scissorRect);
 // pass a temporary value.
 c.setScissor(0, { { 0, 0 },{ 640, 480 } });
 
+// pass a fixed size array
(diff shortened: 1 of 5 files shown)


theHamsta:
Just regarding the usecase: I use 
```
std::vector<uint32_t> composeVertSpirv {
#include "compose.frag.num"
};
```
 with c++17 it might also work `std::array` https://en.cppreference.com/w/cpp/container/array/deduction_guides

But it would of course be nice to have it work also with raw arrays

asuessenbach:
Again, some useful extension of functionality.

PR #1331, opened 2022-06-20
Add flag adder/remover methods

tomix1024:
When building the contents of a struct by chaining `.setXY()` calls, I think it would be useful if the various flag-like variables could be modified in-place, instead of simply being overwritten.

For this purpose, I added a `addXY()` and `removeXY()` method to the structs for all fields that are bitmasks.
The functions simply set the specified bits, or unset the specified, leaving the other bits in the bitmask alone.

As it is right now I see one potential issue:
We now need the `FlagTraits<FlagBitsType>::allFlags`, even for enums that are empty.
To make the compiler happy, I added a default implementation that sets `FlagTraits<FlagBitsType>::allFlags` to the default-initialized underlying type, since there are no flags anyways, this should be consistent.
The problem is that this way, the `FlagTraits` are specified for _any_ type, so if one forgets to specify the `FlagTraits` for a non-empty enum, the `FlagTraits<FlagBitsType>::allFlags` would be incorrect and the compiler would not complain.
Of cource, when an existing non-empty enum is modified in the future, the same problem exists anyways (so maybe this is a non-issue?).
Alternatively the FlagTraits could also be specialized for all empty bitmasks, but I think this would unneccessarily bloat the code.

### DIFF
--- a/VulkanHppGenerator.cpp
+++ b/VulkanHppGenerator.cpp
@@ -10315,7 +10315,7 @@ std::string VulkanHppGenerator::generateStructSetter( std::string const & struct
   if ( member.type.type != "VkStructureType" )  // filter out StructureType, which is supposed to be immutable !
   {
     static const std::string templateString = R"(
-    ${constexpr}${structureName} & set${MemberName}( ${memberType} ${reference}${memberName}_ ) VULKAN_HPP_NOEXCEPT
+    ${constexpr}${structureName} & ${operation}${MemberName}( ${memberType} ${reference}${memberName}_ ) VULKAN_HPP_NOEXCEPT
     {
       ${assignment};
       return *this;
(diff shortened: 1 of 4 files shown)


asuessenbach:
Thanks for providing this addition to the struct interfaces.
But I think, with this addition, you get a lot of very simple code sequences in vulkan_structs.hpp without much benefit. Those two operations can easily be done by using the assignment operators of `class Flags`.
If at all, you could just add an `add` and a `remove` function to `class Flags`.

asuessenbach:
@tomix1024 Do you want to continue on this PR or may I just close it?

tomix1024:
After playing around with it a little more I think you are right. The benefit is very limited. For the `Flags` class I think the current operators are sufficient.

PR #1334, opened 2022-06-21
Speed up CI by using ninja and parallel build

ylz-at:


ylz-at:
I see build time reduce from 14-20 min to 7-8 min on windows, which looks promising.

ylz-at:
I see some speed up among windows jobs, it takes around 10 min to finish now.

But some jobs hanging for over 2h🤔.

asuessenbach:
@ylz-at Do you want to continue with this PR?

ylz-at:
@asuessenbach I didn't figure out what is the issue with ninja. I can close this if you need and reopen it when I have a valid solution.

PR #1379, opened 2022-08-04
Minor cleanup in utils.cpp: remove helper function createDebugUtilsMessengerEXT.

asuessenbach:


SetoKaiba:
I think that the method define in utils.hpp should be removed as well.

PR #1393, opened 2022-08-19
fix: make VULKAN_HPP_RUN_GENERATOR work again

theHamsta:
Apparently, the statement setting ${vulkan_hpp} was removed with along
with the other generated header files the other generated header files.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -121,6 +121,9 @@ add_executable(VulkanHppGenerator
   ${TINYXML2_HEADERS}
 )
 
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -263,5 +279,6 @@ if (TESTS_BUILD)
 endif ()
 
 if (${VULKAN_HPP_INSTALL})
```
theHamsta, line 281: I'm not sure why the install target is behind a feature guard. The install target is not executed when building `ALL` only when explicitly requested

PR #1396, opened 2022-08-19
Fix clang bugs

### LINKED ISSUE #1394: std::format usage breaks compilation with clang versions prior to 14 using C++20
Clang versions prior to 14 have missing support for std::format. Unfortunately they do have an effectively empty `<format>` header, so the test at lines 13 and 3006 of vulkan_to_string.hpp is satisfied, but causes a build error: No member named 'format' in namespace 'std'.

This affects the current release version of Xcode, which ships with clang 13.

Possibly a specific check for clang 14+ is required before use of std::format?

theHamsta:
Fixes compilation using clang. 

- For clang-16 on Linux using libstdc++. Include <utility> for `std::exchange` 
```
In file included from /home/stephan/projects/Vulkan-Hpp/debug/RAII_Samples/SurfaceFormats/CMakeFiles/RAII_SurfaceFormats.dir/cmake_pch.hxx:5:
/home/stephan/projects/Vulkan-Hpp/vulkan/vulkan_raii.hpp:28:14: error: no template named 'exchange' in namespace 'std'; did you mean simply 'exchange'?
      return std::exchange<T>( obj, std::forward<U>( newValue ) );
             ^~~~~
/home/stephan/projects/Vulkan-Hpp/vulkan/vulkan_raii.hpp:25:49: note: 'exchange' declared here
    VULKAN_HPP_CONSTEXPR_14 VULKAN_HPP_INLINE T exchange( T & obj, U && newValue )
                                                ^
/home/stephan/projects/Vulkan-Hpp/vulkan/vulkan_raii.hpp:28:14: error: no template named 'exchange' in namespace 'std'; did you mean simply 'exchange'?
      return std::exchange<T>( obj, std::forward<U>( newValue ) );
             ^~~~~
/home/stephan/projects/Vulkan-Hpp/vulkan/vulkan_raii.hpp:25:49: note: 'exchange' declared here
    VULKAN_HPP_CONSTEXPR_14 VULKAN_HPP_INLINE T exchange( T & obj, U && newValue )

```
- For clang-13 on Mac using libc++. Use more reliable feature test macro for `std::format`. Fixes #1394

### DIFF
--- a/VulkanHppGenerator.cpp
+++ b/VulkanHppGenerator.cpp
@@ -405,6 +405,7 @@ void VulkanHppGenerator::generateVulkanRAIIHppFile() const
 #  define VULKAN_RAII_HPP
 
 #include <memory>
+#include <utility>  // std::exchange, std::forward
 #include <vulkan/vulkan.hpp>
 
 #if !defined( VULKAN_HPP_RAII_NAMESPACE )
--- a/VulkanHppGenerator.cpp
+++ b/VulkanHppGenerator.cpp
@@ -516,7 +516,7 @@ void VulkanHppGenerator::generateVulkanToStringHppFile() const
 
 #include <vulkan/vulkan_enums.hpp>
 
-#if ( ( 20 <= VULKAN_HPP_CPP_VERSION ) && __has_include( <format> ) )
+#if __cpp_lib_format
 #  include <format>   // std::format
 #else
 #  include <sstream>  // std::stringstream
@@ -5034,7 +5034,7 @@ std::string VulkanHppGenerator::generateEnumsToString() const
 
   VULKAN_HPP_INLINE std::string toHexString( uint32_t value )
   {
-#if ( ( 20 <= VULKAN_HPP_CPP_VERSION ) && __has_include( <format> ) )
+#if __cpp_lib_format
     return std::format( "{:x}", value );
 #else
     std::stringstream stream;
--- a/vulkan/vulkan_raii.hpp
+++ b/vulkan/vulkan_raii.hpp
@@ -9,6 +9,7 @@
 #define VULKAN_RAII_HPP
 
 #include <memory>
+#include <utility>  // std::exchange, std::forward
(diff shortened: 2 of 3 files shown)


VulkanHppGenerator.cpp
```diff
@@ -515,7 +516,7 @@ void VulkanHppGenerator::generateVulkanToStringHppFile() const
 
 #include <vulkan/vulkan_enums.hpp>
 
-#if ( ( 20 <= VULKAN_HPP_CPP_VERSION ) && __has_include( <format> ) )
+#if ( ( 20 <= VULKAN_HPP_CPP_VERSION ) && __cpp_lib_format )
```
theHamsta, line 519: the check for cpp 20 is actually not necessary. `#if __cpp_lib_format` would be enough

theHamsta:
`std::format` is usable with clang and g++. But it appears that they don't advertise full support via `__cpp_lib_format` in contrast to MSVC

VulkanHppGenerator.cpp
```diff
@@ -405,6 +405,7 @@ void VulkanHppGenerator::generateVulkanRAIIHppFile() const
 #  define VULKAN_RAII_HPP
 
 #include <memory>
+#include <utility> // std::exchange
```
asuessenbach, line 408: Please include `utility` only if `( 14 <= VULKAN_HPP_CPP_VERSION )`.

VulkanHppGenerator.cpp
asuessenbach: That's right. That is, please remove the check against `VULKAN_HPP_CPP_VERSION`.

asuessenbach (changes requested):

VulkanHppGenerator.cpp
theHamsta: both `std::exchange` and `std::forward` (https://en.cppreference.com/w/cpp/utility/forward) require `<utility>`

asuessenbach (approved):

PR #1408, opened 2022-09-14
Add vcpkg installation instructions

Adela0814:
vulkan-hpp is available as a port in vcpkg, a C++ library manager that simplifies installation for vulkan-hpp and other project dependencies. Documenting the install process here will help users get started by providing a single set of commands to build vulkan-hpp, ready to be included in their projects.

We also test whether our library ports build in various configurations (dynamic, static) on various platforms (OSX, Linux, Windows: x86, x64, UWP, ARM) to keep a wide coverage for users.

I'm a maintainer for vcpkg, and [here is what the port script looks like](https://github.com/microsoft/vcpkg/blob/master/ports/vulkan-hpp/portfile.cmake). We try to keep the library maintained as close as possible to the original library.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,6 +38,20 @@ To build the local samples and tests you'll have to clone this repository and ru
 
 Optional: To update the Vulkan-Hpp and its submodules execute ```git pull --recurse-submodules```.
 
+### Installing vulkan-hpp using vcpkg
+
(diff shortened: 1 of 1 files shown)


asuessenbach:
Thanks for supporting Vulkan-Hpp with vcpkg!
I'm happy to have this little chapter in the readme now.

PR #1455, opened 2022-11-23
Added ClearColorValue constructors

MennoVink:
This enables simply writing `ClearColorValue( 1.0f, 0.0f, 0.0f, 1.0f )` without having to create a std::array.


asuessenbach:
I like that constructor! Thanks a lot for this idea.
But vulkan_structs.hpp is generated (as all the other vulkan*.hpp files here). Therefore, adding those constructors needs to be done in VulkanHppGenerator.cpp.
I've done that with #1457, therefore I close your PR.

MennoVink:
Cool, thanks for picking it up 👍

PR #1474, opened 2023-01-07
Don't throw on eErrorOutOfDateKHR

MrSapps:
Typical vulkan tutorial code follows this pattern:

```c
    result = mPresentQueue->presentKHR(presentInfo);
    if (result == vk::Result::eErrorOutOfDateKHR || result == vk::Result::eSuboptimalKHR)
    {
        recreateSwapChain();
    }
 ```

Which without this PR isn't possible since presentKHR will throw on eErrorOutOfDateKHR.

### DIFF
--- a/vulkan/vulkan_raii.hpp
+++ b/vulkan/vulkan_raii.hpp
@@ -13549,16 +13549,16 @@ namespace VULKAN_HPP_NAMESPACE
       return std::make_pair( static_cast<VULKAN_HPP_NAMESPACE::Result>( result ), imageIndex );
     }
 
-    VULKAN_HPP_NODISCARD VULKAN_HPP_INLINE VULKAN_HPP_NAMESPACE::Result Queue::presentKHR( const VULKAN_HPP_NAMESPACE::PresentInfoKHR & presentInfo ) const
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by asuessenbach** [comment]
@Paul, you might have notices, that all the vulkan*.hpp files are generated, so any explicit changes on one of those files are pointless.
As you correctly stated, presentKHR throws on eErrorOutOfDateKHR. This is, because that result value is specified as an "error code" for that function. And error codes are mapped to exceptions, here.
In order to handle that error code similar to the code pattern with vulkan C-functions, you would need to do something like
```
vk::Result result;
try
{
	result = queue.present( present_info );
}
catch ( vk::OutOfDateKHRError const & /*err*/ )
{
	result = vk::Result::eErrorOutOfDateKHR;
}
if ( result == vk::Result::eSuboptimalKHR || result == vk::Result::eErrorOutOfDateKHR )
{
        recreateSwapChain();
}
```

theHamsta:
You can also define `VULKAN_HPP_NO_EXCEPTIONS=1` to avoid exceptions globally.

asuessenbach:
Yes, exactly.
One is specified as a success code, the other as an error code.
And, by the way, if you don't want exceptions at that point, you could explicitly use the dispatcher here, like this:
VkResult result = mPresentQueue.getDispatcher()->vkQueuePresentKHR( *mPresentQueue, reinterpret_cast<const VkPresentInfoKHR *>( &presentInfo ) );

PR #1525, opened 2023-03-07
extend workflow run to generate vulkansc*.hpp

asuessenbach:


CMakeLists.txt
```diff
@@ -163,6 +163,7 @@ option (VULKAN_HPP_RUN_GENERATOR "Run the HPP generator" OFF)
 if (VULKAN_HPP_RUN_GENERATOR)
   add_custom_command(
     COMMAND ${PROJECT_NAME}
+    COMMAND ${PROJECT_NAME} -api vulkansc
```
theHamsta, line 166: At the moment, the workflow ignores tags for vulkansc and only works on regular releases (the newest regular release). See https://github.com/KhronosGroup/Vulkan-Hpp/blob/d796a40abbfc958768dc4fe04c0752dcba661da9/.github/workflows/update-header-pr.yml#L33

So regular Vulkan releases will generate VulkanSC headers. This will be probably fine since Vulkan releases more frequently. The workflow currently doesn't check whether there has been a VulkanSC release.

PR #1625, opened 2023-07-18
Fix debug layers callback return value

CodesOtakuYT:
> The application should always return VK_FALSE. The VK_TRUE value is reserved for use in layer development.

Reference: https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/PFN_vkDebugUtilsMessengerCallbackEXT.html#_description


asuessenbach:
Thanks for spotting this issue! And the fix.

PR #1634, opened 2023-08-07
Fix validation layer warning on minImageCount of 2.

asuessenbach:
Resolves validation layer best-practices warning:
`A Swapchain is being created with minImageCount set to 2, which means double buffering is going to be used. Using double buffering and vsync locks rendering to an integer fraction of the vsync rate. In turn, reducing the performance of the application if rendering is slower than vsync. Consider setting minImageCount to 3 to use triple buffering to maximize performance in such cases.`


water-chika:
```surfaceCapabilities.maxImageCount``` maybe 0.
Below statement is from Vulkan Specification:
```maxImageCount``` is the maximum number of images the specified device supports for a ```swapchain```
created for the surface, and will be either 0, or greater than or equal to minImageCount. A value of
0 means that there is no limit on the number of images, though there may be limits related to
the total amount of memory used by presentable images.

PR #1646, opened 2023-08-24
Switch __QNXNTO__ macro check to __QNX__

mgorchak-blackberry:
During last 23 years the QNX Neutrino microkernel was active, __QNXNTO__ macro belongs to it. Now QNX Muon microkernel is under development, so __QNX__ macro covers both versions: Neutrino and Muon.

First public announce of QNX Muon was here: https://blogs.blackberry.com/en/2023/04/get-in-the-sdv-podcast-from-blackberry-season-2-episode-3

### DIFF
--- a/snippets/DynamicLoader.hpp
+++ b/snippets/DynamicLoader.hpp
@@ -10,7 +10,7 @@
     {
       if ( !vulkanLibraryName.empty() )
       {
-#  if defined( __unix__ ) || defined( __APPLE__ ) || defined( __QNXNTO__ ) || defined(__Fuchsia__)
+#  if defined( __unix__ ) || defined( __APPLE__ ) || defined( __QNX__ ) || defined(__Fuchsia__)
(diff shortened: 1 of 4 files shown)


mtavenrath:
@mgorchak-blackberry When switching from __QNXNTO__ to __QNX__, wouldn't this break older versions of the QNX toolchain using Vulkan-Hpp? If so, it'd be better to add the __QNX__ check in addition to the __QNXNTO__ check to avoid breaking legacy code.

mgorchak-blackberry:
__QNX__ is always defined since QNX 2.x. Nowadays __QNX__ is the only common macro for all QNX versions. While __QNXNTO__ will be present in the first versions based on the Muon kernel, but later it could disapper. So __QNX__ is the only safe option. Thanks!

asuessenbach:
Thanks a lot for that information. Will merge your PR then.
Note, though, that we're using tinyxml2 for parsing the xml file. Maybe you should also ask them (https://github.com/leethomason/tinyxml2.git) for that same change.

PR #1667, opened 2023-09-27
Update workflow deps

theHamsta:
Update the used workflows to their newest version to suppress deprecation warnings as in https://github.com/KhronosGroup/Vulkan-Hpp/actions/runs/6319631419

Smoke test for the update-headers-pr seems to still run as before
https://github.com/theHamsta/Vulkan-Hpp/actions/runs/6325102376/job/17175926924

Also replace
```
            -DSAMPLES_BUILD=OFF
            -DTESTS_BUILD=OFF
```
with the actually existing CMake options.

### DIFF
--- a/.github/workflows/ci-macos.yml
+++ b/.github/workflows/ci-macos.yml
@@ -18,7 +18,7 @@ jobs:
         cxx_standard: [11, 14, 17, 20]
 
     steps:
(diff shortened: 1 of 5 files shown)


asuessenbach:
Thanks a lot for these updates and corrections!

PR #1678, opened 2023-10-06
Fix C++20 module build with -DVULKAN_HPP_NO_SMART_HANDLE

shananton:
In `vulkan.cppm`, `vk::UniqueHandle` and `vk::SharedHandle` are under `#if !defined( VULKAN_HPP_DISABLE_ENHANCED_MODE )`, even though they're actually controlled by `VULKAN_HPP_NO_SMART_HANDLE` in the headers.

This PR fixes that issue.
I actually wanted to build with `VULKAN_HPP_NO_SMART_HANDLE`, because I get an internal compiler error with MSVC otherwise :) Haven't been able to figure out what exactly causes it yet

### DIFF
--- a/vulkan/vulkan.cppm
+++ b/vulkan/vulkan.cppm
@@ -45,10 +45,8 @@ export namespace VULKAN_HPP_NAMESPACE
   using VULKAN_HPP_NAMESPACE::ArrayProxy;
   using VULKAN_HPP_NAMESPACE::ArrayProxyNoTemporaries;
   using VULKAN_HPP_NAMESPACE::Optional;
-  using VULKAN_HPP_NAMESPACE::SharedHandle;
   using VULKAN_HPP_NAMESPACE::StridedArrayProxy;
   using VULKAN_HPP_NAMESPACE::StructureChain;
-  using VULKAN_HPP_NAMESPACE::UniqueHandle;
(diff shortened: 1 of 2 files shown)


vulkan/vulkan.cppm
```diff
@@ -45,10 +45,8 @@ export namespace VULKAN_HPP_NAMESPACE
   using VULKAN_HPP_NAMESPACE::ArrayProxy;
   using VULKAN_HPP_NAMESPACE::ArrayProxyNoTemporaries;
   using VULKAN_HPP_NAMESPACE::Optional;
-  using VULKAN_HPP_NAMESPACE::SharedHandle;
```
theHamsta, line 48: This file is autogenerated. Could you try to find the bug in `VulkanHppGenerator.cpp` that causes `vulkan.cppm` generated in the wrong way?

shananton: Didn't know this, I guess I'll have a look

shananton: @theHamsta found it :)

asuessenbach:
Thanks a lot for spotting and fixing this glitch.
And especially for getting the modules working again with Visual Studio, at least with some reduced feature support.
An issue against that is already filed, by the way (https://developercommunity.visualstudio.com/t/On-building-a-C20-module:-fatal--error/10469799#T-ND10485943).

PR #1679, opened 2023-10-06
Use add_subdirectory-friendly CMAKE variables

shananton:
This adjusts some uses of cmake variables, so that Vulkan-Hpp can be built as a subdirectory of another project.


asuessenbach:
Looks like a good improvement of this project.
Thanks a lot for your PR!

PR #1710, opened 2023-10-23
Add extension macros as constexpr

### LINKED ISSUE #1706: Missing items from c++20 module
Writing a hello triangle using the .cppm, all very snazzy and compile times are much improved (if only clangd/intellisense worked).

Couple of things I've found missing so far, mainly related to macros.

Can the cppm define storage vk::defaultDispatchLoaderDynamic? Having to manually write out the definition of VULKAN_HPP_DEFAULT_DISPATCH_LOADER_DYNAMIC_STORAGE is bug prone due to a lack of access to macros for things like dllimport/dllexport. Defining storage in the module should be fine with regards to the ODR, unlike defining it in a header.

Could the .cppm export constexpr variables containing extension names? (i.e. aliases for VK_KHR_SWAPCHAIN_EXTENSION_NAME and the like), as again we don't have access to macros.

sharadhr:
Extension name and spec version macros have been defined as `constexpr` in `vulkan.hpp`, and exported in `vulkan.cppm`.

Partially resolves #1706.


VulkanHppGenerator.cpp
```diff
@@ -23,14 +23,6 @@
 #include <regex>
 #include <sstream>
 
-struct MacroData
```
sharadhr, line 26: I've moved this to the `.hpp` file, because this was hanging around here weirdly, and it made more sense to stay with the rest of the structs in the header.

VulkanHppGenerator.cpp
```diff
@@ -13957,6 +14062,21 @@ void VulkanHppGenerator::readRequireEnum(
             m_constants[name] = { type, value, line };
           }
         }
+
+        // if we have a name and value, insert to enumConstants map
+        if ( attributes.contains( "name" ) && attributes.contains( "value" ) )
+        {
+          /** If I do readEnumsConstants(), it would require that I move "type" from the `required` map to the `optional` map at
```
sharadhr, line 14069: @asuessenbach, thoughts on this bit?

asuessenbach: Yep, I agree. Storing those name/value pairs in a separate bucket is appropriate, as they are very special.

VulkanHppGenerator.cpp
```diff
@@ -13957,6 +14062,21 @@ void VulkanHppGenerator::readRequireEnum(
             m_constants[name] = { type, value, line };
           }
         }
+
+        // if we have a name and value, insert to enumConstants map
+        if ( attributes.contains( "name" ) && attributes.contains( "value" ) )
```
asuessenbach, line 14067: At this point, it's already checked that there is a "value".
I think, you could just add that unconditionally, as "name" was checked as required above.
Maybe you could add a check like
```
        checkForError(
          !supported || name.ends_with( "_EXTENSION_NAME" ) || name.ends_with( "_SPEC_VERSION" ), line, "encountered unexpected enum <" + name + ">" );
```

VulkanHppGenerator.hpp
```diff
@@ -212,11 +220,12 @@ class VulkanHppGenerator
...
   struct RequireData
   {
-    std::string              depends   = {};
-    std::vector<std::string> commands  = {};
-    std::vector<std::string> constants = {};
-    std::vector<std::string> types     = {};
-    int                      xmlLine   = {};
+    std::string                        depends       = {};
+    std::vector<std::string>           commands      = {};
+    std::map<std::string, std::string> enumConstants = {};
```
asuessenbach, line 225: Maybe change that to some `std::vector<NameValuePair>`, or so.

asuessenbach (commented):
Your PR looks great! Thanks a lot for your contribution.
Just a couple of more or less minor issues...

VulkanHppGenerator.cpp
```diff
@@ -4989,6 +4981,103 @@ std::string VulkanHppGenerator::generateConstexprDefines() const
...
+    auto const VENDORPascalCaseStripPrefix = []( std::string const & macro )
+    {
+      auto       prefixStripped = stripPrefix( macro, "VK_" );
+      auto const vendor         = prefixStripped.substr( 0, prefixStripped.find( '_' ) );
+      return vendor + toCamelCase( stripPrefix( prefixStripped, vendor + "_" ) );
+    };
+
+    auto const & extensionMacro = std::find_if( enumConstants.begin(),
+                                                enumConstants.end(),
+                                                []( auto const & keyval ) { return keyval.first.find( "_EXTENSION_NAME" ) != std::string::npos; } )
```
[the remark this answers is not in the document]
sharadhr: See my comment below about `starts_with` and C++20.

VulkanHppGenerator.cpp
```diff
@@ -4989,6 +4981,103 @@ std::string VulkanHppGenerator::generateConstexprDefines() const
...
+        deprecatedPrefix = replaceWithMap( deprecatedPrefixTemplate, { { "message", message } } );
+      }
+      else
+      {
+        message          = replaceWithMap( promotedExtensionMessageTemplate, { { "extensionName", name }, { "promotedTo", promotedTo } } );
+        deprecatedPrefix = replaceWithMap( deprecatedPrefixTemplate, { { "message", message } } );
+      }
+    }
+
+    auto thisExtensionConstexpr =
```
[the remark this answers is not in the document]
sharadhr: Not really. It was for debugging; I'll get rid of it.

VulkanHppGenerator.hpp
sharadhr: I feel the `std::vector` semantics are a little weird here, and sort of breaking abstraction. Also, this means I have to define a `using NameValuePair = std::pair<...>` which also feels strange. 

If you insist, I can do it, but I'm personally not so sure.

VulkanHppGenerator.cpp
```diff
@@ -4989,6 +4981,103 @@ std::string VulkanHppGenerator::generateConstexprDefines() const
...
+  static auto const deprecatedMessageTemplate      = std::string{ R"("The ${extensionName} extension has been deprecated.")" };
+  static auto const obsoletedMessageTemplate       = std::string{ R"("The ${extensionName} extension has been obsoleted by ${obsoletedBy}.")" };
+  static auto const promotedVersionMessageTemplate = std::string{ R"("The ${extensionName} extension has been promoted to core in version ${promotedTo}." )" };
+  static auto const promotedExtensionMessageTemplate = std::string{ R"("The ${extensionName} extension has been promoted to ${promotedTo}.")" };
+
+  std::string previousEnter, previousLeave;
+
+  // I really, really wish C++ had discards for structured bindings...
+  for ( auto const & [deprecatedBy, isDeprecated, name, number, obsoletedBy, platform, promotedTo, depends, requireDatas, type, xmlLine] : m_extensions )
+  {
```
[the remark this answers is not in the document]
sharadhr: I use almost all the variables, so I thought, hey, might as well expand the structured binding.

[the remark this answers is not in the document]
asuessenbach: Kind of agree...
But I think, given the number of elements and the length of the loop body, using e.g. `extension.name` or `extension.isDeprecated` makes it easier to grasp where those values come from.

VulkanHppGenerator.cpp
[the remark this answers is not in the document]
asuessenbach: Yep, VulkanHppGenerator is (currently) compiled with C++20.
Only the generated files need to be compilable with all versions since C++11.

VulkanHppGenerator.hpp
asuessenbach: I tend to keep the order of elements as they are encountered in vk.xml. That's why there are just `std::vector`s here.
But as that is definitely not important for those values here, I'm fine with a `std::map` as well.

asuessenbach:
Thanks for all your work!
When you say, it's ready to merge, I'd be happy to do so.

sharadhr:
I think it's good to go :)

PR #1725, opened 2023-11-13
Fix ArrayWrapper1D::operator std::string() and ArrayWrapper1D::operator std::string_view() to return a string/string_view of the correct length.

### LINKED ISSUE #1718: A recent change to ArrayWrapper1D may have broken string comparison in some cases
I believe this change was in pr #1698. Basically the way the cast used to work is using [this constructor](https://en.cppreference.com/w/cpp/string/basic_string/basic_string#:~:text=constexpr%20basic_string(%20const%20CharT*%20s%2C%0A%C2%A0%20%C2%A0%20%C2%A0%20%C2%A0%20%C2%A0%20%C2%A0%20%C2%A0%20%C2%A0%20%C2%A0%20%C2%A0%20%C2%A0%20%C2%A0%20const%20Allocator%26%20alloc%20%3D%20Allocator()%20)%3B) which copied the char array up to the first null character. But it was changed in response to issue #1708 which is a reasonable issue to disambiguate a constructor call. However, some things such as `ExtensionProperties` which has a member `extensionName` of type `ArrayWrapper1D<char, 256>`, when cast to a string gives a strange result which is likely not intended. That is for example, take "VK_KHR_swapchain" when cast using the new constructor creates the string "VK_KHR_swapchain\0\0\0\0... repeated until string length of 256". This breaks some string comparisons when comparing to the macro **VK_KHR_SWAPCHAIN_EXTENSION_NAME**, which is supposed to be the best way to ensure proper naming of extensions. 

Some string comparisons should still work for example casting to a c-string and using strcmp should still take into account the null character as being the end of the string. However some code that is all over the vulkan-tutorial repo `requiredExtensions.erase(extension.extensionName);` is now broken because it is internally using c++ string comparison methods. Telling users to always explicitly cast their strings how they want could be one way to fix this. Or maybe redefining the type of extensionName and other similar ArrayWrapper1D char arrays to use the c++ string class might work.  I'm not quite sure how to fix this issue or if it is considered an issue at all but I feel it is worth bringing it up and seeing what you want to do with it.

asuessenbach:
Resolves #1718.

### DIFF
--- a/snippets/ArrayWrapper1D.hpp
+++ b/snippets/ArrayWrapper1D.hpp
@@ -46,14 +46,14 @@ public:
   template <typename B = T, typename std::enable_if<std::is_same<B, char>::value, int>::type = 0>
   operator std::string() const
   {
-    return std::string( this->data(), N );
+    return std::string( this->data(), strnlen( this->data(), N ) );
   }
 
 #if 17 <= VULKAN_HPP_CPP_VERSION
   template <typename B = T, typename std::enable_if<std::is_same<B, char>::value, int>::type = 0>
   operator std::string_view() const
   {
-    return std::string_view( this->data(), N );
+    return std::string_view( this->data(), strnlen( this->data(), N ) );
   }
 #endif
 
--- a/tests/ArrayWrapper/ArrayWrapper.cpp
+++ b/tests/ArrayWrapper/ArrayWrapper.cpp
@@ -28,7 +28,7 @@ int main( int /*argc*/, char ** /*argv*/ )
 {
   vk::ArrayWrapper1D<char, 10> aw1( { 'f', 'o', 'o', 'b', 'a', 'h' } );
   std::string                  s1 = aw1;
-  assert( s1.length() == 10 );
+  assert( s1.length() == 6 );
   std::cout << "<" << aw1 << ">" << std::endl;
 
   // s1 = aw1;  // 'operator =' is ambiguous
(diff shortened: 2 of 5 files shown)


snippets/ArrayWrapper1D.hpp
```diff
@@ -46,14 +46,14 @@ class ArrayWrapper1D : public std::array<T, N>
   template <typename B = T, typename std::enable_if<std::is_same<B, char>::value, int>::type = 0>
   operator std::string() const
   {
-    return std::string( this->data(), N );
+    return std::string( this->data(), std::min( N, strlen( this->data() ) ) );
```
amaiorano, line 49: This works, but the strlen call can take a variable amount of time depending if there is no null character in the buffer. It's also UB to read beyond the buffer, which may trip UBSAN for some. How about using something like strnlen?

amaiorano: I suppose the right function to call would be strlen_s: https://en.cppreference.com/w/c/string/byte/strlen

amaiorano: Actually, scratch using strlen_s, as it's [not widely available](https://stackoverflow.com/questions/66346502/which-is-most-standard-strnlen-or-strnlen-s). I would stick to strnlen, which seems to be available in all C++ compilers.

asuessenbach: You're right, `strnlen` was the function I was looking for.
Thanks for your review!

tests/ArrayWrapper/ArrayWrapper.cpp
```diff
@@ -28,7 +28,7 @@ int main( int /*argc*/, char ** /*argv*/ )
 {
   vk::ArrayWrapper1D<char, 10> aw1( { 'f', 'o', 'o', 'b', 'a', 'h' } );
   std::string                  s1 = aw1;
-  assert( s1.length() == 10 );
+  assert( s1.length() == 6 );
```
amaiorano, line 31: For completeness, I would add 2 more tests for boundary conditions:
1. ArrayWrapper1D of size 10 with 10 non-null-terminator characters initializing it. This should return size 10.
2. ArrayWrapper1D of size 10 with 9 non-null-terminator characters and an explicit null-terminator at index 9. This should return size 9.

amaiorano:
Can we please land this change?

amaiorano:
Thank you!

PR #1729, opened 2023-11-13
Use vk::Format that matches coloredCubeData in DynamicUniform RAII sample

cutieskye:
Hi! The samples in this repository are a gem to anyone learning Vulkan.
I stumbled upon a subtle logical error.
The chosen `vk::Format` in the DynamicUniform RAII sample does not match `coloredCubeData`, which results in some information about color being discarded.
The corresponding [sample](https://github.com/KhronosGroup/Vulkan-Hpp/blob/4cb522cc511de9826b0691361ef2c786d3582b07/samples/DynamicUniform/DynamicUniform.cpp#L140) from the `vk` namespace uses the correct format.
Since these code snippets carry educational purpose, we should fix this issue to avoid confusion.


asuessenbach:
Thanks for spotting this issue. And fixing it directly!

PR #1730, opened 2023-11-15
Add missing includes

poconn:
Hi! This adds some missing includes to `vulkan_extension_inspection.hpp` - when disabling enhanced mode they're not pulled in from `vulkan.hpp`, resulting in build errors in `vulkan.cppm`. Thanks!


asuessenbach:
Thanks for spotting this issue and directly fixing it!

PR #1734, opened 2023-11-16
Add support of noexcept-path for vk::raii classes

### LINKED ISSUE #1498: Add RAII support with VULKAN_HPP_NO_EXCEPTIONS
Google / Android does not use exceptions ([see here](https://google.github.io/styleguide/cppguide.html#Exceptions)). However, the RAII classes are very useful!

Could we update the RAII generator to guard the throwing constructors for handle types using:

```cpp
#ifndef VULKAN_HPP_NO_EXCEPTIONS
  Image(...)
  {
     ...
     if (result != success)
     {
       throwResultException(...);
     }
  }
#endif
```

guard the `throwResultException()` function definition, and introduce static "create" functions that return a `std::expected<HandleType, vk::Result>` using:

```cpp
  static std::expected<Image, VULKAN_HPP_NAMESPACE::Result> create(...)
  {
    ...
    if (result != success)
    {
      return std::unexpected(result);
    }
    return Image(...);
  }
```

Bonus points if we can use custom defines such as `VULKAN_HPP_RAII_EXPECTED_CLASS` /`VULKAN_HPP_RAII_UNEXPECTED_CLASS` to use custom "expected" implementations when compiling for C++ version < 23 (for example, `android::base::expected`). If these `VULKAN_HPP_RAII_EXPECTED_CLASS` /`VULKAN_HPP_RAII_UNEXPECTED_CLASS` defines are not defined, potentially need to guard the `create()` functions based on the version. 

See https://android-review.googlesource.com/c/device/google/cuttlefish/+/2404731 for a quick attempt at this.

asuessenbach:
Resolves #1498.


README.md
```diff
@@ -889,6 +893,10 @@ By defining ```VULKAN_HPP_NO_SMART_HANDLE``` before including vulkan.hpp, the he
 
 With C++20, the so-called spaceship-operator ```<=>``` is introduced. If that operator is supported, all the structs and classes in vulkan.hpp use the default implementation of it. As currently some implementations of this operator are very slow, and others seem to be incomplete, by defining ```VULKAN_HPP_NO_SPACESHIP_OPERATOR``` before including vulkan.hpp you can remove that operator from those structs and classes.
 
+#### VULKAN_HPP_RAII_NO_EXCEPTIONS
+
+If both, VULKAN_HPP_NO_EXCEPTIONS and VULKAN_HPP_EXPECTED are defined, the vk::raii-classes don't throw exceptions. That is, the actual constructors are not available, but the creation-functions must be used.
```
theHamsta, line 898: could you add a "e.g." to clarify what creation function is referred to (`e.g. raii_instance.createDevice instead of vk::raii::Device(...)`)

asuessenbach: I've added a not to look at the vk_raii_ProgrammingGuide.md, and extended that a bit.

PR #1755, opened 2023-12-20
Check for libvulkan.1.dylib

juan-lunarg:


richard-lunarg (approved):
I've verified this fixes the issue I was seeing last interim SDK build.

theHamsta (approved):
Would agree to this PR since the same logic is used for Linux.

asuessenbach:
Thanks for spotting this issue and fixing it!

Issue #1752, opened 2023-12-18
vulkan.hpp VULKAN_HPP_ENABLE_DYNAMIC_LOADER_TOOL pollutes WinAPI again


syukosevn:
Hello, I'm using modular windows headers and vulkan.hpp defines conflict with "system" defines.
```#  elif defined( _WIN32 )
typedef struct HINSTANCE__ * HINSTANCE;
#    if defined( _WIN64 )
typedef int64_t( __stdcall * FARPROC )();
#    else
typedef int( __stdcall * FARPROC )();
#    endif
extern "C" __declspec( dllimport ) HINSTANCE __stdcall LoadLibraryA( char const * lpLibFileName );
extern "C" __declspec( dllimport ) int __stdcall FreeLibrary( HINSTANCE hLibModule );
extern "C" __declspec( dllimport ) FARPROC __stdcall GetProcAddress( HINSTANCE hModule, const char * lpProcName );
#  endif
```

Could you, please, move these lines under additional `#if` something like:
```
#ifndef VULKAN_HPP_NO_WIN32_PROTOTYPES
...
```
I assume that won't mess up any existing code. Thanks

PR #1756, opened 2023-12-22
Added VULKAN_HPP_NO_WIN32_PROTOTYPES preprocessor condition 

syukosevn:
Added `VULKAN_HPP_NO_WIN32_PROTOTYPES` preprocessor condition which disables dynamic loader win32 types declaration.

### DIFF
--- a/README.md
+++ b/README.md
@@ -893,6 +893,11 @@ By defining ```VULKAN_HPP_NO_SMART_HANDLE``` before including vulkan.hpp, the he
 
 With C++20, the so-called spaceship-operator ```<=>``` is introduced. If that operator is supported, all the structs and classes in vulkan.hpp use the default implementation of it. As currently some implementations of this operator are very slow, and others seem to be incomplete, by defining ```VULKAN_HPP_NO_SPACESHIP_OPERATOR``` before including vulkan.hpp you can remove that operator from those structs and classes.
 
+#### VULKAN_HPP_NO_WIN32_PROTOTYPES
+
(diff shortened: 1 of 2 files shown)


theHamsta:
Related issue https://github.com/KhronosGroup/Vulkan-Hpp/issues/1752

asuessenbach:
@syukosevn Would you mind to add some words about VULKAN_HPP_NO_WIN32_PROTOTYPES in the Readme.md?

syukosevn:
Added description in README

README.md
```diff
@@ -913,6 +913,11 @@ See ```VULKAN_HPP_EXPECTED```.
 
 With this define you can include a reflection mechanism on the vk-structures. It adds a function ```reflect``` that returns a tuple-version of the structure. That tuple then could easily be iterated. But at least for now, that feature takes lots of compile-time resources, so currently it is recommended to enable that feature only if you're willing to pay that price.
 
+#### VULKAN_HPP_NO_WIN32_PROTOTYPES
+
+By default, if ```DYNAMIC_LOADER_TOOL``` is enabled on Win32, ```vulkan.hpp``` declares ```HANDLE```, ```LoadLibraryA```, and other required symbols. It could cause conflicts with the ```Windows.h``` alternatives, such as ```WindowsHModular```.
```
asuessenbach, line 918: Maybe use the complete term `VULKAN_HPP_ENABLE_DYNAMIC_LOADER_TOOL` here, instead of just `DYNAMIC_LOADER_TOOL`.
Also, `HANDLE` is not declared here, but `HINSTANCE`.

asuessenbach (commented):
Please sorry for being picky... but the configuration defines are listed in alphabetical order here. Would you please move your the description of your new define at the appropriate position?

README.md
syukosevn: Done, I assume

PR #1767, opened 2024-01-05
C++20 named module documentation updates

### LINKED ISSUE #1753: Fix for Cmake for vulkan_hpp c++ modules
I believe the instructions in the README for using Cmake and C++20 modules have mistakes in them.

This is what I had to do to make Cmake happy.

```
find_package(Vulkan COMPONENTS dxc SPIRV-Tools REQUIRED)
add_library(VulkanHppModule)

target_include_directories(VulkanHppModule
    PRIVATE
        "${Vulkan_INCLUDE_DIR}"
)

target_sources(VulkanHppModule PUBLIC
  FILE_SET CXX_MODULES
  FILES ${Vulkan_INCLUDE_DIR}/vulkan/vulkan.cppm
)
target_compile_definitions(VulkanHppModule PUBLIC VULKAN_HPP_DISPATCH_LOADER_DYNAMIC=1 VULKAN_HPP_NO_STRUCT_CONSTRUCTORS) 
```

Note:
- addition of the target_include_directories
- The path for the files in the CXX_MODULE file set was wrong
- Specific to my case I believe ( I am not a Cmake Expert) I had to add the BASE_DIR because it complained about being in a different directory. I added that make code beside the definition of my renderer cmake target.
- Plus an example how we can add target_compile_definitions and VULKAN_HPP_DISPATCH_LOADER_DYNAMIC and any target that depends on VulkanHppModule will also get that define.

sharadhr:
- Fix CMake instructions for the C++20 named module
- Provided examples on module config
- Elaborated on dynamic dispatcher setup for the module

Resolves #1753.

### DIFF
--- a/README.md
+++ b/README.md
@@ -713,6 +713,13 @@ CMake provides the [FindVulkan module](https://cmake.org/cmake/help/latest/modul
 # find Vulkan SDK
 find_package( Vulkan REQUIRED )
 
+# Require Vulkan version ≥ 1.3.256 (earliest version when the Vulkan module was available)
+if( ${Vulkan_VERSION} VERSION_LESS "1.3.256" )
+  message( FATAL_ERROR "Minimum required Vulkan version for C++ modules is 1.3.256. "
+           "Found ${Vulkan_VERSION}."
+  )
+endif()
+
 # set up Vulkan C++ module as a library
 add_library( VulkanHppModule )
 target_sources( VulkanHppModule PRIVATE
--- a/README.md
+++ b/README.md
@@ -699,15 +699,15 @@ If you intend to use CMake's C++ module support (and possibly Ninja), then more 
 * CMake 3.28 or later
 * Ninja 1.10.2 or later
 
+> [!WARNING]
+> The Vulkan-Hpp C++ named module is still experimental. Some suggested ways to use it in your projects are below. The long-term goal is to submit patches to the CMake [`FindVulkan`](https://cmake.org/cmake/help/latest/module/FindVulkan.html) module so that users may transparently configure the named module, without needing to declare it as an additional library in consumer CMake code.
+
 ##### Usage with CMake
 
 CMake is recommended for use with the Vulkan-Hpp named module, as it provides a convenient platform-agnostic way to configure your project.
-CMake version 3.28 or later is required to support C++ modules.
+CMake version 3.28 or later is required to support C++ modules. Refer to the [CMake documentation](https://cmake.org/cmake/help/latest/manual/cmake-cxxmodules.7.html) on the topic.
 
 CMake provides the [FindVulkan module](https://cmake.org/cmake/help/latest/module/FindVulkan.html), which may be used to source the Vulkan SDK and Vulkan headers on your system.
-
-> [!WARNING]
-> The Vulkan-Hpp C++ named module is still experimental. Some suggested ways to use it in your projects are below. The long-term goal is to submit patches to the CMake [`FindVulkan`](https://cmake.org/cmake/help/latest/module/FindVulkan.html) module so that users may transparently configure the named module, without needing to declare it as an additional library in consumer CMake code.
 
 ```cmake
 # find Vulkan SDK
@@ -717,8 +717,8 @@ find_package( Vulkan REQUIRED )
 add_library( VulkanHppModule )
 target_sources( VulkanHppModule PRIVATE
   FILE_SET CXX_MODULES
-  BASE_DIRS ${Vulkan_INCLUDE_DIR}/vulkan
-  FILES ${Vulkan_INCLUDE_DIR}/vulkan.cppm
+  BASE_DIRS ${Vulkan_INCLUDE_DIR}
+  FILES ${Vulkan_INCLUDE_DIR}/vulkan/vulkan.cppm
 )
 target_link_libraries( VulkanHppModule PUBLIC Vulkan::Vulkan )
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -695,41 +693,62 @@ This feature requires a recent compiler with complete C++20 support:
...
-# set up Vulkan C++ module
-add_library(VulkanCppModule)
-target_sources(VulkanCppModule PRIVATE
+# set up Vulkan C++ module as a library
+add_library( VulkanHppModule )
+target_sources( VulkanHppModule PRIVATE
   FILE_SET CXX_MODULES
+  BASE_DIRS ${Vulkan_INCLUDE_DIR}/vulkan
   FILES ${Vulkan_INCLUDE_DIR}/vulkan.cppm
 )
```
monamimani, line 722: Does this really work?
I think for me what really worked was
```
target_sources(VulkanHppModule PUBLIC
...
  **BASE_DIRS ${Vulkan_INCLUDE_DIR} 
  FILES ${Vulkan_INCLUDE_DIR}/vulkan/vulkan.cppm**
)```

Because the cppm file is in the vulkan folder.

**Review by monamimani — COMMENTED** [review]
Beside that little comment everything looks fine. :)

File: README.md (hunk 1 above)
Comment (sharadhr) [reply README.md:722]: I just tested it—both work fine. Agreed that just using `${Vulkan_INCLUDE_DIR} is simpler, though...

Comment (monamimani) [reply README.md:722]: For me that line
`FILES ${Vulkan_INCLUDE_DIR}/vulkan.cppm`

needs to be
`FILES ${Vulkan_INCLUDE_DIR}/vulkan/vulkan.cppm`

if not I get this error

> [cmake] CMake Error at CMakeLists.txt:20 (target_sources):
[cmake]   Cannot find source file:
[cmake] 
[cmake]     C:/VulkanSDK/1.3.268.0/Include/vulkan.cppm
[cmake] 
[cmake] 
[cmake] CMake Error at CMakeLists.txt:18 (add_library):
[cmake]   No SOURCES given to target: VulkanHppModule

As you can see I am using the SDK (1.3.268.0) on windows . My Vulkan module is in `C:/VulkanSDK/1.3.268.0/Include/vulkan/vulkan.cppm`

What is your setup?

Comment (sharadhr) [reply README.md:722]: Huh, that was a typo, and I changed the wrong line to boot. Proof that I shouldn't make commits at 2am.  

I also have `FILES ${Vulkan_INCLUDE_DIR}/vulkan/vulkan.cppm`; I'll change this right away.

**Comment by monamimani** [comment]
No worries :)
I also added for good measure
`set(CMAKE_CXX_STANDARD 20)`

and I also needed this on my side to make things compile if not you can't include the macros header and the compilation of the cppm fails.
```
target_include_directories(VulkanHppModule
PUBLIC
"${Vulkan_INCLUDE_DIR}"
)
```

**Comment by sharadhr** [comment]
> and I also needed this on my side to make things compile if not you can't include the macros header and the compilation of the cppm fails.
> 
> ```
> target_include_directories(VulkanHppModule
> PUBLIC
> "${Vulkan_INCLUDE_DIR}"
> )
> ```

Hmm, I didn't need that. Linking to `Vulkan::Vulkan` should be sufficient, as it [asks CMake to include the Vulkan header directory, too](https://gitlab.kitware.com/cmake/cmake/-/blob/master/Modules/FindVulkan.cmake#L610).

**Comment by monamimani** [comment]
You are right, Vulkan::Vulkan was linking on an other target. Linking it to `VulkanHppModule` make it so we don't need the `target_include_directories`. 
So all is good to go I think.

**Review by monamimani — APPROVED** [review]
All good!

File: README.md [hunk 2]
```diff
@@ -695,41 +693,69 @@ This feature requires a recent compiler with complete C++20 support:
...
+CMake version 3.28 or later is required to support C++ modules. Refer to the [CMake documentation](https://cmake.org/cmake/help/latest/manual/cmake-cxxmodules.7.html) on the topic.
 
 CMake provides the [FindVulkan module](https://cmake.org/cmake/help/latest/module/FindVulkan.html), which may be used to source the Vulkan SDK and Vulkan headers on your system.
-**Note that this module does not yet provide an IMPORTED target for the Vulkan C++ module, so you must set it up manually.**
-
-To use CMake with C++ modules, you must first set up `vulkan.cppm` as the source for a library target with `FILE_SET` configured  `TYPE = CXX_MODULES`, and then link it into your project.
 
 ```cmake
 # find Vulkan SDK
 find_package( Vulkan REQUIRED )
```
theHamsta, line 714: maybe we could add version requirement for the vulkan sdk that is sufficiently new to have the `*.cppm` file? Or is this too much maintenance overhead to make sure the version is sufficient for all features shown below? Probably it makes sense for a project to always require a certain version of the Vulkan SDK depending on the features they are using.

sharadhr: > Probably it makes sense for a project to always require a certain version of the Vulkan SDK

Agreed.

If there's a version requirement setup, I feel this is stuff that ought to go in `vulkan_core.h` rather than here. Bunch of `#ifdef`s and `assert()`s should suffice.

monamimani: @theHamsta, you could probably use the defines `VK_HEADER_VERSION` or `VK_HEADER_VERSION_COMPLETE`. Since VulkanHpp is generated from the spec it is deployed in lock step with vulkan version in the SDK.

If you don't use the SDK and build it up your self them most likely you are already getting a specific git release or git hash, so you know if you have it already.

And if you are at top of tree well you know also.

That should cover it I think.

theHamsta: But in the sample the `.cppm` from a system installation of Vulkan is used. You want then to find one that is new enough for the requirements of your project and also one that has `.cppm`

sharadhr: > you could probably use the defines VK_HEADER_VERSION or VK_HEADER_VERSION_COMPLETE

@monamimani, with recent changes to Vulkan-Hpp, you could use `vk::HeaderVersion` 😉 

https://github.com/KhronosGroup/Vulkan-Hpp/blob/890e32f27b590acd64624141b7d554bf9881c4a0/vulkan/vulkan.hpp#L6830

@theHamsta, I'll add the version check in, no big deal. The sample is in a way meant to be copy-pasted, and it'd be good if the minimum example worked regardless of the version, or at least errored with an appropriate message.

monamimani: > @monamimani, with recent changes to Vulkan-Hpp, you could use vk::HeaderVersion 😉

I am on my phone, I assume you are referring to the constexpr variable. I think no ask was released with that change yet.

asuessenbach:
@sharadhr and @monamimani Thanks a lot for your great support here! The result looks good.
If you both agree to the current state of this PR, I will merge it.

sharadhr:
> @sharadhr and @monamimani Thanks a lot for your great support here! The result looks good. If you both agree to the current state of this PR, I will merge it.

LGTM!

monamimani:
> @sharadhr and @monamimani Thanks a lot for your great support here! The result looks good.
> If you both agree to the current state of this PR, I will merge it.

Yep let's go 😃

PR #1779, opened 2024-02-02
Add support of attribute "len" for array-sized structure members.

### LINKED ISSUE #1777: build failure with null-terminated static array members
In internal MR https://gitlab.khronos.org/vulkan/vulkan/-/merge_requests/6448 I'm seeing CI failure in the hpp-generate stage. I'm not sure, but this is probably because the point of the MR is to allow 'len' attributes to be used and well-defined WRT struct members that are arrays of static size, including 'len="null-terminated"' for static arrays containing strings (as quite a few do). We'd like to get this merged within a week or two, so if you're able to resolve this, that would help.

asuessenbach:
Resolves #1777.

### DIFF
--- a/VulkanHppGenerator.cpp
+++ b/VulkanHppGenerator.cpp
@@ -7969,8 +7969,9 @@ std::string VulkanHppGenerator::generateLenInitializer(
     assert( ( arrayIt->lenExpressions.front() == litit->first->name ) ||
             ( ( arrayIt->lenExpressions.front() == "codeSize / 4" ) && ( litit->first->name == "codeSize" ) ) );
 
-    assert( arrayIt->name.starts_with( "p" ) );
-    std::string argumentName = startLowerCase( stripPrefix( arrayIt->name, "p" ) ) + "_";
+    assert( arrayIt->type.isPointer() || !arrayIt->arraySizes.empty() );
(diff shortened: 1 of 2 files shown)


oddhack:
I notice your CI job is taking quite a long time - is this because the jobs are intrinsically lengthy, or are you running into Actions limitations? If it's an Actions issue, please chime on on the internal khronos-general/khronos-issues/-/issues/127 where we're discussing concerns about Actions as currently configured, and how that's impacting our workflows.

asuessenbach:
@oddhack Thanks for your question on my CI jobs.
Actually, they **are** long. For each CI job, I'm running a generation and compile test with five C++ version (11, 14, 17, 20, 23) in debug and release resulting in 10 build-and-run steps. That takes a while, but seems to be faster than running 10 separate CI jobs, which took ages 'til they were scheduled to run.
Would you suggest a different approach?

oddhack:
We're talking about issues people have with Actions, and things we might do about it, on the internal issue I linked above - so good to get your experiences / needs raised there. James is in discussion with Github staff about our options as well. Anything we do to substantially improve matters will cost money, possibly a lot of money, but Actions is starting to become a blocker in significant ways as we are currently using it.

PR #1788, opened 2024-02-05
allow len="1" from param in parsing vk.xml

### LINKED ISSUE #1787: Allow len="1" in parameter attributes
Internal MR 6453 is getting this hpp-generate error in CI:

> caught exception: VulkanHppGenerator: Spec error on line 11712: attribute <len> holds an unknown value <1>

This is defined as a legal value in the registry schema document, though has only previously been used for nested pointers - in this case it just means the command always returns a single instance of the pointed-to structure. Note this is different than the outstanding internal MR 6448 although both touch on the 'len' attribute.

PanGao-h:
In registry.rnc, the description of len in <param> allows special values such as "null-terminated" or "1". But somehow the value "1" is missed.

```
#   <param> are function parameters, in order
#     len - if the member is an array, len may be one or more of the following
#           things, separated by commas (one for each array indirection):
#           another member of that struct, 'null-terminated' for a string,
#           '1' to indicate it is just a pointer (used for nested pointers),
#           or a latex equation (prefixed with 'latexmath:')
```

@oddhack added:

Fixes #1787

### DIFF
--- a/VulkanHppGenerator.cpp
+++ b/VulkanHppGenerator.cpp
@@ -12813,7 +12813,7 @@ std::pair<bool, VulkanHppGenerator::ParamData> VulkanHppGenerator::readCommandPa
         }
         else
         {
-          checkForError( ( attribute.second == "null-terminated" ) || isLenByStructMember( attribute.second, params ),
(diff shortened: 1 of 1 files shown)


asuessenbach:
@PanGao-h adding support for the value "1" for the len attribute of command arguments takes a little more than just this check (see #1789).
If you're ok with it, I'll just take that MR and close yours?

PR #1803, opened 2024-02-13
Disabled `vk::raii::exchange()` implementation for C++14 or newer

syukosevn:
Replaced with `using std::exchange`

```
namespace sample
{
    using std::exchange;

    void func()
    {
        std::vector<vk::raii::CommandBuffer> bufs = ...;
        exchange(bufs, {}); //< that line causes 'ambiguous call' compiler error
    }
}

```

### DIFF
--- a/VulkanHppGenerator.cpp
+++ b/VulkanHppGenerator.cpp
@@ -578,17 +578,17 @@ namespace VULKAN_HPP_NAMESPACE
 {
   namespace VULKAN_HPP_RAII_NAMESPACE
   {
+#  if ( 14 <= VULKAN_HPP_CPP_VERSION )
+    using std::exchange;
+#  else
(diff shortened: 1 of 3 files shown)


asuessenbach:
This seems to be a reasonable modification. Thanks a lot for your contribution.
Would you mind to add the generated headers as well and see if the CI keeps happy?

syukosevn:
I'm not sure what I did.
Compiled VulkanHppGenerator, ran `VHG -api vulkan` and `-api vulkansc`, got several formatting diffs in `vulkan_funcs.hpp`, so I reverted them.
Not sure if I have to update hpp's in `Vulkan-Headers/include/vulkan/vulkan_raii.hpp`

asuessenbach:

PR #1806, opened 2024-02-18
Fix for VULKAN_HPP_TYPESAFE_CONVERSION defined to 0

### LINKED ISSUE #1801: VULKAN_HPP_TYPESAFE_CONVERSION cannot be set to 0 to disable implicit conversions
The manual currently states the following at https://github.com/KhronosGroup/Vulkan-Hpp?tab=readme-ov-file#cc-interop-for-handles:

> On 64-bit platforms this define is set to 1 by default and can be set to 0 to disable implicit conversions.

The headers use
`#if defined(VULKAN_HPP_TYPESAFE_CONVERSION)`,

which will return true regardless of what VULKAN_HPP_TYPESAFE_CONVERSION is set to as long as it's defined. Defining an identifier as 0 does not undefine it.

This needs to be changed to

`#if (VULKAN_HPP_TYPESAFE_CONVERSION == 1)`. 

Similarly, the code that defines it if undefined needs to be changed.

I can prepare a PR if desired.

d235j:
This fixes #1801. Headers will have to be regenerated/copied.


asuessenbach:
I think, you don't need to check if `VULKAN_HPP_TYPESAFE_CONVERSION` is defined _and_ compare it with 1. The pure comparison should suffice, as undefined values are considered to be 0.
Besides that, would be great, if you could add the generated headers as well, as the CI would then check if all (used) compilers agree about the correctness of this statement.

d235j:
> I think, you don't need to check if VULKAN_HPP_TYPESAFE_CONVERSION is defined and compare it with 1. The pure comparison should suffice, as undefined values are considered to be 0.

This will trip up `-Wundef` on some compilers or `/wd4668` on MSVC. This warning is not included in `-Wall`; if it is not a concern then I can change the PR.

> Besides that, would be great, if you could add the generated headers as well, as the CI would then check if all (used) compilers agree about the correctness of this statement.

Is there a contribution guide available that indicates what steps should be taken before making a PR? I noticed many whitespace differences between the headers in the repository and those I generated (and I have `clang-format` installed).

asuessenbach:
> This will trip up -Wundef on some compilers or /wd4668 on MSVC. This warning is not included in -Wall; if it is not a concern then I can change the PR.

I think, those warnings are not a concern here. So please change the PR.

> Is there a contribution guide available that indicates what steps should be taken before making a PR? I noticed many whitespace differences between the headers in the repository and those I generated (and I have clang-format installed).

I'm sorry, there's nothing like a contribution guide. Maybe I should introduce one, at some point.
Regarding clang-format: I think, currently clang-format version 14 is used. So, if you could try that one?
And even if there are whitespace differences: would be great to have the generated files in this PR to let the CI check if all (tested) compilers are happy with that change.

Otherwise, if you're ok with that, I could hijack your PR, make it on my own and check the results.

d235j:
Since I left the “Allow edits from maintainers” checked you should be able to commit and push to my PR (this is a newer GitHub feature). Feel free to utilize it. Otherwise I can update the PR later. As for clang-format, I believe I was using the latest released version (17) on macOS.
 
Might it be possible to implement a CI process where the CI detects if the headers need updating and automatically updates them?

asuessenbach:
@d235j I've adjusted your PR a little. If you are ok with those changes, I'd merge it.

> Might it be possible to implement a CI process where the CI detects if the headers need updating and automatically updates them?

The detection step should be possible, but I don't see/know how to automatically update them.

PR #1827, opened 2024-03-17
Add missing glslang headers

ShabbyX:


asuessenbach:
What compiler or what platform does need that header to be included?

ShabbyX:
Reproduces for me with this cmake config on Linux:

```
cmake -DVULKAN_HPP_SAMPLES_BUILD=ON -DVULKAN_HPP_SAMPLES_BUILD_ONLY_DYNAMIC=ON -DVULKAN_HPP_TESTS_BUILD=ON -DVULKAN_HPP_TESTS_BUILD_ONLY_DYNAMIC=ON -DVULKAN_HPP_BUILD_WITH_LOCAL_VULKAN_HPP=ON -DVULKAN_HPP_PRECOMPILE=OFF -DALLOW_EXTERNAL_SPIRV_TOOLS=ON
```

I believe it found `g++` as the compiler (version 13.2.0). Regardless, the files here use `glslang::InitializeProcess()` and such, which is defined in `ShaderLang.h`. The files include `SPIRV/GlslangToSpv.h` from glslang, but that file doesn't include `ShaderLang.h`.

I'm not really sure how it's been compiling without these includes!

asuessenbach:
> The files include `SPIRV/GlslangToSpv.h` from glslang, but that file doesn't include ShaderLang.h

Well, `GlslangToSpv.h` includes `glslang/Include/intermediate.h`, which includes `../Include/Types.h`, which finally includes `../Public/ShaderLang.h`.
That is, it should compile fine on your end as well!

ShabbyX:
> Well, GlslangToSpv.h includes glslang/Include/intermediate.h

I see now, that's no longer true on ToT glslang: https://github.com/KhronosGroup/glslang/blob/main/SPIRV/GlslangToSpv.h

The version in Vulkan-Hpp's submodule does have that include: https://github.com/KhronosGroup/glslang/blob/f1cb8608b390a7f51b4ae0d62cd415ba47a59b86/SPIRV/GlslangToSpv.h

So you'd want this change next time you roll glslang.

To provide more context on why I'm building with ToT glslang, I'm working on a CI script for Vulkan to notify of potential issues before new Vulkan spec versions are released, as well as support pre-release development of various repos while an extension is in-flight. The script necessarily has to work with ToT or close to it for this work.

asuessenbach:
> I see now, that's no longer true on ToT glslang: https://github.com/KhronosGroup/glslang/blob/main/SPIRV/GlslangToSpv.h

Bummer!
So, thanks for fixing this future issue pro-actively!

PR #1829, opened 2024-03-19
Extent type trait vk::CppType to map from VkHandle to vk::Handle

asuessenbach:


VulkanHppGenerator.cpp
```diff
@@ -7661,6 +7662,12 @@ std::string VulkanHppGenerator::generateHandle( std::pair<std::string, HandleDat
 
 ${CppType}
 
+  template <>
+  struct CppType<Vk${className}, VK_NULL_HANDLE>
```
dneto0, line 7666: On 32-bit builds non-dispatchable handles boil down to a 64bit int type.
So these end up as conflicting templates, which causes compilation failure.

PR #1844, opened 2024-04-09
Fixed Validation Error From Zero MaxImageCount in VkSurfaceCapabilitiesKHR

ahcox:
Correctly handle a zero max image count in VkSurfaceCapabilitiesKHR which indicates no upper limit to the number of images in a swapchain.
_"maxImageCount is the maximum number of images the specified device supports for a swapchain created for the surface, and will be either 0, or greater than or equal to minImageCount. A value of 0 means that there is no limit on the number of images, though there may be limits related to the total amount of memory used by presentable images."_
https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkSurfaceCapabilitiesKHR.html

Fixes validation error VUID-VkSwapchainCreateInfoKHR-minImageCount-01271


asuessenbach:
Good catch!
Thanks for reporting this issue and fixing it straight away.

PR #1856, opened 2024-04-27
Check value of `VULKAN_HPP_ENABLE_DYNAMIC_LOADER_TOOL` instead of existence

### LINKED ISSUE #1855: VULKAN_HPP_ENABLE_DYNAMIC_LOADER_TOOL and vulkan.cppm don't mix
Everywhere else where `VULKAN_HPP_ENABLE_DYNAMIC_LOADER_TOOL` is checked, it is checked against the value it expands to, e.g.:

```
#if VULKAN_HPP_ENABLE_DYNAMIC_LOADER_TOOL
...
#endif
```

See for example how the DynamicLoader class itself is protected:

https://github.com/KhronosGroup/Vulkan-Hpp/blob/1de6e89eb29bdb019eed2733fa7ed9fa72da08fa/vulkan/vulkan.hpp#L16677-L16679

This is because the value is forced to be defined to 1 if it isn't already set:
https://github.com/KhronosGroup/Vulkan-Hpp/blob/1de6e89eb29bdb019eed2733fa7ed9fa72da08fa/vulkan/vulkan_hpp_macros.hpp#L74-L76

However, the modules currently check whether it is defined at all:

https://github.com/KhronosGroup/Vulkan-Hpp/blob/1de6e89eb29bdb019eed2733fa7ed9fa72da08fa/vulkan/vulkan.cppm#L4678
https://github.com/KhronosGroup/Vulkan-Hpp/blob/1de6e89eb29bdb019eed2733fa7ed9fa72da08fa/vulkan/vulkansc.cppm#L1706

The immediate issue is that this should check the value, not whether it is defined.

The meta issue is that the whole "enable" system of vulkan hpp seems really weird. I shouldn't have to define `ENABLE_X=0` to disable feature X at all, the lack of a define should be enough. If it is default on, then provide a `DISABLE_X` macro. It really isn't surprising that this bug slipped through with the really unintuitive way these ENABLE macros are set up.

My ultimate goal is that I want to use vk::raii::Context with my own getInstanceProcAddr pointer since I already have it. It seems a bit weird to me that the constructor doesn't always exist, but I can see a train of thought that might justify that decision.

sharadhr:
Resolves #1855


asuessenbach:
Thanks for fixing this issue that fast!

PR #1874, opened 2024-05-12
RAII Sample 16 fix

tomatih:
Fixed a small typo where, in the RAII API versioning sample, the instance was being created without the version parameter.


cutieskye (approved):
Good catch - this line should correspond to https://github.com/KhronosGroup/Vulkan-Hpp/blob/b2810ca912ec8478a34f121b77f7899a9ca30675/samples/16_Vulkan_1_1/16_Vulkan_1_1.cpp#L60

theHamsta (approved):
Thank you!

PR #1923, opened 2024-07-18
Add move semantics for non-raii hpp handles. (#1919)

### LINKED ISSUE #1919: Move semantics for non-raii handles
For example: I'm using `vk::Buffer` in the standard layout structure(it shouldn't have any constructors, including move-constructor) so I can actively use designated initializers. 
Something like this:
```
struct VertexBuffer {
  vk::Buffer buffer;
  
  ~VertexBuffer() { vkDestroyBuffer(...); }
}
```
Of course `vkDestroyBuffer` is getting called during `emplace_back(std::move(vbuff))` with valid buffer handle.
Wouldn't it be more intuitive to implement move semantics for handles with something like this?
```
Buffer(Buffer&& rhs) {
  : m_buffer(exchange(rhs.m_buffer, {}))
{
}
```

Would you accept a PR with these changes?

syukosevn:
Added define `VULKAN_HPP_HANDLES_MOVE_EXCHANGE` which enables `exchange` in handles' move-constructors.
Moved `exchange` from `raii` to root `vk` namespace.

closes #1919


syukosevn:
Had issues with codegen: clang-format makes too much changes to existing headers. I suppose it's because of newer version(16 or 18).
Tell me if I have to install clang-format 15 for correct behavior.

asuessenbach:
Great PR, thank you!

Would be perfect if you could use clang-format 15 and also add the generated files, but I would also accept your change as is...
But please adjust the year in the copyright header in the two new files, and the comment about ArrayProxy in the test. Also, I don't get why you check agains nullptr there in two directions?

syukosevn:
Two directions `nullptr` check: it was a copy-paste from Handles test. I will remove redundant checks.

syukosevn:
Successfully reproduced build env. Will update PR after 1.3.291 headers pipeline completion

asuessenbach:
Looks great now, thanks a lot for your contribution.

PR #1931, opened 2024-07-23
Don't export symbols in `vk::detail` namespace

### LINKED ISSUE #1928: [c++20 module] functions from vk::detail:: pulled into vk::
`vulkan.cppm` uses `namespace vk { ... using vk::detail::{createResultValue, ignore, resultCheck} ... }`, so those symbols are present in the `vk::` namespace instead of the original `vk::detail::`

sharadhr:
Resolves #1928.

### DIFF
--- a/VulkanHppGenerator.cpp
+++ b/VulkanHppGenerator.cpp
@@ -5864,7 +5864,7 @@ std::string VulkanHppGenerator::generateCppModuleUsings() const
 
   // some hardcoded types
   auto const hardCodedResultValueTypes =
-    std::array{ "ResultValue", "ResultValueType", "detail::createResultValueType", "detail::ignore", "detail::resultCheck" };
+    std::array{ "ResultValue", "ResultValueType" };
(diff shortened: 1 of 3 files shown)


asuessenbach:
@sharadhr This PR just removed the symbols from the vk::detail namespace.
Would it be possible to export them in the vk::detail namespace, instead?

sharadhr:
I'm sure that's possible, but do we really want to expose the `detail` stuff? I was under the impression the point of `detail` was to tell users that those symbols are not meant to be part of the public user-facing API, and not exporting them in the C++ module makes this very explicit.

asuessenbach:
You're kind of right, that stuff is part of the detail namespace and thus not part of the "official" vulkan.
But it seems, at least some guys want to use at least some of those functions. And as they are in the detail-namespace they (should) know, that they could be changed without further notice.
That is, you can use them now (on your own risk) when you don't use modules, but you can't use them with modules. Which might be worth to be changed.

PR #1939, opened 2024-08-07
Add IWYU pragmas to let code-completers use vulkan.hpp.

### LINKED ISSUE #1936: IYWU pragmas for vulkan_{enums,funcs,structs}.hpp
As my code completer I am using clangd. Clangd automatically includes the relevant header when completing a function/struct/class. When code-completing a Vulkan-Hpp enum, it will include the header `<vulkan/vulkan_enum.hpp>`. Similarly for structs.

The headers `vulkan_enums.hpp`, `vulkan_funcs.hpp` and `vulkan_structs.hpp` are not meant to be included standalone from what I understand. They are meant to be included as part of including `vulkan.hpp`. When these headers are included standalone, a compilation error is the result.

Clangd respects the include-what-you-use pragmas: https://clangd.llvm.org/guides/include-cleaner#iwyu-pragmas

I believe what is needed to make clangd (and also the include-what-you-use tool) insert the correct header completion is to add a line like this to the three files:

```cpp
// IWYU pragma: private; include "vulkan.hpp"
```

See also: https://github.com/include-what-you-use/include-what-you-use/blob/master/docs/IWYUPragmas.md#iwyu-pragma-private

asuessenbach:
Resolves #1936.

### DIFF
--- a/VulkanHppGenerator.cpp
+++ b/VulkanHppGenerator.cpp
@@ -99,6 +99,9 @@ void VulkanHppGenerator::generateEnumsHppFile() const
 #ifndef VULKAN_ENUMS_HPP
 #  define VULKAN_ENUMS_HPP
 
+// include-what-you-use: make sure, vulkan.hpp is used by code-completers
+// IWYU pragma: private; include "vulkan.hpp"
+
 namespace VULKAN_HPP_NAMESPACE
 {
 ${Flags}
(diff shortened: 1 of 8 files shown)


VulkanHppGenerator.cpp
```diff
@@ -99,6 +99,9 @@ void VulkanHppGenerator::generateEnumsHppFile() const
 #ifndef VULKAN_ENUMS_HPP
 #  define VULKAN_ENUMS_HPP
 
+// include-what-you-use: make sure, vulkan.hpp is used by code-completers
+// IWYU pragma: private; include "vulkan.hpp"
```
rwols, line 103: I finally had some time to really dig into this :D.

This line doesn't work. It uses a semicolon whereas it should use a comma. When the comma is used, clangd (and the IWYU tool, presumably) starts to pick up this pragma directive.

.

The directive is now to include "vulkan.hpp". This makes clangd (and the IWYU tool, presumably) insert the line `#include "vulkan.hpp"`. Is this correct? From my understanding, the canonical include (`-I`) directory should be such that `#include "vulkan/vulkan.hpp"` is the correct include line. I think this should say `include "vulkan/vulkan.hpp"` instead of `include "vulkan.hpp"`.

PR #1978, opened 2024-10-23
Add some special handling debug code for vk::raii::DescriptorPool

asuessenbach:


stripe2933:
I think this change should be reverted.

This assertion is not in the spec, and user should be able to choose how descriptor set's ownership to be managed.

It is more intuitive to think `VkDescriptorSet` and `VkCommandBuffer` ownership is tied to their corresponding `VkDescriptorPool` and `VkCommandPool`, and the assumption of the former can be rethought by specifying `VK_DESCRIPTOR_POOL_CREATE_FREE_DESCRIPTOR_SET_BIT` flag. This should be an option, not a forced assertion.

> Another workaround is to create the descriptor sets without the RAII capabilities (vk::raii::DescriptorSet has pretty limited capabilities anyway):
> 
> ```c++
> std::vector<vk::DescriptorSet> descriptorSets = (*device).allocateDescriptorSets(vk::DescriptorSetAllocateInfo(*descriptorPool, ...));
> ```
> 
> This works well when `VK_DESCRIPTOR_POOL_CREATE_FREE_DESCRIPTOR_SET_BIT` is not set, which is often the case. If it is set, the RAII descriptor set class can be used, and care must only be taken that the descriptor sets' destructors are called before the pool's.

As the code like above (from @jnhyatt), doesn't have to set the free descriptor set flag bit.

asuessenbach:
@stripe2933 You're right, this assertion is wrong.  I'll remove it again.

The "correct" check would be in the vk::raii::DescriptorSet constructor. Unfortunately, there is no way to get the required information out of what we have there.

So, thanks a lot for spotting this issue.
But it would have been better if you actually had filed an issue, instead of commenting on an already merged PR. Such a comment could easily be missed.

stripe2933:
> But it would have been better if you actually had filed an issue, instead of commenting on an already merged PR. Such a comment could easily be missed.

I didn't know it is not notified. Anyway, thank you!

PR #1983, opened 2024-10-28
Move all the dispatcher related classes into namespace detail

asuessenbach:


maxime-modulopi:
Hello @asuessenbach, what was the motivation behind this change?
This seems to go against recommended usage (see #2111).

PR #1988, opened 2024-11-01
Undefine major and minor through <sys/sysmacros.h>

gulfemsavrun:
<sys/sysmacros.h> is included through some other headers. This results in major(x) being resolved to gnu_dev_major(x) which is an expression in a constructor initializer list. This patch includes <sys/sysmacros.h> at the very beginning, and undefines major and minor to avoid them to be resolved to gnu_dev_major(x) or gnu_dev_minor(x).

### DIFF
--- a/vulkan/vulkan.hpp
+++ b/vulkan/vulkan.hpp
@@ -8,6 +8,13 @@
 #ifndef VULKAN_HPP
 #define VULKAN_HPP
 
+// <sys/sysmacros.h> is included through some other headers, and
+// this results in major(x) being resolved to gnu_dev_major(x)
+// which is an expression in a constructor initializer list.
+#include <sys/sysmacros.h>
+#undef major
+#undef minor
+
 #include <vulkan/vulkan_hpp_macros.hpp>
 
 #if defined( VULKAN_HPP_ENABLE_STD_MODULE ) && defined( VULKAN_HPP_STD_MODULE )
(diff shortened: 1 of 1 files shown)


asuessenbach:
I think, `sys/sysmacros.h` is not available on every platform supported by Vulkan-Hpp. Therefore, you can't just include it.

Essentially, you just moved the undef of `major` and `minor` a few lines up... where exactly would you need that, and why?

And, by the way, vulkan.hpp is a generated file, you can't modify that file directly. Changes of that kind should be done to the appropriate files in `Vulkan-Hpp/snippets`, maybe `snippets/defines.hpp`.

gulfemsavrun:
@asuessenbach thanks for your response. I'm not familiar this codebase, and we currently have an issue in our project, where we use `Vulkan-Headers` as `third_party` code. The issue is that the line at  `third_party/Vulkan-Headers/include/vulkan/vulkan.hpp:23` has `include <vector>`. We use `libc++` from Clang in our project. There is an upstream change in `libc++`,  where it stopped transitively including `sys/sysmacros.h` for `include <vector>`.
https://github.com/llvm/llvm-project/pull/99705

This started resulting in major(x) being resolved to gnu_dev_major(x). This seems to be an known issue for `include <tuple>` as explained in the comments. I was trying to extend this to any header that might transitively include `sys/sysmacros.h`. What is the proper way of fixing this issue in this code base?

asuessenbach:
> This started resulting in major(x) being resolved to gnu_dev_major(x)

I still don't get, which `major(x)` is wrongly resolved here. Where does it come from. And why doesn't the `undef` of `major` and `minor` that is already there does not help?

And: I think, just unconditionally including `sysmacros.h` here is not an option.

asuessenbach:
Closing this, assuming it has been resolved elsewhere.

PR #1989, opened 2024-11-04
Switch CI on MacOS from macos-13 to macos-14, using just one compiler

asuessenbach:


.github/workflows/ci-macos.yml
```diff
@@ -11,9 +11,9 @@ jobs:
     
     strategy:
       matrix:
-        os: [macos-13]
+        os: [macos-13, macos-14-large]
```
spencer-lunarg, line 14: Currently we have been seeing a spike in start up time for CI to run across many KhronosGroup Repos (sometimes in terms of hours to start)

I have been going around all KhronosGroup repos and been tracking down those running on MacOS as those cost 10x the minutes towards are github actions minutes (which we have an internal issue tracked here https://gitlab.khronos.org/khronos-general/khronos-issues/-/issues/127)

I notice from previous runs of Vulkan-HPP that the 3 current MacOS builds take a total of 4 hours across that 3 builds, that is costing 40 hours of CI minutes. This change will be almost be tripling it and highly suggest we do  not accept this change and limit the MacOS runs to what is absolutely required in the short term
https://github.com/KhronosGroup/Vulkan-Hpp/actions/runs/11608159059

cc @outofcontrol @marty-johnson59

asuessenbach: @spencer-lunarg Ok, I can reduce the MacOS builds again to just one runner, no problem.
Would you suggest to keep macos-13, or switch to macos-14-large? Is there a perf difference to be expected?
From my point of view, it probably doesn't matter, as both runners support the very same set of compilers.

spencer-lunarg: personally I would just `macos-latest`

spencer-lunarg: as for the `[clang++, g++-12, g++-13, g++-14]` I guess I would just use the default version in `macos-latest` 

another thing to think about is adding `if: github.ref == 'refs/heads/main'` so things only go after things are merged

I guess not sure about this repo, but in VVL, after 1 year, I have yet to find a single time where ONLY the MacOS job failed, but everything else passed

spencer-lunarg: I was just told that gcc just points to clang in Mac

Looking at your [https://github.com/KhronosGroup/Vulkan-Hpp/actions/runs/11680120614/job/32522471595](gcc++) you can see it is just `clang` using a toolchain

image

and my other question is why someone on a Mac would opt to use gcc and not clang since Apple uses clang for everything to my knowledge

asuessenbach: > I was just told that gcc just points to clang in Mac

That's funny!!
Will reduce to just use clang here. Makes just one release and one debug build on Mac.

spencer-lunarg (approved):
Thanks for being understanding here

PR #1997, opened 2024-11-22
fix: allow aliases in enums of video.xml

theHamsta:
The spec update to 1.3.302 introduces the `alias` attribute for enums in video.xml in the form of the deprecation of `STD_VIDEO_AV1_COLOR_PRIMARIES_BT_UNSPECIFIED`. See https://github.com/KhronosGroup/Vulkan-Headers/blob/36872f9062b17b1a30b8ed1d81ca5ea6bb608a72/registry/video.xml#L1503

This change tries to handle the `alias` attribute in the most minimally invasive manner to unblock update to 1.3.302. See failing CI pipeline https://github.com/KhronosGroup/Vulkan-Hpp/actions/runs/11963893922/job/33355215029

This is not a proper fix since enum handling of VideoHppGenerator and VideoHppGenerator should properly deduplicated.

### DIFF
--- a/VideoHppGenerator.cpp
+++ b/VideoHppGenerator.cpp
@@ -446,10 +446,10 @@ void VideoHppGenerator::readEnumsEnum( tinyxml2::XMLElement const * element, std
 {
   int                                line       = element->GetLineNum();
   std::map<std::string, std::string> attributes = getAttributes( element );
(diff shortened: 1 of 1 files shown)


theHamsta:
Superseded by https://github.com/KhronosGroup/Vulkan-Hpp/pull/1999

PR #2025, opened 2024-12-13
Add missing include headers

atetubou:
This is to fix build error when we set use_libcxx_modules=true in chromium build.


asuessenbach:
Note, that the headers you modified are generated. You would need to adjust the generating code, instead.

What errors do you get on that platform?

asuessenbach:
Closing this, assuming it has been resolved otherwise.

PR #2040, opened 2025-01-09
Correct the name of two flags in the readme.

asuessenbach:


M2-TE:
I think you should add `VULKAN_HPP_NO_UNION_CONSTRUCTORS` to the readme as well. It seems the old flag `VULKAN_HPP_NO_CONSTRUCTORS` included union constructors and only setting `VULKAN_HPP_NO_STRUCT_CONSTRUCTORS` breaks designated initializers for unions, at least for `vk::ClearValue` in my case

asuessenbach:
Ah, I see. Thanks again for that hint.
That is, I'll adjust the source to the documentation, instead :)

PR #2047, opened 2025-01-20
CMake: Moving options below project declaration

M2-TE:
This PR moves project options and cleans up CMake stuff around it. I can split up the changes into different PRs if needed or if only parts of this (if any) should be merged.

1.  Moved the options for CMake variables below the project declaration, as they do not impact the project creation itself.
2.  To accomodate this, `cmake_minimum_required` was set to 3.12 and only gets optionally checked for 3.28, in case `VULKAN_HPP_ENABLE_CPP20_MODULES` is enabled. Added another check for 3.30 if `VULKAN_HPP_ENABLE_STD_MODULE` is enabled as well.
3. Using CMake built-in variable `PROJECT_IS_TOP_LEVEL` as default value for option `VULKAN_HPP_GENERATOR_BUILD`. As this is only available from 3.21 onwards, manually checking the path to set the variable, just as in VulkanHpp.
4. As the `CMAKE_CXX_STANDARD` should only really be set by top-level projects, `CMAKE_CXX_STANDARD` is only set if `PROJECT_IS_TOP_LEVEL` is true.


M2-TE:
There seems to be an issue under windows-2022 with clang-format

```
CUSTOMBUILD : clang-format error : unable to overwrite file D:/a/Vulkan-Hpp/Vulkan-Hpp/vulkan/vulkan_enums.hpp: permission denied [D:\a\Vulkan-Hpp\Vulkan-Hpp\build\17\Debug\build_vulkan_hpp.vcxproj]
CUSTOMBUILD : clang-format error : unable to overwrite file D:/a/Vulkan-Hpp/Vulkan-Hpp/vulkan/vulkan.cppm: permission denied [D:\a\Vulkan-Hpp\Vulkan-Hpp\build\17\Debug\build_vulkan_hpp.vcxproj]
````

but I don't think I should have impacted anything regarding permissions.. @asuessenbach any idea as to why it fails ?

asuessenbach:
Every now and then, it happens that a runner has some hickup.
Such error can usually be "resolved" by re-running that failing jobs step. As I did here.

asuessenbach:
Improves the `CMakeLists.txt` substantially!

PR #2048, opened 2025-01-20
CMake: Alias for c++20 modules target

M2-TE:
Created an alias target similar to Vulkan::Hpp, while also renaming the Vulkan-Hpp target to VulkanHpp to be similar in name to VulkanHppModule.

Moving VulkanHpp target above module target allows linking to it for its include directory.


asuessenbach:
It seems, the windows runners have some problems right now.
With #2050 I've changed the windows workflow to do one job per C++-standard, instead of looping over them. That seems to be more stable.
So, if you could rebase your PR and try again?

M2-TE:
Looks like that did the trick

asuessenbach:
Looks good again. Thanks for your contribution!

PR #2054, opened 2025-01-22
CMake: Adding generator dependency to `VulkanHpp` interface target for propagation and linking it to other targets

M2-TE:
The `VulkanHpp` was sort of disconnected from everything, so I reused it as a target for both `VulkanHppModule` and the `vulkan_hpp__setup_vulkan_include` function to nicely unify the include directories. As the `VulkanHpp` headers may be requested to be generated first, a dependency was added to ensure that it and other targets wait for the generators to have run if `VULKAN_HPP_RUN_GENERATOR` is enabled.

To set up the dependency, the generator targets were moved to the very top to run before all else, only preceded by the function to set warning levels. Tinyxml and clang-format related things were also moved into the `VULKAN_HPP_GENERATOR_BUILD` conditional branch as well, running them only when necessary. Also moved the `VulkanHpp` and `VulkanHppModule` targets to the top, right below the generators to ensure they are available for `vulkan_hpp__setup_vulkan_include`.

`Vulkan-Headers/include` was added as an include directory to `VulkanHpp`, eliminating the need to additionally include it for a target afterwards. As far as I know, the hpp headers need the base headers anyways. One thing that may be desirable here is to use something like
```cmake
add_subdirectory(Vulkan-Headers)
target_link_libraries(VulkanHpp INTERFACE Vulkan::Headers)
```
to link to the explicit target that Vulkan-Headers provides, but I do not know if/how this will interact with `find_package(Vulkan)` in the tests, where statically linking is necessary. Also, users may already have the Vulkan-Headers fetched somewhere and may want to use those instead of having to download them as again as a git submodule, so it might make sense to first look if the `Vulkan::Headers` target exists before `add_subdirectory()`, or adding another `option()` or something. I'm leaving that on the table for now and want to first see if all the tests will succeed.


CMakeLists.txt
```diff
@@ -49,12 +49,174 @@ option( VULKAN_HPP_ENABLE_CPP20_MODULES "Build Vulkan-Hpp as C++20 module; requi
...
+	add_custom_target( build_video_hpp ALL DEPENDS "${vulkan_video_hpp}" "${video_spec}" )
+endif()
+
+# Create Vulkan-Hpp interface target
+add_library( VulkanHpp INTERFACE )
+add_library( Vulkan::Hpp ALIAS VulkanHpp )
+target_include_directories( VulkanHpp INTERFACE "${CMAKE_CURRENT_SOURCE_DIR}" )
+target_include_directories( VulkanHpp INTERFACE "${CMAKE_CURRENT_SOURCE_DIR}/Vulkan-Headers/include" )
+if( VULKAN_HPP_RUN_GENERATOR )
+	add_dependencies( VulkanHpp build_vulkan_hpp )
```
asuessenbach, line 188: I think, `VulkanHpp` should depend on `build_video_hpp` as well.

asuessenbach (commented):
Some nice changes again. Thanks a lot.
Just one question.

CMakeLists.txt
M2-TE: I thought so as well, but I copied this dependency from the `setup__include` function, which looked like this
https://github.com/KhronosGroup/Vulkan-Hpp/blob/16bfa65a5929290085d3c76caf705c4b9cbca949/CMakeLists.txt#L86-L94

should I change the dep to include `build_video_hpp` regardless?

asuessenbach: Yes, please.

PR #2066, opened 2025-02-07
CMake: Added unified options for `VulkanHpp` and `VulkanHppModule` compile definitions

M2-TE:
I created a CMake option for every flag-type `VULKAN_HPP_*` option. It seems a bit cleaner to define these compile options in CMake (on demand), rather than providing them in the code itself. It will also propagate such options to other libraries (most notably `VulkanMemoryAllocator`), which they could use to check if e.g. dynamic dispatch is enabled among other things.

The real reason behind this however was the ability to give `VulkanHppModule` access to all of these options without having to manually fish out the target name from the `Vulkan-Hpp` `CMakeLists.txt`. Since modules are much more stand-alone, propagation of options via CMake is much better than manually setting these compile definitions for each Vulkan-dependant library separately.

There are a few things left to note before thinking about merging anything:
- `VULKAN_HPP_CPP20_MODULE_DYNAMIC_DISPATCHER` option should probably be deprecated, but I've left it in for now as to not break anything.
- Options that are not simple flags like `VULKAN_HPP_NAMESPACE` are currently left out as they don't work with `option`. They would need to be `set`, but would be rather awkward for e.g. providing a custom assert function
- The options do not have a description yet
- There is also the matter of linking to the `vulkan-1` lib, which should be done in case `VULKAN_HPP_DISPATCH_LOADER_DYNAMIC` is disabled. Since this was already not the case before, I have left it as is.
- It might be worth thinking about using these options in the generators to cut down the resulting headers, which could have a pretty substantial impact on their size. It's way out of scope for this PR, but I just wanted to mention it.

I am leaving this as a draft for the aforementioned reasons and because I want to know if something like this is even desired. So please let me know what you think @asuessenbach


CMakeLists.txt
```diff
@@ -47,7 +47,38 @@ option( VULKAN_HPP_TESTS_BUILD_ONLY_DYNAMIC "Build only dynamic" OFF )
...
+# VULKAN_HPP_DEFAULT_DISPATCHER
+# VULKAN_HPP_DEFAULT_DISPATCHER_TYPE
+# VULKAN_HPP_EXPECTED
+# VULKAN_HPP_UNEXPECTED
+# VULKAN_HPP_HASH_COMBINE
+# VULKAN_HPP_INLINE
+# VULKAN_HPP_NAMESPACE
+# VULKAN_HPP_STORAGE_API
+# VULKAN_HPP_ENABLE_DYNAMIC_LOADER_TOOL
+# TODO: allow generation of shorter vulkan header files by reading compile definitions?
```
asuessenbach, line 81: Might be worth a try again.
But earlier investigations resulted in very marginal improvements by skipping generation of parts of the headers.

asuessenbach (commented):
I like the approach of having all those flags listed in the `CMakeLists.txt`, not only in the `Readme.md`.
You're right, they would need some appropriate description.
And in order to help humans to find something in all those flags, it might be worth to sort them alphabetically.
Regarding those non-boolean defines: I don't think, it's worth to move them into CMakeLists.txt as well. They are nicely handled in some compact form in vulkan_hpp_macros.hpp.

CMakeLists.txt
```diff
@@ -197,6 +228,59 @@ if( VULKAN_HPP_RUN_GENERATOR )
 	add_dependencies( VulkanHpp build_vulkan_hpp build_video_hpp )
 endif()
 
+# set up compile definitions
+if ( VULKAN_HPP_DISABLE_ENHANCED_MODE )
+	target_compile_definitions(VulkanHpp INTERFACE "VULKAN_HPP_DISABLE_ENHANCED_MODE")
+endif()
+if ( VULKAN_HPP_DISPATCH_LOADER_DYNAMIC )
+	target_compile_definitions(VulkanHpp INTERFACE "VK_NO_PROTOTYPES")
+	target_compile_definitions(VulkanHpp INTERFACE "VULKAN_HPP_DISPATCH_LOADER_DYNAMIC")
```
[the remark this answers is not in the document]
M2-TE: Oh right, they should be separated. To ensure consistency, I think it makes sense to create `VK_NO_PROTOTYPES` (after `option()` section) if it was not yet defined and (only then) set it to the value of `VULKAN_HPP_DISPATCH_LOADER_DYNAMIC` by default. In case both were not defined, they are both `OFF` as expected.

In which case should we link to `vulkan-1`? When either is `ON`, only one of them is, or both? I would like to put a placeholder at the correct location for now

CMakeLists.txt
M2-TE: Fair enough. Do you happen to still have any numbers on hand regarding the size differences?

asuessenbach: > Do you happen to still have any numbers on hand regarding the size differences?

Unfortunately, no.
And it was not about file size differences, but about compile time differences. Which where in the range of less than 5%, at best, with some substantially more complex generation patterns.

CMakeLists.txt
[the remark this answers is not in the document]
asuessenbach: I think, the correct move would be to link to `vulkan-1`, when `VULKAN_HPP_DISPATCH_LOADER_DYNAMIC` is off.

asuessenbach:
Looks great now!

PR #2071, opened 2025-02-11
CMake: Update minimum version to 3.22.1

M2-TE:
Updated to match Vulkan-Headers, Vulkan Validation Layers and SPIR-V Tools.
See KhronosGroup/Vulkan-Headers#525

Removed ci for `ubuntu-20.04` as it only ships `CMake 3.16.3`


asuessenbach:
I don't see why we should push the CMake minimum version to 3.22.1. We can live with what we have, can't we?

M2-TE:
The plan was to add `Vulkan-Headers` via `add_subdirectory` or `FetchContent` instead of just adding the include directory, as we could make use of their version check and also prevent the currently duplicated header files (vulkan.hpp is present in both Vulkan-Headers and Vulkan-Hpp), while also providing the `Vulkan::Headers` target, if people need it.

These things are not strictly necessary, but I figured it would make sense to update to the same version as the `Vulkan-Headers` dependency, which would also enable some extra aforementioned things. A version range via `3.12...3.22.1` might also do the trick to not place any hard requirements.

asuessenbach:
In fact, I don't think, it's needed.
And what's more important here: I don't like to cut down the CI by removing the checks on Ubuntu 20.04, just to keep CMake happy.
That is: as long as there's no substantial reason to require a newer version of CMake I'd like to stay with it.

PR #2073, opened 2025-02-15
vulkan_handles: Remove the redundant semicolon after ctor definition

gnoliyil:
Semicolons after in-class function definitions are redundant, and may trigger the "-Wextra-semi" warning of compilers, for example:

* gcc: https://gcc.gnu.org/onlinedocs/gcc/C_002b_002b-Dialect-Options.html#index-Wextra-semi
* clang: https://clang.llvm.org/docs/DiagnosticsReference.html#wextra-semi

This change removes the redundant semicolon in the code generator and the generated hpp file.

Bug: https://fxbug.dev/378964821


gnoliyil:
@asuessenbach could you please take a look?

asuessenbach:
Thanks a lot for spotting this issue... and fixing it right away!

PR #2077, opened 2025-02-17
C++20 module: Hide to_string behind VULKAN_HPP_NO_TO_STRING

### LINKED ISSUE #2076: Only optionally include `vulkan_to_string.hpp` in `vulkan.cppm`
The current `vulkan.cppm` C++20 module always includes the `vulkan_to_string.hpp` header in the preamble

https://github.com/KhronosGroup/Vulkan-Hpp/blob/9dbf676cdbe6059815deba828005cee73864b9b8/vulkan/vulkan.cppm#L18-L27

while optionally exporting the corresponding symbols based on `VULKAN_HPP_NO_TO_STRING`

https://github.com/KhronosGroup/Vulkan-Hpp/blob/9dbf676cdbe6059815deba828005cee73864b9b8/vulkan/vulkan.cppm#L962-L965

It might make sense to also prevent inclusion of the header within the global module fragment preamble, where preprocessor commands are still valid, as this will prevent the preprocessor from having to go over the (rather large) header file:
```cpp
#ifndef VULKAN_HPP_NO_TO_STRING
#  include <vulkan/vulkan_to_string.hpp>
#endif /*VULKAN_HPP_NO_TO_STRING*/ 

export module vulkan_hpp;
```

~~I would open a PR about this, but I'm not quite sure where exactly in the generator code this part of the .cppm is built.~~

M2-TE:
Resolves #2076

### DIFF
--- a/snippets/CppModuleFileTemplate.hpp
+++ b/snippets/CppModuleFileTemplate.hpp
@@ -18,7 +18,7 @@ module;
 #include <vulkan/${api}_raii.hpp>
 #include <vulkan/${api}_shared.hpp>
 #ifndef VULKAN_HPP_NO_TO_STRING
-#  include <vulkan/${api}_to_string.hpp>
+#include <vulkan/${api}_to_string.hpp>
 #endif
 
 export module ${api}_hpp;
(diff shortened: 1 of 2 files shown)


asuessenbach:
Nice catch!
Thanks again.

PR #2083, opened 2025-02-19
CMake: Reintroduced #2075

M2-TE:
FindVulkan is now optional, as generators do not need it


asuessenbach:
As already said in #2075: VulkanHppGenerator doesn't really need that include directory

That is, wouldn't it be a better fit to add this functionality to the `vulkan_hpp__setup_vulkan_include` function (currently lines 337-348 in CMakeLists.txt), which is used for every library, sample and test?

M2-TE:
Including the Vulkan C headers in `vulkan_hpp__setup_vulkan_include` would work for all local things, but VulkanHpp and the module would no longer work as standalone targets outside the scope of this project (unless they call that function manually, which is usually not inteded for external projects in CMake) and requires users to include/link the C headers manually

Though, `Vulkan-Headers` is probably the better repo to use as a primary way to include _all_ vulkan headers (albeit losing the ability to customize the generator or formatting and with currently worse CMake support), so I agree with putting the C includes into the function

PR #2087, opened 2025-02-21
Add additional type conversion operators to the struct classes to make it easier to interoperate with the Vulkan C API

pdaniell-nv:
These new operators allow the Vulkan-Hpp struct classes to be more easily used directly with the Vulkan C API.

For example, instead of this:
```c++
vk::BufferCreateInfo bufferInfo { .size{sizeof vertexData}, .usage{vk::BufferUsageFlagBits::eVertexBuffer} };
vmaCreateBuffer(allocator, &static_cast<VkBufferCreateInfo const &>(bufferInfo), &bufferAllocationInfo, &vertexBuffer, &vertexBufferAllocation, nullptr);
```

Apps can do this:
```c++
vk::BufferCreateInfo bufferInfo { .size{sizeof vertexData}, .usage{vk::BufferUsageFlagBits::eVertexBuffer} };
vmaCreateBuffer(allocator, bufferInfo, &bufferAllocationInfo, &vertexBuffer, &vertexBufferAllocation, nullptr);
```

### DIFF
--- a/VulkanHppGenerator.cpp
+++ b/VulkanHppGenerator.cpp
@@ -11983,9 +11983,14 @@ ${constructorsAndSetters}
       return *reinterpret_cast<Vk${structureType}*>( this );
     }
 
-    operator Vk${structureType} const 8() VULKAN_HPP_NOEXCEPT
+    operator Vk${structureType} const *() const VULKAN_HPP_NOEXCEPT
     {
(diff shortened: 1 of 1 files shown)


asuessenbach:
Yes, having both, pointer cast operators as well as reference cast operators (const and non-const) seems to be reasonable.

PR #2096, opened 2025-03-06
Optimize `to_string` functions for enums

LocalSpook:
Currently, these functions are implemented rather inefficiently. Here's a random one to demonstrate:

```cpp
VULKAN_HPP_INLINE std::string to_string( CommandBufferUsageFlags value )
{
  if ( !value )
    return "{}";

  std::string result;
  if ( value & CommandBufferUsageFlagBits::eOneTimeSubmit )
    result += "OneTimeSubmit | ";
  if ( value & CommandBufferUsageFlagBits::eRenderPassContinue )
    result += "RenderPassContinue | ";
  if ( value & CommandBufferUsageFlagBits::eSimultaneousUse )
    result += "SimultaneousUse | ";

  return "{ " + result.substr( 0, result.size() - 3 ) + " }";
}
```

The `substr()` call incurs a copy of the string.
Then, because we _prepend_ `{`, all the string's contents need to be memmoved up in memory.
And overall we're not taking advantage of NRVO.

With a few tweaks, we can solve all those problems ([compare the assembly](https://godbolt.org/z/ohd78dov5)):

```cpp
VULKAN_HPP_INLINE std::string to_string( CommandBufferUsageFlags value )
{
  std::string result = "{ "; // ⭐ Instead of prepending {, initialize with it.
  if ( value & CommandBufferUsageFlagBits::eOneTimeSubmit )
    result += "OneTimeSubmit | ";
  if ( value & CommandBufferUsageFlagBits::eRenderPassContinue )
    result += "RenderPassContinue | ";
  if ( value & CommandBufferUsageFlagBits::eSimultaneousUse )
    result += "SimultaneousUse | ";

  if ( result.size() > 2 ) {
    result.resize(result.size() - 1);  // ⭐ No copy
    result.back() = '}';
  } else { 
    result = "{}";
  }
  return result; // ⭐ NRVO
}
```

I then found that avoiding the `resize()` call [generates even smaller code](https://godbolt.org/z/8GG5Wq7r9), so the final version looks like:

```cpp
VULKAN_HPP_INLINE std::string to_string( CommandBufferUsageFlags value )
{
  std::string result = "{";
  if ( value & CommandBufferUsageFlagBits::eOneTimeSubmit )
    result += " OneTimeSubmit |";
  if ( value & CommandBufferUsageFlagBits::eRenderPassContinue )
    result += " RenderPassContinue |";
  if ( value & CommandBufferUsageFlagBits::eSimultaneousUse )
    result += " SimultaneousUse |";

  if ( result.size() > 1 )
    result.back() = '}';
  else 
    result = "{}";
  return result;
}
```

### DIFF
--- a/VulkanHppGenerator.cpp
+++ b/VulkanHppGenerator.cpp
@@ -2570,16 +2570,45 @@ std::string VulkanHppGenerator::generateBitmaskToString( std::map<std::string, B
   std::string enumName    = stripPrefix( bitmaskBitsIt->first, "Vk" );
 
   std::string str;
-  if ( bitmaskBitsIt->second.values.empty() ||
-       std::none_of( bitmaskBitsIt->second.values.begin(), bitmaskBitsIt->second.values.end(), []( auto const & evd ) { return evd.supported; } ) )
+  std::string emptyValue = "{}";
+  std::string toStringChecks;
+  std::string previousEnter, previousLeave;
+  for ( auto const & value : bitmaskBitsIt->second.values )
   {
-    static std::string bitmaskToStringTemplate = R"(
+    if ( value.supported )
+    {
+      std::string valueName = generateEnumValueName( bitmaskBitsIt->first, value.name, true );
+      if ( value.value == "0" )
+      {
+        assert( emptyValue == "{}" );
+        emptyValue = valueName.substr( 1 );
+      }
+      else if ( !value.bitpos.empty() )
+      {
+        const auto [enter, leave] = generateProtection( value.protect );
+        toStringChecks += ( ( previousEnter != enter ) ? ( previousLeave + enter ) : "" ) + "    if ( value & " + enumName + "::" + valueName +
+                          " ) result += \" " + valueName.substr( 1 ) + " |\";\n";
+        previousEnter = enter;
+        previousLeave = leave;
+      }
+    }
+  }
+  if ( !previousLeave.empty() )
+  {
+    assert( previousLeave.ends_with( "\n" ) );
+    toStringChecks += previousLeave;
+    previousLeave.resize( previousLeave.size() - strlen( "\n" ) );
+  }
+
+  if ( toStringChecks.empty() )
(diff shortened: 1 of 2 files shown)


asuessenbach:
That's a nice little improvement!

asuessenbach:
And your PR revealed an error!
To fix that, would you please change the last call to `replaceWithBitmap` in `generateBitmaskToString` with something like this:
```
    if ( toStringChecks.empty() )
    {
      static const std::string bitmaskToStringTemplate = R"(
  VULKAN_HPP_INLINE std::string to_string( ${bitmaskName} value )
  {
    VULKAN_HPP_ASSERT(!value);
    return "${emptyValue}";
  }
)";

      str += replaceWithMap( bitmaskToStringTemplate, { { "bitmaskName", bitmaskName }, { "emptyValue", emptyValue } } );
    }
    else
    {
      static const std::string bitmaskToStringTemplate = R"(
  VULKAN_HPP_INLINE std::string to_string( ${bitmaskName} value )
  {
    std::string result = "{";
${toStringChecks}
    if ( result.size() > 1 )
      result.back() = '}';
    else 
      result = "${emptyValue}";
    return result;
  }
)";

      str += replaceWithMap( bitmaskToStringTemplate, { { "bitmaskName", bitmaskName }, { "emptyValue", emptyValue }, { "toStringChecks", toStringChecks } } );
    }
```

LocalSpook:
I believe this snippet would still trigger `-Wunused-parameter` when compiling with `-DNDEBUG`, because the assert disappears; is that okay?

```cpp
  VULKAN_HPP_INLINE std::string to_string( ${bitmaskName} value )
  {
    VULKAN_HPP_ASSERT(!value);
    return "${emptyValue}";
  }
```

asuessenbach:
> I believe this snippet would still trigger -Wunused-parameter when compiling with -DNDEBUG

Correct. And your change resolves it nicely.

PR #2100, opened 2025-03-11
Remove the parallel file generation due to too many file access issues on CI builds.

asuessenbach:


asuessenbach:
> I have never benchmarked it
I did once... and formatting took about 5 times as long as generating :(

I think, for now, I'll just take this one and done.

PR #2105, opened 2025-03-17
CMake: Cleanup of vulkan target and cpp module test

M2-TE:
1. When using dynamic dispatch, unix should link to `CMAKE_DL_LIBS` (previously only done on the tests)
2. `CXX_STANDARD_REQUIRED` is implied by cxx_std_xx so it can be omitted
3. `CXX_EXTENSIONS` only need to be disabled for clang 16 and should otherwise remain untouched
4. Tests for cpp modules should set `CXX_SCAN_FOR_MODULES` to explicitly scan for modules, else `vulkan_hpp` may not be found. They also internally handle dynamic dispatching things, so the CMake side was simplified.
5. ~~Added primitive testing for cpp modules to CI, currently only ubuntu to see if it even works. It is not pretty, but I manually exclude gcc <= 14 and clang <= 17, since this will not add work in the future when new compiler versions are added. Modules also require CMake 3.28, so I limited it to Ubuntu 24.04~~ The compilers are a little too old still for stable compilation of modules (gcc 15 and clang 19 are pretty good), so I will retract this portion. Maybe we should do it in Windows instead for now?

### DIFF
--- a/.github/workflows/ci-ubuntu.yml
+++ b/.github/workflows/ci-ubuntu.yml
@@ -70,46 +70,21 @@ jobs:
         do
           for build_type in Debug Release
           do
-            # filter out compilers that should not compile cpp modules (clang <= 17, gcc <= 14), as well as cmake versions < 3.28
-            if [[ ${{matrix.os}} == ubuntu-24.04                                                                                                                      \
-              && ${{matrix.compiler}} != clang++-10 && ${{matrix.compiler}} != clang++-11 && ${{matrix.compiler}} != clang++-12 && ${{matrix.compiler}} != clang++-13 \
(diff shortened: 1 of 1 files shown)


asuessenbach:
> Maybe we should do it in Windows instead for now?

Would be a good to have at least one platform testing/compiling with modules.
Would you like to create such a PR?

Besides that, thanks a lot for your contribution!

M2-TE:
> Would be a good to have at least one platform testing/compiling with modules.

@asuessenbach Yea, that's my goal as well. GCC 14+ and Clang 16+ should both theoretically be able to compile modules, but they each have some unique oddities that makes it hard to get consistent results..

> Would you like to create such a PR?

Yea I will try my hand at Windows, since MSVC seems to have some pretty good module support!

PR #2110, opened 2025-03-19
ci: C++20 module testing

M2-TE:
- Added Ubuntu 24.04 with clang++-19
- Added modules testing with clang++-16 to 19
- Modules should theoretically work with g++-14, but there seems to be a bug when the testing source file includes the standard library in any way

### DIFF
--- a/.github/workflows/ci-ubuntu.yml
+++ b/.github/workflows/ci-ubuntu.yml
@@ -70,21 +70,45 @@ jobs:
         do
           for build_type in Debug Release
           do
-            echo "================================================================================="
-            echo "Building C++"$cpp_standard" in "$build_type" mode on "${{matrix.os}}" with "${{matrix.compiler}}
-            echo "================================================================================="
-            cmake -B build/$cpp_standard/$build_type -GNinja \
-              -DVULKAN_HPP_SAMPLES_BUILD=ON                  \
-              -DVULKAN_HPP_SAMPLES_BUILD_ONLY_DYNAMIC=ON     \
-              -DVULKAN_HPP_TESTS_BUILD=ON                    \
-              -DVULKAN_HPP_TESTS_BUILD_ONLY_DYNAMIC=ON       \
-              -DVULKAN_HPP_BUILD_WITH_LOCAL_VULKAN_HPP=ON    \
-              -DVULKAN_HPP_PRECOMPILE=OFF                    \
-              -DVULKAN_HPP_RUN_GENERATOR=ON                  \
-              -DCMAKE_CXX_COMPILER=${{matrix.compiler}}      \
-              -DCMAKE_CXX_STANDARD=$cpp_standard             \
-              -DCMAKE_BUILD_TYPE=$build_type
-            cmake --build build/$cpp_standard/$build_type --parallel
+            # selectively filter out compilers that do not have the requisite C++20 modules capabilities
+            if [[ ${{matrix.compiler}} != clang++-10 && ${{matrix.compiler}} != clang++-11 && ${{matrix.compiler}} != clang++-12 && ${{matrix.compiler}} != clang++-13
+              && ${{matrix.compiler}} != clang++-14 && ${{matrix.compiler}} != clang++-15
+              && ${{matrix.compiler}} != g++-11 && ${{matrix.compiler}} != g++-12 && ${{matrix.compiler}} != g++-13 && ${{matrix.compiler}} != g++-14 ]]
+            then
+              echo "================================================================================="
+              echo "Building C++"$cpp_standard" with modules in "$build_type" mode on "${{matrix.os}}" with "${{matrix.compiler}}
+              echo "================================================================================="
+              cmake -B build/$cpp_standard/$build_type -GNinja \
+                -DVULKAN_HPP_SAMPLES_BUILD=ON                  \
+                -DVULKAN_HPP_SAMPLES_BUILD_ONLY_DYNAMIC=ON     \
+                -DVULKAN_HPP_TESTS_BUILD=ON                    \
+                -DVULKAN_HPP_TESTS_BUILD_ONLY_DYNAMIC=ON       \
+                -DVULKAN_HPP_ENABLE_CPP20_MODULES=ON           \
+                -DVULKAN_HPP_DISPATCH_LOADER_DYNAMIC=ON        \
+                -DVULKAN_HPP_BUILD_WITH_LOCAL_VULKAN_HPP=ON    \
+                -DVULKAN_HPP_PRECOMPILE=OFF                    \
+                -DVULKAN_HPP_RUN_GENERATOR=ON                  \
+                -DCMAKE_CXX_COMPILER=${{matrix.compiler}}      \
+                -DCMAKE_CXX_STANDARD=$cpp_standard             \
+                -DCMAKE_BUILD_TYPE=$build_type
+              cmake --build build/$cpp_standard/$build_type --parallel
+            else
+              echo "================================================================================="
+              echo "Building C++"$cpp_standard" in "$build_type" mode on "${{matrix.os}}" with "${{matrix.compiler}}
+              echo "================================================================================="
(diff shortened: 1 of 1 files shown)


M2-TE:
@asuessenbach Where exactly do the compilers get installed for CI? Ubuntu 24.04 should have access to clang++-19 via apt

asuessenbach:
We don't explicitly install any compiler, we just use what the runners are providing, as listed in [Ubuntu2404-Readme](https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md). That is, there are only `Clang: 16.0.6, 17.0.6, 18.1.3` on Ubuntu24.04.

.github/workflows/ci-ubuntu.yml
```diff
@@ -70,21 +70,45 @@ jobs:
...
+                -DVULKAN_HPP_ENABLE_CPP20_MODULES=ON           \
+                -DVULKAN_HPP_DISPATCH_LOADER_DYNAMIC=ON        \
+                -DVULKAN_HPP_BUILD_WITH_LOCAL_VULKAN_HPP=ON    \
+                -DVULKAN_HPP_PRECOMPILE=OFF                    \
+                -DVULKAN_HPP_RUN_GENERATOR=ON                  \
+                -DCMAKE_CXX_COMPILER=${{matrix.compiler}}      \
+                -DCMAKE_CXX_STANDARD=$cpp_standard             \
+                -DCMAKE_BUILD_TYPE=$build_type
+              cmake --build build/$cpp_standard/$build_type --parallel
+            else
```
asuessenbach, line 96: Instead of having two branches here with nearly identical settings, you maybe could introduce some variable `enable_cpp20_modules = ...`, and use it like `-DVULKAN_HPP_ENABLE_CPP20_MODULES=$enable_cpp20_modules`?

Or maybe even add that to the strategy matrix above and use it like `-DVULKAN_HPP_ENABLE_CPP20_MODULES=${{enable_cpp20_modules}}`?

(Didn't try if that works at all)

.github/workflows/ci-ubuntu.yml
```diff
@@ -70,21 +70,46 @@ jobs:
...
+                -DVULKAN_HPP_TESTS_BUILD=ON                    \
+                -DVULKAN_HPP_TESTS_BUILD_ONLY_DYNAMIC=ON       \
+                -DVULKAN_HPP_BUILD_WITH_LOCAL_VULKAN_HPP=ON    \
+                -DVULKAN_HPP_PRECOMPILE=OFF                    \
+                -DVULKAN_HPP_RUN_GENERATOR=ON                  \
+                -DCMAKE_CXX_COMPILER=${{matrix.compiler}}      \
+                -DCMAKE_CXX_STANDARD=$cpp_standard             \
+                -DCMAKE_BUILD_TYPE=$build_type
+              cmake --build build/$cpp_standard/$build_type --parallel
+            fi
```
asuessenbach, line 112: Somehow the part using cpp_standard=23 is missing?

.github/workflows/ci-ubuntu.yml
M2-TE: I also quite dislike the branches, so something like this would be much preferred. I will give the matrix solution a shot first

M2-TE: Looks to work just fine, though I can test the variable solution as well if you don't like the additional ON/OFF info that is now in the runner names, e.g.:
`CI Ubuntu / build (ubuntu-24.04, clang++-18, ON)`

M2-TE:
- module testing is now being done with `windows-2022`, `clang++-17` and `clang++-18`
- `clang++-16` theoretically compiles modules, but does not seem to produce BMIs with CMake, so we skip it
- macos won't be able to test these, as `clang++-15` is the latest version on the macos-14 and macos-15 runners

There is one peculiar thing with the vulkan module though, you may have noticed that I added `#include <memory>` to the `Cpp20Modules` test, even though it does not use it directly. On Windows, this is needed for it to compile, else it, for some reason, has duplicate definitions for `std::unique_ptr`.
On Ubuntu with g++-14, there is a similar problem with all contents of the memory header being redefined when _any_ header containing any of those definitions is included in `Cpp20Modules` (which is why I have to exclude it from testing).

These point towards there being some issue with these symbols that are from the memory header, but I can't figure out where it stems from. It is not relevant for this specific PR, but do you have any idea that might help pinpoint this issue?

M2-TE:
Seems like runners may run out of storage space, since, on Ubuntu at least, all builds are done at separate locations, i.e.:
`/home/runner/work/Vulkan-Hpp/Vulkan-Hpp/build/<cpp_standard>/<Debug/Release>`
Do we need to retain the build folders afterwards? Otherwise it may be best to delete them or instead split the work across multiple runners.

asuessenbach:
> Seems like runners may run out of storage space since

That's a good observation. Will split into multiple runners.

asuessenbach:
With #2116, the ubuntu CI has been split into three separate .yml files. That is, your PR needs some rework. I'm sorry for that!

M2-TE:
Thanks to your PR, the module testing on ubuntu was actually simplified a lot, so thank you!
It can only really be tested on Ubuntu 24 anyways, since it needs CMake 28+ and because of that, there are much fewer compilers to exclude. And no duplicate work for cpp-23 either!

asuessenbach:
Your PR includes Vulkan-Headers. I assume, that's not intended?

M2-TE:
Good catch, looks like I did a sloppy merge from main without updating the submodule.. I apologize about that!

M2-TE:
I've tried to get modules working on macos since macos-15 has access to clang-18, but it does not seem to work properly and I do not have an apple device to further troubleshoot this, so I'm postponing module testing on that one.

Since I was already at it, I've done the same changes to macos so that it is split up into multiple runners like on windows and ubuntu, while adding macos-15 in addition to macos-14.

.github/workflows/ci-macos.yml
```diff
@@ -11,8 +11,13 @@ jobs:
...
     strategy:
       matrix:
-        os: [macos-14]
+        os: [macos-14, macos-15]
         compiler: [clang++]
+        cpp_standard: [11, 14, 17, 20, 23]
+        build_type: [Debug, Release]
+        exclude:
+          - os: macos-14
+            compiler: clang++-18
```
asuessenbach, line 20: The macos-14 runner does not have clang++-18, so you don't need to explicitly exclude it here.

.github/workflows/ci-macos.yml
```diff
@@ -30,21 +35,15 @@ jobs:
       
     - name: Loop over cpp_standards (11, 14, ...) and build_types (Debug, Release)
```
asuessenbach, line 36: The name could be change to just "Build".

.github/workflows/ci-ubuntu-24.yml
```diff
@@ -37,11 +37,14 @@ jobs:
 
     - name: Build
       run: |
+        modules_flag=${{(contains(fromJSON('["clang++-16", "g++-12", "g++-13", "g++-14"]'), matrix.compiler) || matrix.cpp_standard < 20) && 'OFF' || 'ON'}}
```
asuessenbach, line 40: Would be nice, if the ubuntu and the windows CI had the same ON/OFF logic here.

M2-TE: I opted to change the windows portion to make these have the same logic. It's a bit messy to chain all those (compiler != clang++-XX && ...) together

.github/workflows/ci-macos.yml
M2-TE: Oh right, leftover from trying to compile modules on here

.github/workflows/ci-ubuntu-24.yml
asuessenbach: That's perfectly fine. Thanks.

asuessenbach:
Hopefully, we will catch module related issues earlier now.

PR #2123, opened 2025-04-01
Replace ubuntu24.04 workflow by a self-hosted linux runner

asuessenbach:


M2-TE:
> /bin/sh: 1: CMAKE_CXX_COMPILER_CLANG_SCAN_DEPS-NOTFOUND: not found

You will need `clang-tools` (on Ubuntu) to get the clang module dependency scanner

PR #2146, opened 2025-04-25
C++20 Module: Fixing dynamic dispatch on Windows MSVC

M2-TE:
Under Windows, we need to export the `defaultDispatchLoaderDynamic` symbol, as simply importing vulkan_hpp will not provide it without including the macro header in every .cppm file that imports vulkan_hpp. I'm actually suprised it worked under Clang. It wasn't caught by the tests, as they manually include `vulkan_hpp_macros.hpp` for dynamic storage, so the symbol will always be present in that single .cppm source.

This also enables the storage placement with modules to be handled without the macro header inclusion:
```cpp
import vulkan_hpp;
namespace vk::detail {
    DispatchLoaderDynamic defaultDispatchLoaderDynamic;
}
```

### DIFF
--- a/VulkanHppGenerator.cpp
+++ b/VulkanHppGenerator.cpp
@@ -5615,6 +5615,9 @@ std::string VulkanHppGenerator::generateCppModuleUsings() const
   {
     using VULKAN_HPP_NAMESPACE::detail::DispatchLoaderBase;
     using VULKAN_HPP_NAMESPACE::detail::DispatchLoaderDynamic;
+#if defined( VULKAN_HPP_DISPATCH_LOADER_DYNAMIC )
+    using VULKAN_HPP_NAMESPACE::detail::defaultDispatchLoaderDynamic;
+#endif
 #if !defined( VK_NO_PROTOTYPES )
     using VULKAN_HPP_NAMESPACE::detail::DispatchLoaderStatic;
     using VULKAN_HPP_NAMESPACE::detail::getDispatchLoaderStatic;
--- a/VulkanHppGenerator.cpp
+++ b/VulkanHppGenerator.cpp
@@ -5616,9 +5616,9 @@ std::string VulkanHppGenerator::generateCppModuleUsings() const
     using VULKAN_HPP_NAMESPACE::detail::DispatchLoaderBase;
     using VULKAN_HPP_NAMESPACE::detail::DispatchLoaderDynamic;
 #if !defined( VULKAN_HPP_DEFAULT_DISPATCHER )
-#   if VULKAN_HPP_DISPATCH_LOADER_DYNAMIC == 1
-      using VULKAN_HPP_NAMESPACE::detail::defaultDispatchLoaderDynamic;
-#   endif
+# if VULKAN_HPP_DISPATCH_LOADER_DYNAMIC == 1
+    using VULKAN_HPP_NAMESPACE::detail::defaultDispatchLoaderDynamic;
+# endif
 #endif
 #if !defined( VK_NO_PROTOTYPES )
     using VULKAN_HPP_NAMESPACE::detail::DispatchLoaderStatic;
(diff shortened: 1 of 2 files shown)


asuessenbach:
Thanks again for spotting this issue and fixing it right away!

dsvensson:
Having issues with this, hopefully just confusion.

1. Start of `vulkan.cppm` has `#include <vulkan/vulkan_hpp_macros.hpp>`
2. `vulkan/vulkan_hpp_macros.hpp` does if `!defined( VULKAN_HPP_DEFAULT_DISPATCHER )`
    1.  `extern VULKAN_HPP_STORAGE_API DispatchLoaderDynamic defaultDispatchLoaderDynamic;`
    2. `#define VULKAN_HPP_DEFAULT_DISPATCHER ::VULKAN_HPP_NAMESPACE::detail::defaultDispatchLoaderDynamic`
    3. `#define VULKAN_HPP_DEFAULT_DISPATCH_LOADER_DYNAMIC_STORAGE ...`
3. `vulkan.cppm` then goes on with `export namespace VULKAN_HPP_NAMESPACE` where it checks if `!defined( VULKAN_HPP_DEFAULT_DISPATCHER` and `VULKAN_HPP_DISPATCH_LOADER_DYNAMIC == 1` and if so exports the `vk::detail::defaultDispatchLoaderDynamic;`.

Doesn't this mean that if the `defaultDispatchLoaderDynamic` extern and the `VULKAN_HPP_DEFAULT_DISPATCH_LOADER_DYNAMIC_STORAGE` are declared, then `defaultDispatchLoaderDynamic` will not be exported as `VULKAN_HPP_DEFAULT_DISPATCHER` will be defined and the guard is `!defined`?

M2-TE:
I think you are right @dsvensson, I simply copied the macro guards.

For some reason, this fixed the issues I had on MSVC, though I'm not entirely sure why in retrospect. The correct solution should be to simply remove the `!defined(VULKAN_HPP_DEFAULT_DISPATCHER)`, I'll create a PR for it.

Thanks for pointing this out!

dsvensson:
Although, regardless of what I do, 

```c++
import vulkan_hpp;

namespace vk::detail {
DispatchLoaderDynamic defaultDispatchLoaderDynamic;
}
```

...or...

```c++
```


....I seem to get:

```c++
vk-dispatcher.ccm:37:2: error: reference to 'defaultDispatchLoaderDynamic' is ambiguous
   37 |         VULKAN_HPP_DEFAULT_DISPATCHER.init();
      |         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

While that can be ignored on MSVC - and runs and works etc 😬 , it fails on clang 20. Do I have to declare that in a regular .cpp file? Would be nice with a proper example. I guess non-module TU for the storage it is 😭.

M2-TE:
> Do I have to declare that in a regular .cpp file?

Yes, use your first solution (I would recommend this one):
```cpp
import vulkan_hpp;
namespace vk::detail {
    DispatchLoaderDynamic defaultDispatchLoaderDynamic;
}
```
or the second solution, but do not forget to include the macro header
```cpp
import vulkan_hpp;
```

Both of these compile on Clang 20 without warnings or errors for me

> I guess non-module TU for the storage it is

I've been thinking about how to place dynamic dispatch storage into the vulkan.cppm module itself (which would greatly simplify this), but it is very tricky with the current setup. Stick with the .cpp implementation for now!

dsvensson:
Tried putting it in a separate module,

```c++
export module gpu.vk.dispatcher.storage;

import vulkan_hpp;

namespace vk::detail {
DispatchLoaderDynamic defaultDispatchLoaderDynamic;
}
```

```
error: declaration 'defaultDispatchLoaderDynamic' attached to named module 'gpu.vk.dispatcher.storage' cannot be attached to other modules
	8 | DispatchLoaderDynamic defaultDispatchLoaderDynamic;
|                       ^
```

So I guess regular .cpp file.. and yeah, that works.

PR #2158, opened 2025-05-07
C++20 Module: Exporting PFN_* symbols

### LINKED ISSUE #2154: no vulkan_core.h in vulkan.cppm
I have experiments with vulkan modules, i try to assess PFN_vkGetInstanceProcAddr in which locate in vulkan_core.h as I found out, there is no vulkan_core.h in vulkan.cppm, any reason why, will it be add?

for now i use #include vulkan_core.h, i need PFN_vkGetInstanceProcAddr for context.

M2-TE:
Resolves #2154

Some people will want to have access to the PFN_* symbols when importing vulkan.cppm. I've updated the generator to add these symbols under the `PFN TYPES` group, directly after `BASE TYPES`. Let me know if these should be placed somewhere else (particularly since they are not part of the vk:: namespace).

The new function `appendCppModulePFNCommands` to construct the string data is basically a cut down version of `appendDispatchLoaderDynamicCommands`, since the output was very similar.

### DIFF
--- a/VulkanHppGenerator.cpp
+++ b/VulkanHppGenerator.cpp
@@ -719,7 +719,7 @@ void VulkanHppGenerator::appendDispatchLoaderDynamicCommands( std::vector<Requir
   }
 }
 
-void VulkanHppGenerator::appendCppModulePFNCommands( std::vector<RequireData> const & requireData,
+void VulkanHppGenerator::appendCppModuleCommands( std::vector<RequireData> const & requireData,
                                                      std::set<std::string> &          listedCommands,
                                                      std::string const &              title,
                                                      std::string &                    commandMembers ) const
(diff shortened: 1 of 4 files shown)


asuessenbach:
> I've updated the generator to add these symbols under the `PFN TYPES` group, directly after `BASE TYPES`.

Do these symbols really belong there? They'd be listed inside the `export namespace VULKAN_HPP_NAMESPACE` section, which sounds wrong. If at all, shouldn't they be listed at the very end, outside of any namespace, like `::VkGeometryInstanceFlagsKHR`?

And, after all, do they really belong into `vulkan.cppm`? Maybe we could introduce a new file, like `vulkanc.cppm`, or so, instead?

M2-TE:
>  If at all, shouldn't they be listed at the very end, outside of any namespace, like ::VkGeometryInstanceFlagsKHR?

absolutely, that would be a better placement for them.

It does feel odd to export these Vulkan C symbols in the Hpp module, but there's no Hpp counterpart for them, despite still being used in some Hpp-only projects. Even for those just using the default dynamic dispatcher, `PFN_vkGetInstanceProcAddr` and `PFN_vkGetDeviceProcAddr` are sometimes needed to pass them over to VMA for example.


If we were to introduce something like `vulkanc.cppm`, are there other things that we could export besides the PFN symbols? If it's only these PFN symbols, I'm not sure it would be worth it.

PR #2159, opened 2025-05-07
Use ref qualifier for RAII handles to prevent creating reference to a temporary.

stripe2933:
This changes separate the `const vk::<handle-type> &operator*() const` method to two ref-qualified methods. It can prevent retrieving const lvalue reference from rvalue RAII handle.

Before

```c++
vk::raii::ShaderModule shaderModule { ... };
*shaderModule // const vk::ShaderModule&

*vk::raii::ShaderModule { ... } // const vk::ShaderModule&
*std::move(shaderModule) // const vk::ShaderModule&
```

After

```c++
vk::raii::ShaderModule shaderModule { ... };
*shaderModule // const vk::ShaderModule&

*vk::raii::ShaderModule { ... } // vk::ShaderModule
*std::move(shaderModule) // vk::ShaderModule
```

### DIFF
--- a/VulkanHppGenerator.cpp
+++ b/VulkanHppGenerator.cpp
@@ -8970,7 +8970,12 @@ ${moveAssignmentInstructions}
       return *this;
     }
 
-    VULKAN_HPP_NAMESPACE::${handleType} const & operator*() const VULKAN_HPP_NOEXCEPT
+    VULKAN_HPP_NAMESPACE::${handleType} const & operator*() const & VULKAN_HPP_NOEXCEPT
+    {
+      return m_${handleName};
+    }
+
+    VULKAN_HPP_NAMESPACE::${handleType} operator*() const && VULKAN_HPP_NOEXCEPT
     {
       return m_${handleName};
     }
--- a/vulkan/vulkan_raii.hpp
+++ b/vulkan/vulkan_raii.hpp
@@ -3099,7 +3099,12 @@ namespace VULKAN_HPP_NAMESPACE
         return *this;
       }
 
-      VULKAN_HPP_NAMESPACE::Instance const & operator*() const VULKAN_HPP_NOEXCEPT
+      VULKAN_HPP_NAMESPACE::Instance const & operator*() const & VULKAN_HPP_NOEXCEPT
+      {
+        return m_instance;
+      }
+
+      VULKAN_HPP_NAMESPACE::Instance operator*() const && VULKAN_HPP_NOEXCEPT
       {
         return m_instance;
       }
@@ -3424,7 +3429,12 @@ namespace VULKAN_HPP_NAMESPACE
         return *this;
       }
 
-      VULKAN_HPP_NAMESPACE::PhysicalDevice const & operator*() const VULKAN_HPP_NOEXCEPT
+      VULKAN_HPP_NAMESPACE::PhysicalDevice const & operator*() const & VULKAN_HPP_NOEXCEPT
+      {
+        return m_physicalDevice;
+      }
+
+      VULKAN_HPP_NAMESPACE::PhysicalDevice operator*() const && VULKAN_HPP_NOEXCEPT
       {
         return m_physicalDevice;
       }
@@ -4090,7 +4100,12 @@ namespace VULKAN_HPP_NAMESPACE
         return *this;
       }
 
-      VULKAN_HPP_NAMESPACE::Device const & operator*() const VULKAN_HPP_NOEXCEPT
+      VULKAN_HPP_NAMESPACE::Device const & operator*() const & VULKAN_HPP_NOEXCEPT
+      {
+        return m_device;
+      }
+
+      VULKAN_HPP_NAMESPACE::Device operator*() const && VULKAN_HPP_NOEXCEPT
       {
         return m_device;
       }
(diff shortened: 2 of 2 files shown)


asuessenbach:
@stripe2933: Could you provide some code snippet that does not compile with what we currently have and would compile with your changes?

stripe2933:
I tested all CMake targets in `RAII_Samples` folder and everything built successfully, so nothing seems breaking with the current code.

asuessenbach:
> I tested all CMake targets in RAII_Samples folder and everything built successfully, so nothing seems breaking with the current code.

Sure. But I asked for what we gain with your change. What we currently can't do (or can't as easily do), but can do with your change.

stripe2933:
This change is not aimed to enhance the functionality. Instead, it can prevent the lifetime issue, what like we differentiate `vk::ArrayProxy` and `vk::ArrayProxyNoTemporaries`.

```c++
std::addressof(<T&>) // valid
std::addressof(<const T&&>) // invalid; prevent taking the address of const rvalues

std::span { <std::ranges::contiguous_range auto&> } // valid
std::span { <std::ranges::contiguous_range auto const&&> } // invalid, as making view from temporary is bad

vk::ArrayProxyNoTemporaries { <std::ranges::contiguous_range auto&> } // valid
vk::ArrayProxyNoTemporaries { <std::ranges::contiguous_range auto const&&> } // invalid, as making view from temporary is bad
```

All of these functions are making pointer/reference from the lvalue reference and are preventing making from temporary(rvalue reference). It is unintuitive that allowing making a reference from a temporary.

```c++
vk::SubmitInfo submitInfo {
    .waitSemaphores = { *vk::raii::Semaphore { ... } }, // Before changes, *vk::raii::Semaphore returns const vk::Semaphore&, therefore can be passed to vk::ArrayProxyNoTemporaries<const vk::Semaphore>.
    ....
}; // Here the created semaphore will be destroyed as the context end. We want vk::ArrayProxyNoTemporaries to prevent this lifetime issue, but failed.
queue.submit(submitInfo); // dangling
```

STL containers also differentiates the inner value accessing methods by the ref-qualifier. [`std::optional::operator*`](https://en.cppreference.com/w/cpp/utility/optional/operator*) [`std::expected::operator*`](https://en.cppreference.com/w/cpp/utility/expected/operator*)

asuessenbach:
Note that in the examples from the std-namespace, the `operator*() const&&` returns a `const T&& `, not a `T` as in your PR.

And also note, that something like
```
vk::SubmitInfo submitInfo {
    .waitSemaphores = { *vk::raii::Semaphore { ... } }, // Before changes, *vk::raii::Semaphore returns const vk::Semaphore&, therefore can be passed to vk::ArrayProxyNoTemporaries<const vk::Semaphore>.
    ....
}; // Here the created semaphore will be destroyed as the context end. We want vk::ArrayProxyNoTemporaries to prevent this lifetime issue, but failed.
queue.submit(submitInfo); // dangling
```
seems to be highly dangerous!
A `vk::raii::Semaphore` is supposed to destroy the actual semaphore on its destruction. `vk::ArrayProxyNoTemporaries` does not own the data and thus is not supposed to keep its elements alive.

stripe2933:
> Note that in the examples from the std-namespace, the `operator*() const&&` returns a `const T&& `, not a `T` as in your PR.

Okay, I fixed it to return the const rvalue reference.

> And also note, that something like
> 
> ```
> vk::SubmitInfo submitInfo {
>     .waitSemaphores = { *vk::raii::Semaphore { ... } }, // Before changes, *vk::raii::Semaphore returns const vk::Semaphore&, therefore can be passed to vk::ArrayProxyNoTemporaries<const vk::Semaphore>.
>     ....
> }; // Here the created semaphore will be destroyed as the context end. We want vk::ArrayProxyNoTemporaries to prevent this lifetime issue, but failed.
> queue.submit(submitInfo); // dangling
> ```
> 
> seems to be highly dangerous! A `vk::raii::Semaphore` is supposed to destroy the actual semaphore on its destruction. `vk::ArrayProxyNoTemporaries` does not own the data and thus is not supposed to keep its elements alive.

Yes, what I means the PR can prevent the code like it. After this PR, the given code cannot be compiled.

asuessenbach:
That is, you mean, with the current state, you could get some code compiled, that should not be valid?
Could you please provide such a code snippet?

I tried this
`vk::SubmitInfo submitInfo( { *vk::raii::Semaphore{} } );`
and this
`vk::SubmitInfo submitInfo{ .pWaitSemaphores = { &*vk::raii::Semaphore{} } };`
And got some compile time error on attempting to reference a deleted function. Which is correct.

asuessenbach:
@stripe2933 : Would you mind to give us some example of "bad coding", that's currently possible but isn't anymore with your change?

stripe2933:
Sorry for late reply.

> All of these functions are making pointer/reference from the lvalue reference and are preventing making from temporary(rvalue reference). It is unintuitive that allowing making a reference from a temporary.
> 
> ```c++
> vk::SubmitInfo submitInfo {
>     .waitSemaphores = { *vk::raii::Semaphore { ... } }, // Before changes, *vk::raii::Semaphore returns const vk::Semaphore&, therefore can be passed to vk::ArrayProxyNoTemporaries<const vk::Semaphore>.
>     ....
> }; // Here the created semaphore will be destroyed as the context end. We want vk::ArrayProxyNoTemporaries to prevent this lifetime issue, but failed.
> queue.submit(submitInfo); // dangling
> ```

I gave you wrong example. It uses `{ *vk::raii::Semaphore { ... } }`, which constructs `std::initializer_list<vk::Semaphore>` and safely rejected from `vk::ArrayProxyNoTemporaries<const T>` constructor.

---

 What my code of intention is:

```c++
// Note: using enhanced feature, designated initializer will not be available.
vk::SubmitInfo submitInfo {
    /* waitSemaphores= */ *vk::raii::Semaphore { ... }, // type=const vk::Semaphore&, will be converted to vk::ArrayProxyNoTemporaries<const vk::Semaphore>
    ...
}; // Now submitInfo references dangling vk::Semaphore*
```

will be compiled and cause lifetime issue.

asuessenbach:
Yes, you're right!

Thanks a lot for spotting this issue and providing a fix!

PR #2162, opened 2025-05-10
Add support for vulkan_video.cppm module

btmxh:
This PR add generation code for a new file `vulkan_video.cpp`, exporting everything defined in `vulkan_video.hpp` to the `vulkan_video_hpp` module.

Though, there are a couple issues I'm seeing with the current `vulkan_video.hpp`, not sure if any of those is not just an oversight of me:
- `#include <vk_video/vulkan_video_codec_av1std_encode.h>` is guarded under `301 < VK_HEADER_VERSION`, though its structs are defined regardless without no protections.
- `vulkan_video.hpp` does not wrap the defines in `vk_video/*.h` files, such as `STD_VIDEO_H265_CPB_CNT_LIST_SIZE`. Though, these values are mostly array sizes, which can be retrieved from its respective struct in one way or another. `vulkan_video_hpp` currently exports none of those defines.


asuessenbach:
> `#include <vk_video/vulkan_video_codec_av1std_encode.h>` is guarded under `301 < VK_HEADER_VERSIO`N, though its structs are defined regardless without no protections.

This guard was manually added back in the days when this header was added. As it's not backed by the current `video.xml`, I'll remove that guard.

> `vulkan_video.hpp` does not wrap the defines in `vk_video/*.h` files, such as `STD_VIDEO_H265_CPB_CNT_LIST_SIZE`.

Right, should be added!

btmxh:
> This guard was manually added back in the days when this header was added. As it's not backed by the current `video.xml`, I'll remove that guard.

Right, I was expecting that too, but that line's been there for quite a while, so I didn't think it was the case.

Also, should I make a CI test for the generated module file? It currently compiled fine on my machine, but I'm not sure that it works in different situations (e.g. separated from `vulkan.cppm`, different compilers, etc.).

asuessenbach:
> Also, should I make a CI test for the generated module file?

Now that you ask: would be a perfect fit!

M2-TE:
Adding a new target is a good idea, but I doubt it needs to be guarded by an additional option

asuessenbach:
> Adding a new target is a good idea, but I doubt it needs to be guarded by an additional option

I think, @M2-TE is right: having a new target for vulkan_video.cppm is the right path, but no need for an additional option, just use `VULKAN_HPP_ENABLE_CPP20_MODULES` here as well.

btmxh:
Added.

asuessenbach:
Looks great now, thanks a lot for your contribution!

Only one nitpicking, yet: maybe you could introduce some function `vulkan_hpp__setup_module`, or so, in the `CMakeLists.txt`, to handle the more or less identical setup of `VulkanHppModule` and `VulkanVideoHppModule`?

btmxh:
Done, though I see a lot of these lines in CMakeLists.txt:
```cmake
cmake_parse_arguments( TARGET "{options}" "${oneValueArgs}" "${multiValueArgs}" ${ARGN} )
```
Is this some kind of typo?

CMakeLists.txt
```diff
@@ -283,31 +283,66 @@ endif()
...
 if( VULKAN_HPP_ENABLE_CPP20_MODULES )
-	if( VULKAN_HPP_ENABLE_STD_MODULE AND CMAKE_VERSION VERSION_LESS "3.30" )
-		message( FATAL_ERROR "Vulkan-Hpp: C++20 modules with import std require CMake 3.30 or later" )
-	elseif ( CMAKE_VERSION VERSION_LESS "3.28" )
-		message( FATAL_ERROR "Vulkan-Hpp: C++20 modules require CMake 3.28 or later" )
-	endif()
+	function ( vulkan_hpp__setup_module )
+		set( options )
+		set( oneValueArgs NAME )
+		set( multiValueArgs FILE_SET ALIAS )
```
asuessenbach, line 289: Why do you make `ALIAS` a `multiValueArg`? Are you foreseeing multiple aliases?

CMakeLists.txt
```diff
@@ -283,31 +283,66 @@ endif()
...
+		cmake_parse_arguments( TARGET "${options}" "${oneValueArgs}" "${multiValueArgs}" "${ARGN}" )
```
asuessenbach, line 290: It seems to work, but there's no need to have quotation marks around `${ARGN}`.

CMakeLists.txt
```diff
@@ -283,31 +283,66 @@ endif()
...
-		FILE_SET vulkan_module_file
-		BASE_DIRS ${CMAKE_CURRENT_SOURCE_DIR}
-		TYPE CXX_MODULES
-		FILES vulkan/vulkan.cppm )
-	target_link_libraries( VulkanHppModule PUBLIC Vulkan::Hpp )
+		if ( VULKAN_HPP_ENABLE_STD_MODULE AND CMAKE_VERSION VERSION_LESS "3.30" )
+			message( FATAL_ERROR "Vulkan-Hpp: C++20 modules with import std require CMake 3.30 or later" )
+		elseif ( CMAKE_VERSION VERSION_LESS "3.28" )
+			message( FATAL_ERROR "Vulkan-Hpp: C++20 modules require CMake 3.28 or later" )
+		endif()
```
asuessenbach, line 300: This part is independent of any of the arguments to this function. Should be moved out of it.

CMakeLists.txt
```diff
@@ -283,31 +283,66 @@ endif()
...
+		if ( NOT TARGET_FILE_SET )
+			message( FATAL_ERROR "FILE_SET must be defined in vulkan_hpp__setup_module" )
+		endif()
+
+		list(LENGTH TARGET_FILE_SET TARGET_FILE_SET_LENGTH)
+		if ( TARGET_FILE_SET_LENGTH LESS 2 )
+			message(FATAL_ERROR "FILE_SET in vulkan_hpp__setup_module must contain the file set name and at least one source file")
+		endif()
+
+		list( GET TARGET_FILE_SET 0 TARGET_FILE_SET_NAME )
```
asuessenbach, line 328: Why do you list the `FILE_SET_NAME` as the first element of the `FILE_SET`? I would prefer to have an additional `oneValueArg`, instead.

asuessenbach (commented):
In addition to the four issues below, you should add `vulkan_video.cppm` to the `VK_GENERATED_VULKAN_HEADERS` list near the end of the `CMakeLists.txt`.

CMakeLists.txt
btmxh: I initially was making ALIAS optional, and I thought it wouldn't hurt to support multiple aliases as well. I just need to change `if` to `for`, so I went with it.

CMakeLists.txt
btmxh: Was thinking the same, but the check would be unnecessary if `vulkan_hpp__setup_module` is not called. I agree that currently it should be checked regardless in the `VULKAN_HPP_ENABLE_CPP20_MODULES` if block, but the choice to left it in the function was mostly me speculating about how it could be used in the future.
On the other hand, the function is now scoped to that if block, so maybe I should move this out after all.

CMakeLists.txt
btmxh: This is mostly to shorten the code, but now that I've taken another look, I do agree with it being somewhat less readable than a separated argument.

CMakeLists.txt
btmxh: Will fix, it's just me having the habit of excessively quoting things in CMake.

btmxh:
All done (except the `ALIAS` bit). I think we don't need to change `ALIAS` to single-value, as it would not yield any particular benefits.

asuessenbach:
Looks good now. Thanks a lot for your contribution.

PR #2163, opened 2025-05-12
Add new functions getColorFormats, hasRedComponent, hasGreenComponent, hasBlueComponent, and hasAlphaComponent to vulkan_format_traits.hpp

asuessenbach:


vulkan/vulkan_format_traits.hpp
```diff
@@ -6800,6 +6815,254 @@ namespace VULKAN_HPP_NAMESPACE
     }
   }
 
+  // Get all formats with a color component
+  VULKAN_HPP_INLINE std::vector<VULKAN_HPP_NAMESPACE::Format> const & getColorFormats()
+  {
+    static std::vetor<VULKAN_HPP_NAMESPACE::Format> colorFormats = { VULKAN_HPP_NAMESPACE::Format::eR4G4UnormPack8,
```
n0F4x, line 6821: Typo: `vetor` -> `vector`. And shouldn't this be an array and constexpr?

asuessenbach: > And shouldn't this be an array and constexpr?

Yes, that's a good idea.

n0F4x: `colorFormats` should be static.

Issue #2164, opened 2025-05-12
`vulkan.cppm` fails to compile when using `std` module after upgrading to Vulkan-Headers 1.4.313.0


btmxh:
Previously I was using Vulkan-Headers 1.4.309.0, which has
```cpp
#if ( VULKAN_HPP_ASSERT == assert )
#  include <cassert>
#endif
```

But following #2134, `<cassert>` is included along with other headers, so it is only included when `VULKAN_HPP_ENABLE_STD_MODULE` and `VULKAN_HPP_STD_MODULE` is not defined. However, `import std.compat` does not include macros, so `assert` is undefined and compilation fails. Since CI runners do not run the `CppStdModule.cpp` test (afaik), this probably has gone unnoticed for quite some time now.

The simplest way to fix this that I could think of is to either include `<cassert>` unconditionally, or in `vulkan_hpp_macros.hpp` immediately when `VULKAN_HPP_ASSERT` is defined to be `assert`, like so:
```cpp
#if !defined( VULKAN_HPP_ASSERT )
#  include <cassert>
#  define VULKAN_HPP_ASSERT assert
#endif
```

PR #2165, opened 2025-05-12
Fix undefined `assert` when using `std` module

btmxh:
Fix the compilation error problem in #2164 using what's described in the issue.

### DIFF
--- a/vulkan/vulkan.hpp
+++ b/vulkan/vulkan.hpp
@@ -11,11 +11,13 @@
 #include <vulkan/vulkan_hpp_macros.hpp>
 
 #if defined( VULKAN_HPP_ENABLE_STD_MODULE ) && defined( VULKAN_HPP_STD_MODULE )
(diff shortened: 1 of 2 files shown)


vulkan/vulkan.hpp
```diff
@@ -15,8 +15,7 @@
 import VULKAN_HPP_STD_MODULE;
```
M2-TE, line 15: Would it not be better to include `cassert` directly after `string.h` and before `import VULKAN_HPP_STD_MODULE` here, while leaving the original one?
It would cleanly group the std module vs std includes (especially considering that `string.h` is another such exception to import std)

M2-TE: I mean, they are not necessary included twice, its two separate branching paths. It's nice to isolate these, makes them easy to change later without thinking about whether or not it breaks some other macro combination

asuessenbach:

PR #2168, opened 2025-05-14
Always define relational operators for handles

### LINKED ISSUE #2167: Equality operator is not defined when three-way comparison operator is not defaulted
In `include/vulkan/vulkan_handles.hpp` file, `VULKAN_HPP_HAS_SPACESHIP_OPERATOR` is used to declare <=> operator and skip definitions for other relational operators.

But in some compilers, when spaceship operator is not defaulted, operator <=> does not automatically synthesize operator==, and this can cause compilation errors like:

```
vulkan/include/vulkan/vulkan_structs.hpp:39950:24: error: invalid operands to binary expression ('const vkhpp::DisplayKHR' and 'const vkhpp::DisplayKHR')
 39950 |       return ( display == rhs.display ) && ( ( displayName == rhs.displayName ) || ( strcmp( displayName, rhs.displayName ) == 0 ) ) &&
       |                ~~~~~~~ ^  ~~~~~~~~~~~
```

**Code**

```cpp
  // operators to compare VULKAN_HPP_NAMESPACE::-handles
#if defined( VULKAN_HPP_HAS_SPACESHIP_OPERATOR )
  template <typename T, typename std::enable_if<VULKAN_HPP_NAMESPACE::isVulkanHandleType<T>::value, int>::type = 0>
  auto operator<=>( T const & lhs, T const & rhs )
  {
    return static_cast<typename T::NativeType>( lhs ) <=> static_cast<typename T::NativeType>( rhs );
  }
#else //  <--- moving '#endif' here fixes the issue
  template <typename T, typename std::enable_if<VULKAN_HPP_NAMESPACE::isVulkanHandleType<T>::value, int>::type = 0>
  bool operator==( T const & lhs, T const & rhs )
  {
    return static_cast<typename T::NativeType>( lhs ) == static_cast<typename T::NativeType>( rhs );
  }

  template <typename T, typename std::enable_if<VULKAN_HPP_NAMESPACE::isVulkanHandleType<T>::value, int>::type = 0>
  bool operator!=( T const & lhs, T const & rhs )
  {
    return static_cast<typename T::NativeType>( lhs ) != static_cast<typename T::NativeType>( rhs );
  }

  template <typename T, typename std::enable_if<VULKAN_HPP_NAMESPACE::isVulkanHandleType<T>::value, int>::type = 0>
  bool operator<( T const & lhs, T const & rhs )
  {
    return static_cast<typename T::NativeType>( lhs ) < static_cast<typename T::NativeType>( rhs );
  }
#endif

```

kocdemir:
Fixes #2167

### DIFF
--- a/VulkanHppGenerator.cpp
+++ b/VulkanHppGenerator.cpp
@@ -10496,7 +10496,8 @@ ${raiiHandles}
   {
     return *a <=> *b;
   }
-#else
+#endif
+
   template <typename T, typename std::enable_if<VULKAN_HPP_NAMESPACE::VULKAN_HPP_RAII_NAMESPACE::isVulkanRAIIHandleType<T>::value,bool>::type = 0>
   bool operator==( T const & a, T const & b ) VULKAN_HPP_NOEXCEPT
   {
@@ -10514,7 +10515,6 @@ ${raiiHandles}
   {
     return *a < *b;
   }
-#endif
 
   template <typename T, typename std::enable_if<VULKAN_HPP_NAMESPACE::VULKAN_HPP_RAII_NAMESPACE::isVulkanRAIIHandleType<T>::value,bool>::type = 0>
   bool operator==( const T & v, std::nullptr_t ) VULKAN_HPP_NOEXCEPT


M2-TE:
Your change would only affect RAII structures, is that intended?
Also please run the generator so we can see the resulting changes in the actual header

asuessenbach:
In fact, I think, those operators aren't needed at all!
I've crafted a little PR that does some cleanup work on all those compare operators with #2169. Would be great, if @kocdemir  could verify that that works with your use case.

kocdemir:
Sure, I'll kick off some builds with those changes and let you know.

kocdemir:
I applied the diff onto header version 313 we're using on this file: https://cs.android.com/android/platform/superproject/main/+/main:hardware/google/gfxstream/common/detector/Vulkan.h;l=30

which includes the headers like this: 
```
#define VULKAN_HPP_NAMESPACE vkhpp
#define VULKAN_HPP_DISPATCH_LOADER_DYNAMIC 1
#define VULKAN_HPP_ENABLE_DYNAMIC_LOADER_TOOL 1
#define VULKAN_HPP_NO_CONSTRUCTORS
#define VULKAN_HPP_NO_EXCEPTIONS
#define VULKAN_HPP_ASSERT_ON_RESULT
#include <vulkan/vulkan_raii.hpp>
#include <vulkan/vulkan_to_string.hpp>
```

Getting the following warnings:
 
```
In file included from hardware/google/gfxstream/common/detector/Vulkan.h:30:
In file included from hardware/google/gfxstream/third_party/vulkan/include/vulkan/vulkan.hpp:9093:
hardware/google/gfxstream/third_party/vulkan/include/vulkan/vulkan_structs.hpp:1341:10: error: explicitly defaulted three-way comparison operator is implicitly deleted [-Werror,-Wdefaulted-function-deleted]
 1341 |     auto operator<=>( AccelerationStructureCaptureDescriptorDataInfoEXT const & ) const = default;
      |          ^
hardware/google/gfxstream/third_party/vulkan/include/vulkan/vulkan_structs.hpp:1362:52: note: defaulted 'operator<=>' is implicitly deleted because there is no viable three-way comparison function for member 'accelerationStructure'
 1362 |     VULKAN_HPP_NAMESPACE::AccelerationStructureKHR accelerationStructure   = {};
      |                                                    ^
hardware/google/gfxstream/third_party/vulkan/include/vulkan/vulkan.hpp:163:24: note: candidate template ignored: could not match 'const ArrayWrapper1D<char, N>' against 'const vkhpp::AccelerationStructureKHR'
  163 |   std::strong_ordering operator<=>( ArrayWrapper1D<char, N> const & lhs, ArrayWrapper1D<char, M> const & rhs ) VULKAN_HPP_NOEXCEPT
      |                        ^
hardware/google/gfxstream/third_party/vulkan/include/vulkan/vulkan.hpp:163:24: note: candidate template ignored: could not match 'const ArrayWrapper1D<char, N>' against 'const vkhpp::AccelerationStructureKHR'
hardware/google/gfxstream/third_party/vulkan/include/vulkan/vulkan.hpp:209:24: note: candidate template ignored: could not match 'const ArrayWrapper1D<char, N>' against 'const vkhpp::AccelerationStructureKHR'
  209 |   std::strong_ordering operator<=>( std::string const & lhs, ArrayWrapper1D<char, N> const & rhs ) VULKAN_HPP_NOEXCEPT
      |                        ^
hardware/google/gfxstream/third_party/vulkan/include/vulkan/vulkan.hpp:209:24: note: candidate template ignored: could not match 'const ArrayWrapper1D<char, N>' against 'const vkhpp::AccelerationStructureKHR'
hardware/google/gfxstream/third_party/vulkan/include/vulkan/vulkan.hpp:253:24: note: candidate template ignored: could not match 'const ArrayWrapper1D<char, N>' against 'const vkhpp::AccelerationStructureKHR'
  253 |   std::strong_ordering operator<=>( ArrayWrapper1D<char, N> const & lhs, std::string const & rhs ) VULKAN_HPP_NOEXCEPT
      |                        ^
hardware/google/gfxstream/third_party/vulkan/include/vulkan/vulkan.hpp:253:24: note: candidate template ignored: could not match 'const ArrayWrapper1D<char, N>' against 'const vkhpp::AccelerationStructureKHR'
hardware/google/gfxstream/third_party/vulkan/include/vulkan/vulkan_structs.hpp:1341:91: note: replace 'default' with 'delete'
 1341 |     auto operator<=>( AccelerationStructureCaptureDescriptorDataInfoEXT const & ) const = default;
      |                                                                                           ^~~~~~~
      |                                                                                           delete
hardware/google/gfxstream/third_party/vulkan/include/vulkan/vulkan_structs.hpp:1501:10: error: explicitly defaulted three-way comparison operator is implicitly deleted [-Werror,-Wdefaulted-function-deleted]
 1501 |     auto operator<=>( AccelerationStructureCreateInfoKHR const & ) const = default;
      |          ^
hardware/google/gfxstream/third_party/vulkan/include/vulkan/vulkan_structs.hpp:1523:63: note: defaulted 'operator<=>' is implicitly deleted because there is no viable three-way comparison function for member 'buffer'
 1523 |     VULKAN_HPP_NAMESPACE::Buffer                              buffer        = {};
      |                                                               ^
....
...
..
```

asuessenbach:
@kocdemir: ok, the handles need to have at least a defaulted spaceship operator in that situation...
Could you please re-check again if your issue is resolved with my latest PR in #2169?

kocdemir:
Yes, that fixed the issue and the builds didn't produce errors. I'll close this one in favor of that PR then, thank you!

PR #2171, opened 2025-05-18
Added using declaration for SwapchainOwns enum to Module File

### LINKED ISSUE #2170: vk::SwapchainOwns SharedImage flag not exported in C++20 Module
In the examples provided in both the ReadMe and the SharedHandles.cpp sample code, the vk::SwapchainOwns flag is provided as a way of assigning ownership of a SharedImage to the swapchain. When using the C++20 Module, this flag is not visible to intellisense or the compiler. This appears to be a result of the flag enum class not being declared in the .cppm file with a using declaration.

Arialonomus:
This PR adds a using declaration for the SwapchainOwns enum flag, which is utilized in the construction of SharedImages to assign ownership to the SwapchainKHR object. Previously, the definition of this enum class was invisible to the compiler and intellisense if using the C++20 module.

fixes #2170

### DIFF
--- a/vulkan/vulkan.cppm
+++ b/vulkan/vulkan.cppm
@@ -5159,6 +5159,9 @@ export namespace VULKAN_HPP_NAMESPACE
   //=== VK_KHR_swapchain ===
   using VULKAN_HPP_NAMESPACE::SharedSwapchainKHR;
 
+  //=== SharedImage SwapchainOwns enum
+  using VULKAN_HPP_NAMESPACE::SwapchainOwns;
+
   //=== VK_KHR_display ===
   using VULKAN_HPP_NAMESPACE::SharedDisplayKHR;
   using VULKAN_HPP_NAMESPACE::SharedDisplayModeKHR;


M2-TE:
The module file is generated, so you would need to edit the generator code and run it.
Shared handle exports for it are generated here:

https://github.com/KhronosGroup/Vulkan-Hpp/blob/d5b553d969cf73b0bd321ae2d1d01b92aabac366/VulkanHppGenerator.cpp#L5464-L5509

A somewhat quick and dirty solution would be to add something like this after the `SharedHandleTraits` export string:
```cpp
sharedHandleUsings += R"(
  //=== VK_KHR_swapchain additional enum ===
  using VULKAN_HPP_NAMESPACE::SwapchainOwns;
)";
```

where a less hacky solution would be adding the export automatically when extension iteration goes over `VK_KHR_swapchain`

Arialonomus:
Ah gotcha! I probably rushed in a bit too fast on this one, since it seemed like a quick fix. I'll look at the generator code and make the adjustment that way. Thank you for the feedback!

PR #2180, opened 2025-06-12
Include missing header type_traits to vulkan_enums.hpp

### LINKED ISSUE #2178: Missing type_traits include in vulkan_enums.hpp
The `vulkan_enums.hpp` header uses `std::underlying_type` but the header is missing `#include <type_traits>`.

asuessenbach:
Resolves #2178.


dsvensson:
@asuessenbach This breaks std module with at least MSVC. Would be nice with some cppm compile smoke testing in CI to catch issues like this as it seem to happen every now and then.

PR #2192, opened 2025-06-20
README: Added CI test result badges

M2-TE:
Might as well add these


asuessenbach:
That's a nice little addition. Thanks a lot.

PR #2212, opened 2025-07-08
CPPM: Removed faulty macro guard for exporting dynamic dispatch in C++20-module

M2-TE:
Fixed as pointed out by @dsvensson in https://github.com/KhronosGroup/Vulkan-Hpp/pull/2146#issuecomment-3049309781.

Unclear why it even worked in the first place, but this is the correct solution. Export was only necessary for MSVC as far as I can tell.

### DIFF
--- a/VulkanHppGenerator.cpp
+++ b/VulkanHppGenerator.cpp
@@ -5709,10 +5709,8 @@ std::string VulkanHppGenerator::generateCppModuleUsings() const
   {
     using VULKAN_HPP_NAMESPACE::detail::DispatchLoaderBase;
     using VULKAN_HPP_NAMESPACE::detail::DispatchLoaderDynamic;
-#if !defined( VULKAN_HPP_DEFAULT_DISPATCHER )
-# if VULKAN_HPP_DISPATCH_LOADER_DYNAMIC == 1
+#if VULKAN_HPP_DISPATCH_LOADER_DYNAMIC == 1
     using VULKAN_HPP_NAMESPACE::detail::defaultDispatchLoaderDynamic;
-# endif
 #endif
 #if !defined( VK_NO_PROTOTYPES )
     using VULKAN_HPP_NAMESPACE::detail::DispatchLoaderStatic;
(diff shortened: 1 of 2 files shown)


asuessenbach:
Well, unfortunately, that's also not correct.
The `VULKAN_HPP_DEFAULT_DISPATCHER` allows to specify a custom default dispatcher by defining it before including vulkan.hpp (or importing vulkan.cppm).
After the macros have been digested by the compiler (which is obviously done before all the cppm stuff is listed), it's definitely defined, but you can't rely on its actual value. It might be that `defaultDispatchLoader*` or something user provided.

We could define some `VULKAN_HPP_STANDARD_DEFAULT_DISPATCHER_USED`, or so, in macros.hpp:
```
#if !defined( VULKAN_HPP_DEFAULT_DISPATCHER )
#  define VULKAN_HPP_STANDARD_DEFAULT_DISPATCHER_USED
#  if VULKAN_HPP_DISPATCH_LOADER_DYNAMIC == 1
#    define VULKAN_HPP_DEFAULT_DISPATCHER ::VULKAN_HPP_NAMESPACE::detail::defaultDispatchLoaderDynamic
...
```
Then you could check for that `VULKAN_HPP_STANDARD_DEFAULT_DISPATCHER_USED` in vulkan.cppm:
```
#if defined( VULKAN_HPP_STANDARD_DEFAULT_DISPATCHER_USED)
# if VULKAN_HPP_DISPATCH_LOADER_DYNAMIC == 1
```

M2-TE:
As far as I can tell, `VULKAN_HPP_DISPATCH_LOADER_DYNAMIC` should remain defined as 0, even with a custom `VULKAN_HPP_DEFAULT_DISPATCHER`. If the default dynamic dispatcher is manually set as the default dispatcher, the former will have been set to 1 anyways and the symbol export would be required.

In your example, you define `VULKAN_HPP_STANDARD_DEFAULT_DISPATCHER_USED` for both static and dynamic cases, but I think we only really care about the `defaultDispatchLoaderDynamic` export here, so it could just be moved into the proper macro guards.

~But now that I think about it, can't we simply check for whether `VULKAN_HPP_DEFAULT_DISPATCH_LOADER_DYNAMIC_STORAGE` is defined? That should only be the case when we need to export `defaultDispatchLoaderDynamic` and will not require an additional define.~ It gets defined as empty with the static dispatcher so this would not work.

asuessenbach:
> As far as I can tell, `VULKAN_HPP_DISPATCH_LOADER_DYNAMIC` should remain defined as 0, even with a custom `VULKAN_HPP_DEFAULT_DISPATCHER`.

Yep, you're absolutely right!
That is, your initial approach is correct, and my concern was wrong.

Thanks a lot for you contribution, again!

PR #2218, opened 2025-07-16
Tests: Fixed C++20 module test for gcc15

M2-TE:
The module implementation for gcc is finally somewhat stable in gcc15 and it can now compile the vulkan module. Just have to make sure to `import` things after `include` in `.cpp` files, otherwise there will be symbol clashes.
Can include gcc tests once there's a runner with gcc15 available (likely going to be Ubuntu 26).

I've also used the `defaultDispatchLoaderDynamic` in the tests directly instead of relying on macros, as that is more in line with how modules are supposed to work. Rmoves the need to include the macro header. Might be an argument for pulling it out of the `detail` namespace?

I will not touch the `import std` tests for now, as I don't have a working setup for those yet. They should work with these changes though, completely removing the need for any `#include` directives.

### DIFF
--- a/tests/Cpp20Modules/Cpp20Modules.cpp
+++ b/tests/Cpp20Modules/Cpp20Modules.cpp
@@ -15,18 +15,19 @@
 // VulkanHpp Test : Cpp20Module
 //                  Compile test on using c++20 modules
 
-import vulkan_hpp;
-
 #include <memory>   // std::unique_ptr (seems to be needed on Windows)
 #include <string>   // std::string
 #include <iostream> // std::cout
-#include <vulkan/vulkan_hpp_macros.hpp>
+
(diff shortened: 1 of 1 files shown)


asuessenbach:
> Might be an argument for pulling it out of the detail namespace?

Well, not exactly. Those dispatchers are not backed by the vulkan spec but just, well, implementation details for the C++-bindings here. And I think that about 99% of their usage was hidden by some macros that are supposed to do the right thing.

There are a couple of other entities, that are not backed by the vulkan spec and thus should be in the details namespace, like the `vk::StructureChain` or the `vk::ArrayProxy`. But they are too visible to the users to be moved now.

Besides that: Thanks again for your ongoing contribution!

PR #2235, opened 2025-08-01
Include deprecated members in struct ctors

### LINKED ISSUE #2233: Vulkan-Headers CI tests failing against current Vulkan-Docs branch
There are a bunch of errors raised in vulkan_structs.hpp about "vk::DeviceCreateInfo::<membername> must be initialized by mem-initializer in 'constexpr' constructor". This is holding up the Vulkan-Headers update to match to spec update so would appreciate help @asuessenbach . See e,g,

https://github.com/KhronosGroup/Vulkan-Headers/actions/runs/16665329727/job/47170884324

M2-TE:
Might resolve #2233, as constexpr ctor needs to init these members.
This change does not seem to trigger the deprecation warnings in my local testing.

### DIFF
--- a/VulkanHppGenerator.cpp
+++ b/VulkanHppGenerator.cpp
@@ -11509,17 +11509,14 @@ ${popIgnored}
       arguments.push_back( argument );
     }
 
-    if ( member.deprecated.empty() )
-    {
-      // gather the initializers; skip members with exactly one legal value
-      if ( member.value.empty() )
-      {
-        initializers.push_back( member.name + "{ " + member.name + "_ }" );
-      }
-    }
-    else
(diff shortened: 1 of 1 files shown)


asuessenbach:
Thanks for jumping in here!
It's correct, you can assign the deprecated elements without any warnings here, because the warnings are turned off at that point.
But you then also don't need to `detail::ignore` them anymore.
Would you please remove that part as well?

asuessenbach:
... never mind. I already fixed that by a #2236.

Thanks for your contribution, anyways.

M2-TE:
Oh yea, I forgot about the detail::ignore.
No problem though, glad to hear it works!

PR #2242, opened 2025-08-06
CMake: Only set CMAKE_CXX_STANDARD when not already set before

### LINKED ISSUE #2237: (Edit) Which C++ versions should Vulkan spec repo test Hpp headers against?
There's a simple test in

https://github.com/KhronosGroup/Vulkan-Docs/blob/main/tests/hpptest.cpp

that gets compiled against the headers built from Vulkan-Hpp as part of CI. I'm not sure why this did not fail in e.g. https://github.com/KhronosGroup/Vulkan-Docs/actions/runs/16665603415/job/47171453856 (much less the internal gitlab CI which should have caught it earlier). @asuessenbach could you look at the test file and see if there's any apparent reason it would not have compiled vulkan_structs.hpp and found the error, like the Vulkan-Headers test did?

M2-TE:
Without this change, CI will always compile with C++11. Should resolve #2237


asuessenbach:
Bummer!

Thanks a lot for spotting this nasty issue!

And the fact that the CI is failing now shows, it actually was a fix.

M2-TE:
> Thanks a lot for spotting this nasty issue!

Well, it looks like I was the one who introduced it in the first place back in January.. sorry about that!

Ubuntu CI has issues with mismatching compiler versions under C++20 modules, probably related to us setting explicit versions for it. I'll cook up a fix for that.

The Windows error looks a bit less fun to hunt down:
> INTERNAL COMPILER ERROR in 'C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Tools\MSVC\14.44.35207\bin\HostX64\x64\CL.exe'

but should also be related to module mismatches as per the messages somewhere above it.

M2-TE:
Tests were compiled explicitly with C++20, whereas vulkan modules followed the `CMAKE_CXX_STANDARD` (as they should). This breaks, when the `CMAKE_CXX_STANDARD` is > 20. It didn't really make sense to set explicit CXX versions for these tests anyways, they should follow the global standard as well.

The `cxx_std_20` or `cxx_std_23` compile features for the modules should be kept, as they only impose a minimum.

asuessenbach:
Thanks a lot for fixing this issue!

PR #2245, opened 2025-08-07
CMake: Standardize the use of `CMAKE_CXX_STANDARD` across tests

M2-TE:
Some tests were manually setting their `CXX_STANDARD` and should instead use the global `CMAKE_CXX_STANDARD` after checking whether it hits their minimum version threshhold. Should catch future issues where e.g. `Designated Initializers` break in a version above C++20.


M2-TE:
I will close this for now, as I am working on improving test coverage in a bigger PR, which will include these changes

PR #2247, opened 2025-08-07
Remove the superfluous operator<=> for handles.

asuessenbach:
Resolves #2246.

Issue #1815, opened 2024-02-27
C++20 module improvements


sharadhr:
There are a few things I'd like to improve about `vulkan.cppm`, stemming from my own experience, and bug reports around.

- Using any struct that declares/exports some sort of comparison or `operator<=>` requires that the consumer `#include <compare>`, which is very unintuitive; I'd class this as a bug. Some good resources:
  - [DevComm thread: Template exports requiring importing of dependent packages (C++20 modules)](https://developercommunity.visualstudio.com/t/Template-exports-requiring-importing-of-/1425979). The suggestion made was to `export import <compare>;` but header units are generally considered a bad idea.
  - [Reddit comment from a MSVC compiler developer with further detail, making a similar suggestion](https://www.reddit.com/r/cpp/comments/1b0zem7/what_is_the_state_of_modules_in_2024/ksc8ix8/), but this time suggests using `import std;` in the _producing_ module without exporting any additional symbol. Which brings me to...
- Using `import std;` breaks badly with `import vulkan-hpp;` because the latter still relies on `#include <some_std_header>` for all its dependencies. 
  - The general recommendation is to not mix standard headers and the standard library module.
  - I propose using `import std;` guarded behind the appropriate feature macro, [`__cpp_lib_modules`](https://en.cppreference.com/w/cpp/feature_test#cpp_lib_modules). 
  - It's worth noting that the std library module has been back-ported to C++20. 
    - MSVC STL: https://github.com/microsoft/STL/issues/3945
    - LLVM libc++: https://reviews.llvm.org/D158358
    - GNU libstdc++: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=106852 (in-progress)
    
I'd like to spur some discussion on this, and if it's worthwhile, start a relevant PR.

sharadhr:
@asuessenbach and @theHamsta, any thoughts on this?

g-martin772:
Making the vulkan_hpp module work with `import std;` would be great!

sharadhr:
@asuessenbach and @theHamsta, do you have any input on this? Thanks!

asuessenbach:
Please sorry for the delay.
I think, you're right, using `import std` seems to be the right move. Would be great if you could start a PR on it.

stripe2933:
[...]

PR #2248, opened 2025-08-08
Tests: Created std module versions of each test and enabled previously disabled tests

M2-TE:
This is a pretty big one, as I've adjusted the test setup. The main goals:

- Isolate tests as much as possible (being careful what to link, e.g. avoiding `utils` where possible)
- Create separate std module test for each currently existing test (using simple flag, no duplicate sources)
- Unify the CMakeLists.txt setup for each test, as the current setup (probably accidentally) has a few of them disabled.

### Test isolation
Some of the tests set macros inside the source itself, which will not be possible for the std module. This one I've not fully figured out how to solve, but that should be done in a separate PR. Some dependencies also had their compile definitions bleed into the tests (e.g. dynamic dispatcher from utils). Nothing was done to break these tests.

### C++20 module testing
Every test now also compiles with the `vulkan_hpp` module import, creating a separate executable for it. This required some fixes to the tests (e.g. `VK_API_VERSION_11` replaced by `vk::ApiVersion11`), as well as using the explicit dynamic dispatcher via `vk::details::defaultDispatchLoaderDynamic` rather than the macro. These module tests already caught a few edge cases that seem to break, in which case I've currently disabled the tests and put a comment as to why in the respective test's CMakeLists.txt.

### Test setup unification
Some tests had varying setups, like manually linking to vulkan libs, unix dynamic linker, only running when dynamic linker is globally enabled or straightup relying on find_vulkan to even compile the test (CI does not seem to find vulkan, so these were left disabled). I've straightened all these out, which enables some tests that were previously disabled. One of them also seems to catch something, namely the `DeviceFunctions` test catching problems with explicit usage of the static dispatcher.

As far as I can tell, no previously enabled tests are now disabled. Ones that I did not really touch are the `Video` `VideoCpp20Modules` and some of the `DispatchLoaderStatic` tests. These require better target isolation as std modules cannot "undefine" e.g. dynamic dispatcher without doing platform specific things (-U on unix, /U on MSVC) or doing sketchy hacks to remove the defines from the CMake properties.

The currently failing tests should be looked at in separate PRs that address each test individually. Most issues are one of:

- Problems with static dispatcher, but these won't run on CI anyways
- Missing symbols for the std module (would require using extra `#includes`, which should not be needed with the module)
- `ArrayWrapper`, `StridedArrayProxy`, `ArrayProxy*` and their variants break in the std modules

It would also be nice to change the `UniqueHandle` test to no longer require all the extra dependencies. It seems overkill for what the test needs and would also simplify the test setup for modules, which would be really messy right now. It is the only test that requires linking to `utils`.


sharadhr:
Thanks for continuing this, @M2-TE :)

Work has really absorbed all my available time; I was hoping to contribute a little more to Vulkan-Hpp if and when I could.

M2-TE:
Macos and Ubuntu22 builds are failing due to a lack of the dynamic dispatch flag in CI, should be fixed with #2254.

Windows build fails due to lack of `<compare>`, which is needed by `std::partial_ordering`. Automatically declared by the spaceship operator that was already removed in #2247, so won't require any intervention once I merge those into here.

M2-TE:
@asuessenbach there seems to be an issue with the spaceship operator and enums (e.g. `vk::Flags<...>`) when compiled with std modules. The issue seems to be the lack of `<compare>`, which I now include in `vulkan_enums.hpp`. Problem is, MSVC doesn't properly export it alongside `vk::Flags`, so users importing `vulkan_hpp` who also use the spaceship operator are left with implementations that do not have access to `std::partial_ordering`.

All in all, it is a compiler issue, rather than our fault (I believe so at least). For the test I can either:
1. Manually include `<compare>` before importing `vulkan_hpp`
2. Use `VULKAN_HPP_NO_SPACESHIP_OPERATOR` for the `flags.cpp` test
3. Disable the std module testing for the flags test

It will not solve the issue for users, but I'm not quite sure how else to do this. Any other ideas?

<details> <summary>MSVC error log</summary>

```bash
D:\a\Vulkan-Hpp\Vulkan-Hpp\vulkan\vulkan_enums.hpp(45,10): error C3774: cannot find 'std::partial_ordering': Please include <compare> header [D:\a\Vulkan-Hpp\Vulkan-Hpp\build\tests\Flags\Flags_module.vcxproj]
  (compiling source file '../../../tests/Flags/Flags.cpp')
      D:\a\Vulkan-Hpp\Vulkan-Hpp\vulkan\vulkan_enums.hpp(45,10):
      the template instantiation context (the oldest one first) is
          D:\a\Vulkan-Hpp\Vulkan-Hpp\tests\Flags\Flags.cpp(39,23):
          see reference to class template instantiation 'vk::Flags<vk::MemoryHeapFlagBits>' being compiled
  
  IntermTraverse.cpp
D:\a\Vulkan-Hpp\Vulkan-Hpp\vulkan\vulkan_enums.hpp(45,56): error C2610: 'auto vk::Flags<vk::MemoryHeapFlagBits>::operator <=>(const vk::Flags<vk::MemoryHeapFlagBits> &) const': is not a special member function or comparison operator which can be defaulted [D:\a\Vulkan-Hpp\Vulkan-Hpp\build\tests\Flags\Flags_module.vcxproj]
  Intermediate.cpp
  (compiling source file '../../../tests/Flags/Flags.cpp')
      D:\a\Vulkan-Hpp\Vulkan-Hpp\vulkan\vulkan_enums.hpp(45,56):
      default arguments or varargs are not allowed
```

</details>

sharadhr:
>All in all, it is a compiler issue, rather than our fault (I believe so at least). For the test I can either:
> 
> 1. Manually include `<compare>` before importing `vulkan_hpp`
> 2. Use `VULKAN_HPP_NO_SPACESHIP_OPERATOR` for the `flags.cpp` test
> 3. Disable the std module testing for the flags test
> 
> It will not solve the issue for users, but I'm not quite sure how else to do this. Any other ideas?
> 
> MSVC error log

It _is_ a compiler issue; this was declared 'not a bug' in [this response](https://developercommunity.visualstudio.com/t/Template-exports-requiring-importing-of-/1425979#T-N1435887) by an MSVC compiler developer. So users _will_ have to `#include <compare>` or `import std;` if they want to use `import vulkan_hpp`, unless they explicitly define `VULKAN_HPP_NO_SPACESHIP_OPERATOR`. 

This is what I tried to fix in #1815 and #1932.

M2-TE:
@sharadhr wow thanks for the source!

That is kind of an odd decision by them.. but oh well. I will go with the first option of just including `<compare>`.
Exporting `std::partial_ordering` by ourselves in `vulkan.cppm` seems like a bad bandaid solution and using `export import <compare>` also does not work on all compilers yet.

asuessenbach:
Great to see, that you already has sorted it out.

This is exactly what I had proposed how to work around such a "feature".

PR #2250, opened 2025-08-10
Fix for uninitialised deprecated members in DeviceCreateInfo constructor

creijon:
A small fix to initialise the `ppEnabledLayerNames` and `enabledLayerCount` members of `vk::DeviceCreateInfo` from the passed `std::vector`.  These parameters are deprecated but they should still be initialised cleanly.

This reverts some of the changes in d16c62670b02dd3292aa8711702efebfc634645f

Tested on:
Dell Precision 5690 (Meteor Lake-P)
Ubuntu 24.04
Vulkan 1.4.313

### DIFF
--- a/VulkanHppGenerator.cpp
+++ b/VulkanHppGenerator.cpp
@@ -11595,7 +11595,7 @@ ${byString}
 
     std::vector<std::string> arguments, initializers;
     bool                     arrayListed = false;
-    std::string              ignores, templateHeader, sizeChecks, copyOps;
+    std::string              templateHeader, sizeChecks, copyOps;
     for ( auto mit = structData.second.members.begin(); mit != structData.second.members.end(); ++mit )
     {
       // gather the initializers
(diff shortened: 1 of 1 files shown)


M2-TE:
That header is generated, so you have to edit the generator code in `VulkanHppGenerator.cpp`. Take a look at the `generateStructConstructorsEnhanced` function, that's where the relevant ctor is generated.

Once you have made changes there, run the generator (`VULKAN_HPP_RUN_GENERATOR=ON` in CMake).

creijon:
Thanks, I hadn't noticed the code generation.  I'll sort it out.

creijon:
I've looked through the logic in `VulkanHppGenerator.cpp` and I think that the recent changes caused this regression in the codegen.  I ran the build with `VULKAN_HPP_RUN_GENERATOR=ON` and the `vulkan_structs.hpp` file now is generated correctly.  But I'm new to this codebase so I'd kindly ask @asuessenbach to review it.

asuessenbach:
Yes, looks good so far.
Note, though, that the local variable `ignores` isn't used anymore now. So you could remove that, and its usage in the `constructorTemplate` below as well.

And... thanks a lot for your contribution!

creijon:
Good point with the unused `ignores`.  I've removed it from that part of the codegen entirely.

PR #2251, opened 2025-08-10
Fix SwapchainCreateInfoKHR where some drivers report maxImageCount as 0

creijon:
On the Meteor Lake-P GPU the driver reports maxImageCount as 0, which causes the `vk::su::clamp()` function to pass 0 into the `SwapchainCreateInfoKHR` constructor.  This change catches this and forces it to be greater than or equal to the minimum reported.

Tested on:
Dell Precision 5690 (Meteor Lake-P)
Ubuntu 24.04
Vulkan 1.4.313

### DIFF
--- a/RAII_Samples/utils/utils.hpp
+++ b/RAII_Samples/utils/utils.hpp
@@ -364,7 +364,7 @@ namespace vk
           vk::PresentModeKHR         presentMode = vk::su::pickPresentMode( physicalDevice.getSurfacePresentModesKHR( surface ) );
           vk::SwapchainCreateInfoKHR swapChainCreateInfo( {},
                                                           surface,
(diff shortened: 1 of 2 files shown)


M2-TE:
Good catch!
There are quite a few modern GPUs that report `maxImageCount` as 0, so this will be helpful.

Though, there is already a function for image count clamping called `vk::su::clampSurfaceImageCount`, so you can just replace the `vk::su::clamp` with it instead of manually editing `surfaceCapabilities.maxImageCount`. The function lives in the non-RAII samples `utils.hpp` if you want to check it.

creijon:
Cheers, I'll make that change and rebase.

creijon:
I also added a comment in `vk::su::clampSurfaceImageCount` to explain why that logic is there.

asuessenbach:
Great catch!

Thanks a lot.

PR #2252, opened 2025-08-10
Add VULKAN_HPP_NO_STRUCT_DEFAULT_INIT for designated initializer list warnings

tailhwip:
Currently vulkan-structs.hpp default initializes all struct members as in:
```
  public:
    StructureType sType              = StructureType::eApplicationInfo;
    const void *  pNext              = {};
    const char *  pApplicationName   = {};
    uint32_t      applicationVersion = {};
    const char *  pEngineName        = {};
    uint32_t      engineVersion      = {};
    uint32_t      apiVersion         = {};
```
this prevents flags like `-Wmissing-field-initializers` from properly generating warnings as technically everything is initialized, even if not by me explicitly.

This change will render struct members as follows instead:
```
  public:
#if !defined( VULKAN_HPP_NO_STRUCT_DEFAULT_INIT )
    StructureType sType              = StructureType::eApplicationInfo;
    const void *  pNext              = {};
    const char *  pApplicationName   = {};
    uint32_t      applicationVersion = {};
    const char *  pEngineName        = {};
    uint32_t      engineVersion      = {};
    uint32_t      apiVersion         = {};
#else
    StructureType sType = StructureType::eApplicationInfo;
    const void *  pNext;
    const char *  pApplicationName;
    uint32_t      applicationVersion;
    const char *  pEngineName;
    uint32_t      engineVersion;
    uint32_t      apiVersion;
#endif /*VULKAN_HPP_NO_STRUCT_DEFAULT_INIT*/
```
enabling for the absence of default initialization of struct members by setting `VULKAN_HPP_NO_STRUCT_DEFAULT_INIT`.

Happy to discuss any changes or to close the PR if this change is out of scope for the project.

### DIFF
--- a/tests/DesignatedInitializersNoDefault/DesignatedInitializersNoDefault.cpp
+++ b/tests/DesignatedInitializersNoDefault/DesignatedInitializersNoDefault.cpp
@@ -30,18 +30,22 @@ private:
 };
 
 #if defined( __clang__ ) || defined( __GNUC__ )
+#  pragma GCC diagnostic ignored "-Wunused-variable"
+
 #  pragma GCC diagnostic push
 #  pragma GCC diagnostic error "-Wmissing-field-initializers"
 #endif
 
 MyVulkanTest::MyVulkanTest()
-  : applicationInfo{ .pApplicationName   = "My Application",
+  : applicationInfo{ .pNext              = nullptr,
+                     .pApplicationName   = "My Application",
                      .applicationVersion = VK_MAKE_VERSION( 0, 0, 1 ),
                      .pEngineName        = "My Engine",
                      .engineVersion      = VK_MAKE_VERSION( 0, 0, 1 ),
                      .apiVersion         = VK_API_VERSION_1_0 }
 {
-  this->applicationInfo = vk::ApplicationInfo{ .pApplicationName   = "My Application",
+  this->applicationInfo = vk::ApplicationInfo{ .pNext              = nullptr,
+                                               .pApplicationName   = "My Application",
                                                .applicationVersion = VK_MAKE_VERSION( 0, 0, 1 ),
                                                .pEngineName        = "My Engine",
                                                .engineVersion      = VK_MAKE_VERSION( 0, 0, 1 ),
@@ -64,13 +68,15 @@ int main( int /*argc*/, char ** /*argv*/ )
 
 #if ( 20 <= VULKAN_HPP_CPP_VERSION )
   // no default initialization: all members except sType have to be specified; the order has to be respected
-  vk::ApplicationInfo ai2 = { .pApplicationName   = appName,
+  vk::ApplicationInfo ai2 = { .pNext              = nullptr,
+                              .pApplicationName   = appName,
                               .applicationVersion = appVersion,
                               .pEngineName        = engineName,
                               .engineVersion      = engineVersion,
                               .apiVersion         = apiVersion };
 
-  vk::ApplicationInfo ai3{ .pApplicationName = appName,
+  vk::ApplicationInfo ai3{ .pNext              = nullptr,
+                           .pApplicationName   = appName,
                            .applicationVersion = appVersion,
                            .pEngineName        = engineName,
                            .engineVersion      = engineVersion,
--- a/tests/DesignatedInitializersNoDefault/DesignatedInitializersNoDefault.cpp
+++ b/tests/DesignatedInitializersNoDefault/DesignatedInitializersNoDefault.cpp
@@ -31,9 +31,6 @@ private:
 
 #if defined( __clang__ ) || defined( __GNUC__ )
 #  pragma GCC diagnostic ignored "-Wunused-variable"
-
-#  pragma GCC diagnostic push
-#  pragma GCC diagnostic error "-Wmissing-field-initializers"
 #endif
 
 MyVulkanTest::MyVulkanTest()
@@ -85,7 +82,3 @@ int main( int /*argc*/, char ** /*argv*/ )
 
   return 0;
 }
-
-#if defined( __clang__ ) || defined( __GNUC__ )
-#  pragma GCC diagnostic pop
-#endif
--- a/tests/DesignatedInitializersNoDefault/DesignatedInitializersNoDefault.cpp
+++ b/tests/DesignatedInitializersNoDefault/DesignatedInitializersNoDefault.cpp
@@ -16,13 +16,9 @@
 //                   Compile test on using designated initializers
 
 #define VULKAN_HPP_NO_STRUCT_CONSTRUCTORS
+#define VULKAN_HPP_NO_STRUCT_DEFAULT_INIT
 
-#include <iostream>
 #include <vulkan/vulkan.hpp>
-
-#if defined( __clang__ ) || defined( __GNUC__ )
-#  pragma GCC diagnostic ignored "-Wunused-variable"
-#endif
 
 class MyVulkanTest
 {
@@ -32,6 +28,11 @@ public:
 private:
   vk::ApplicationInfo applicationInfo;
 };
+
+#if defined( __clang__ ) || defined( __GNUC__ )
+#  pragma GCC diagnostic push
+#  pragma GCC diagnostic error "-Wmissing-field-initializers"
+#endif
 
 MyVulkanTest::MyVulkanTest()
   : applicationInfo{ .pApplicationName   = "My Application",
@@ -58,38 +59,27 @@ int main( int /*argc*/, char ** /*argv*/ )
   // default initialization is available in any case
   vk::ApplicationInfo ai0;
 
-#if !defined( VULKAN_HPP_NO_STRUCT_CONSTRUCTORS )
-  // due to default initializers, any number of members can be initialized
-  // most IDEs will probably help with the order of the members !
-  vk::ApplicationInfo ai1( appName );
-  vk::ApplicationInfo ai2( appName, appVersion );
-  vk::ApplicationInfo ai3( appName, appVersion, engineName );
-  vk::ApplicationInfo ai4( appName, appVersion, engineName, engineVersion );
-  vk::ApplicationInfo ai5( appName, appVersion, engineName, engineVersion, VK_API_VERSION_1_2 );
-
-  // a structure in namespace vk:: can be copied from the corresponding vulkan C-struct
-  VkApplicationInfo   vai;
-  vk::ApplicationInfo ai6( vai );
-
-#else
-
-  // aggregate initialization: need to explicitly specify sType and pNext, as well as all the other members !
+  // aggregate initialization: need to explicitly specify sType and pNext, as well as all the other members
   vk::ApplicationInfo ai1{ vk::StructureType::eApplicationInfo, nullptr, appName, appVersion, engineName, engineVersion, VK_API_VERSION_1_2 };
 
-#  if ( 20 <= VULKAN_HPP_CPP_VERSION )
-  // designated initializers are available with C++20
+#if ( 20 <= VULKAN_HPP_CPP_VERSION )
+  // no default initialization: all members except sType have to be specified; the order has to be respected
+  vk::ApplicationInfo ai2 = { .pApplicationName = appName,
+                              .applicationVersion = appVersion,
+                              .pEngineName = engineName,
+                              .engineVersion = engineVersion,
+                              .apiVersion = apiVersion };
 
-  // it's allowed, but not recommended to explicitly specify sType
-  vk::ApplicationInfo ai2 = { .sType = vk::StructureType::eApplicationInfo };
-
-  // any number of the members can be specified; the order has to be respected
-  vk::ApplicationInfo ai3 = { .pApplicationName = appName };
-  vk::ApplicationInfo ai4 = { .applicationVersion = 1, .engineVersion = 2 };
-
-  vk::ApplicationInfo ai5{ .pEngineName = engineName };
-  vk::ApplicationInfo ai6{ .pApplicationName = appName, .apiVersion = VK_API_VERSION_1_2 };
-#  endif
-#endif
+  vk::ApplicationInfo ai3{ .pApplicationName = appName,
+                           .applicationVersion = appVersion,
+                           .pEngineName = engineName,
+                           .engineVersion = engineVersion,
+                           .apiVersion = apiVersion };
+#endif /*VULKAN_HPP_CPP_VERSION*/
 
   return 0;
 }
+
+#if defined( __clang__ ) || defined( __GNUC__ )
+#  pragma GCC diagnostic pop
+#endif
--- a/tests/DesignatedInitializersNoDefault/DesignatedInitializersNoDefault.cpp
+++ b/tests/DesignatedInitializersNoDefault/DesignatedInitializersNoDefault.cpp
@@ -12,8 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 //
-// VulkanHpp Tests : DesignatedInitializers
-//                   Compile test on using designated initializers
+// VulkanHpp Tests : DesignatedInitializersNoDefault
+//                   Compile test on init lists without default values
 
 #define VULKAN_HPP_NO_STRUCT_CONSTRUCTORS
 #define VULKAN_HPP_NO_STRUCT_DEFAULT_INIT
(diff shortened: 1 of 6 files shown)


M2-TE:
I feel like the point of designated initializers is to NOT have to set everything, only the selected members. I am unsure of how useful this would actually be to people, but I could be wrong there.

Either way, could you please run the generator, so that we see the header changes? Just configure CMake with `VULKAN_HPP_RUN_GENERATOR=ON` and build the project. With `VULKAN_HPP_TESTS_BUILD=ON` you can also check your test with the new headers.

tests/DesignatedInitializersNoDefault/DesignatedInitializersNoDefault.cpp
```diff
@@ -30,18 +30,22 @@ class MyVulkanTest
 };
 
 #if defined( __clang__ ) || defined( __GNUC__ )
+#  pragma GCC diagnostic ignored "-Wunused-variable"
+
 #  pragma GCC diagnostic push
 #  pragma GCC diagnostic error "-Wmissing-field-initializers"
```
M2-TE: is `-Wmissing-field-initializers` not included in `-Wall`, `-Wextra` or `-Wpedantic`?

tailhwip: You are correct, while `-Wmissing-field-initializers` actually is not, `-Wmissing-designated-field-initializers` appears to be which is sufficient in this case. I've removed the pragma.

tailhwip:
> I feel like the point of designated initializers is to NOT have to set everything, only the selected members.

That's fair. I do think they can serve both purposes; to not have to set everything when you are comfortable with an API and know what to look for, but also to provide extra clarity by naming every parameter when that is not yet the case.

The expanded change has ballooned a little bit, a diff is available at https://diffy.org/diff/f6a6bfb49f60e

If 500kb extra on the structs header is undesirable perhaps, rather than copying the initializer list, each initializer could be wrapped in a macro of sorts that expands into assigning its default value or nothing, but I'm not sure if this would result in other slowdowns on this scale. Would love to hear your thoughts;
```
#if !defined ( VULKAN_HPP_NO_STRUCT_DEFAULT_INIT )
#  define VULKAN_HPP_STRUCT_DEFAULT_INIT(value) = value
#else
#  define VULKAN_HPP_STRUCT_DEFAULT_INIT(value)
#endif /*VULKAN_HPP_NO_STRUCT_DEFAULT_INIT*/
  
public:
    StructureType sType = StructureType::eApplicationInfo;
    const void * pNext VULKAN_HPP_STRUCT_DEFAULT_INIT({});
    const char * pApplicationName VULKAN_HPP_STRUCT_DEFAULT_INIT({});
```

M2-TE:
> but also to provide extra clarity by naming every parameter when that is not yet the case

That is a pretty good point.

Extending an already huge file even more does not seem like the best approach though, so I like the initialization with a macro. It adds a little bit of obfuscation to the constructor, but shouldn't noticably impact compile times.

asuessenbach:
Well, I'm not really sure about the value of something like `VULKAN_HPP_NO_STRUCT_DEFAULT_INIT`.
I can't imagine, that anybody would really want to code even a tiny little toy project with such a setting.

That is, if at all, we should take the macro approach like described above.

M2-TE:
I suppose Vulkan learners should instead just opt to use the full constructors? They have all the parameters that are available for the struct, even if some of them are optional.

Other than that, learners likely need to look at the spec for structs either way, those will give you all the possible members to set, alongside explanations.

tailhwip:
It's not so much about writing the code, where I agree having the spec at hand is a better idea, but more so about being able to read it back confidently later and knowing you're not missing anything that is value initialized internally.

Right now they of course will all seem so obvious, but I am going to forget some of them at some point:
```
// won't show anything as levelCount is value initialized to 0
// the validation layers will catch it but compile time validation trumps runtime validation
vk::ImageSubresourceRange range {
    .aspectMask     = vk::ImageAspectFlagBits::eColor,
    .baseMipLevel   = 0,
    .baseArrayLayer = 0,
    .layerCount     = 1
};

// won't copy anything as size is value initialized to 0
vk::BufferCopy copyRegion { .srcOffset = 0, .dstOffset = 0 };

// shadow maps will be weird
// difficult to spot that depthBiasEnable has been left out which is value initialized to 0
VkPipelineRasterizationStateCreateInfo rasterizer {
    .sType                   = VK_STRUCTURE_TYPE_PIPELINE_RASTERIZATION_STATE_CREATE_INFO;
    .pNext                   = nullptr;
    .flags                   = 0;
    .depthClampEnable        = VK_FALSE;
    .rasterizerDiscardEnable = VK_FALSE;
    .polygonMode             = VK_POLYGON_MODE_FILL;
    .cullMode                = VK_CULL_MODE_BACK_BIT;
    .frontFace               = VK_FRONT_FACE_COUNTER_CLOCKWISE;
    .depthBiasConstantFactor = 1.25f;
    .depthBiasClamp          = 0.0f;
    .depthBiasSlopeFactor    = 1.75f;
    .lineWidth               = 1.0f;
}
```



I'd like to try approach this PR from a different angle. Per the C++20 designated initializer list spec ommitted members are already value initialized. It seems unnecessary to force this value initialization on consumers of the library if the following 2 scenarios have the same result:

Current:
```
struct BufferCreateInfo
{
...

public:
    VULKAN_HPP_NAMESPACE::StructureType     sType                 = StructureType::eBufferCreateInfo;
    const void *                            pNext                 = {};
    VULKAN_HPP_NAMESPACE::BufferCreateFlags flags                 = {};
    VULKAN_HPP_NAMESPACE::DeviceSize        size                  = {};
    VULKAN_HPP_NAMESPACE::BufferUsageFlags  usage                 = {};
    VULKAN_HPP_NAMESPACE::SharingMode       sharingMode           = VULKAN_HPP_NAMESPACE::SharingMode::eExclusive;
    uint32_t                                queueFamilyIndexCount = {};
    const uint32_t *                        pQueueFamilyIndices   = {};
};

BufferCreateInfo buff_info { .size = 512 };
// everything else is value initialized to 0 due to member initializers
```

Proposed:
```
struct BufferCreateInfo
{
...

public:
    VULKAN_HPP_NAMESPACE::StructureType     sType = StructureType::eBufferCreateInfo;
    const void *                            pNext;
    VULKAN_HPP_NAMESPACE::BufferCreateFlags flags;
    VULKAN_HPP_NAMESPACE::DeviceSize        size;
    VULKAN_HPP_NAMESPACE::BufferUsageFlags  usage;
    VULKAN_HPP_NAMESPACE::SharingMode       sharingMode;
    uint32_t                                queueFamilyIndexCount;
    const uint32_t *                        pQueueFamilyIndices;
};

BufferCreateInfo buff_info { .size = 512 };
// everything else is value initialized to 0 as per the C++20 designated initializer list spec
```

I need to look into the hpp generator a little more, but so far it looks like everything is only ever 0 initialized and if not at least 99% is. In that case we might as well leave that up to the compiler to value initialize and let -Wall do its thing. This shouldn't turn into a breaking change overall as -Wmissing-designated-initializer-lists is not a warning that is turned on by default (but is part of -Wall), and if your project uses -Wall (which Vulkan-Hpp also does) it would seem more explicit to turn that warning off via -Wno-missing-designated-initializer-lists if you don't want it rather than make it unusable all together.

M2-TE:
Most, especially larger, projects seem to use -Wall and I feel like Vulkan-Hpp should just be able to compile with all warnings enabled, without requiring to disable specific ones. It may just be one flag, but it would still be annoying to have to set this for every vulkan project that uses designated initializers.

tailhwip:
I don't mean for this to devolve into a discussion about which approach is better, so if we are in dissagreement we should keep things as they are (or at least I can update the PR to the other macro so we can take a look at what that'd look like). But if you have the time let me play devil's advocate, the following is all in good faith;

Whether Vulkan-Hpp compiles in a project with `-Wall` depends on how that consuming project has used Vulkan-Hpp. If they set `-Wall` but ommit members from designated init lists (nothing wrong with that) of course it won't compile. It shouldn't; they are in offence of `-Wall`. They should explicitly set `-Wall -Wno-missing-designated-initializer-lists` as that reflects what they want to do.

If a project wants to depend on `-Wall` it has to embrace all that is `-Wall`. If `-Wall` comes with warnings that are undesirable to the project those warnings should be turned off, rather than designing the API to dance around those warnings and implicitly silence them. Vulkan-Hpp currently prevents consumers of its API from having `-Wall` report on all the warnings it is designed to report on, at least for all of the Vulkan-Hpp code in their project.

**Without** explicit value initialization of struct members in the Vulkan-Hpp API this project and its consumers can enable and disable compiler warnings as they see fit.

**With** explicit value initialization of struct members in the Vulkan-Hpp API this project only pretents to use `-Wall` in its entirety and consuming projects have no choice but to agree that `-Wmissing-designated-init-lists` does not belong in `-Wall`, at least for all of the Vulkan-Hpp code in their project.

M2-TE:
> If they set -Wall but ommit members from designated init lists (nothing wrong with that) of course it won't compile. It shouldn't; they are in offence of -Wall

I absolutely agree in retrospect

> Without explicit value initialization of struct members in the Vulkan-Hpp API this project and its consumers can enable and disable compiler warnings as they see fit.

Compared to the current alternative, this does seem like the better option, so what do you think about a middle ground:

We remove the default inits (the ones with =0 or {}) and perform compiler-specific suppression of the warning, which can be toggled off with some flag like `VULKAN_HPP_WARN_INCOMPLETE_INIT`. The warning suppression should be implemented at the beginning of the structs file kind of like this:

```cpp
#ifndef VULKAN_HPP_WARN_INCOMPLETE_INIT
#  if defined( __clang__ ) || defined( __GNUC__ )
#    pragma GCC diagnostic push
#    pragma GCC diagnostic ignored "-Wmissing-field-initializers"
#  endif
#endif

[...] // structs

#ifndef VULKAN_HPP_WARN_INCOMPLETE_INIT
#  if defined( __clang__ ) || defined( __GNUC__ )
#    pragma GCC diagnostic pop
#  endif
#endif
```

This is already being done to some extent in various locations (e.g. with the deprecation flags), so it would be quite suitable here.

tailhwip:
Maybe the pragma is not even necessary.

Scenario 1: Consuming projects that do not set `-Wall` will be unaffected regardless.

Scenario 2: Consuming projects that already set both `-Wall` and `-Wno-missing-designated-initializer-lists` will also be unaffected.

Scenario 3: Consuming projects that only set `-Wall` most likely already enforce the warning on non-Vulkan-Hpp related code too. They should(?) appreciate the change as it brings everything in line with each other. But it is not going to please everyone.

If scenario 3 is too big a change to enforce on existing projects I'm happy with the proposed pragma. That said the generated warnings would come from their own code, not from the Vulkan-Hpp headers, so they can just as quickly fix it by setting `-Wno-missing-designated-initializer-lists` if they don't want to deal with it.

There is also still the following approach to enjoy value initialization with `-Wall` set, but this could be a big refactor job (that can at least be defferred by setting `-Wno-missing-designated-initializer-lists`):
```
VkPipelineRasterizationStateCreateInfo rasterizer {};
rasterizer.polygonMode = VK_POLYGON_MODE_FILL;
rasterizer.cullMode    = VK_CULL_MODE_BACK_BIT;
rasterizer.frontFace   = VK_FRONT_FACE_COUNTER_CLOCKWISE;
```

If any Vulkan-Hpp header uses designated init lists with ommitted members we might be able to address that in the generator, but even if they exist their usage appears to be very sparse. I will take a closer look tomorrow.

tailhwip:
On second thought we might not have a choice but to force unhappy users from scenario 3 to fix it themselves by setting `-Wno-missing-designated-initializer-lists` or adding their own pragma guards.

If I understand it correctly the proposed pragma guard in the structs header would only apply to the struct code inside the header, not to code using those structs.

M2-TE:
Damn, that's unfortunate..

tailhwip:
What do you think of updating the PR to the macro discussed earlier to optionally remove value initialization without greatly increasing the size of the structs header and to make sure nobody has any immediate breaking changes:
```
#if !defined ( VULKAN_HPP_NO_STRUCT_FORCED_VALUE_INIT )
#  define VULKAN_HPP_STRUCT_VALUE_INIT(value) = value
#else
#  define VULKAN_HPP_STRUCT_VALUE_INIT(value)
#endif /*VULKAN_HPP_NO_STRUCT_FORCED_VALUE_INIT*/
  
public:
    StructureType sType = StructureType::eApplicationInfo;
    const void * pNext VULKAN_HPP_STRUCT_VALUE_INIT({});
    const char * pApplicationName VULKAN_HPP_STRUCT_VALUE_INIT({});
...
```

Then when 1.5 rolls around it can be considered to be removed completely.

asuessenbach:
Well, all those compiler specific warning settings and stuff aside: One of the advantages of Vulkan-Hpp over vulkan.h is, that you can't get uninitialized structs, even when you just instantiate them. And you just have to set the members you actually care about.

That is, there might be no difference for something like
`BufferCreateInfo buff_info { .size = 512 };`
No matter if the other struct members are explicitly initialized or not, they are just 0.

But there's a big difference to just
`BufferCreateInfo buff_info;`
Here, all members, that don't have default initializers on their own (that is: all standard types like uint32_t, or float) are uninitialized.

That said, I don't think, it's really worth to have some choice via a define like `VULKAN_HPP_NO_STRUCT_DEFAULT_INIT`, or so.

tailhwip:
Isn't that exactly what you would expect to happen for `BufferCreateInfo buff_info;` as an aggregate/POD with `VULKAN_HPP_NO_STRUCT_CONSTRUCTORS` set?

If `VULKAN_HPP_NO_STRUCT_CONSTRUCTORS` is not set `BufferCreateInfo buff_info;` will default initialize just fine by value initializing all members as per its user defined constructor.

Looking at vk::BufferCreateInfo:
```
...
    VULKAN_HPP_CONSTEXPR BufferCreateInfo( BufferCreateFlags flags_                 = {},
                                           DeviceSize        size_                  = {},
                                           BufferUsageFlags  usage_                 = {},
                                           SharingMode       sharingMode_           = SharingMode::eExclusive,
                                           uint32_t          queueFamilyIndexCount_ = {},
                                           const uint32_t *  pQueueFamilyIndices_   = {},
                                           const void *      pNext_                 = nullptr ) VULKAN_HPP_NOEXCEPT
      : pNext{ pNext_ }
      , flags{ flags_ }
      , size{ size_ }
      , usage{ usage_ }
      , sharingMode{ sharingMode_ }
      , queueFamilyIndexCount{ queueFamilyIndexCount_ }
      , pQueueFamilyIndices{ pQueueFamilyIndices_ }
    {
    }

...

public:
    StructureType     sType                 = StructureType::eBufferCreateInfo;
    const void *      pNext                 = {};
    BufferCreateFlags flags                 = {};
    DeviceSize        size                  = {};
    BufferUsageFlags  usage                 = {};
    SharingMode       sharingMode           = SharingMode::eExclusive;
    uint32_t          queueFamilyIndexCount = {};
    const uint32_t *  pQueueFamilyIndices   = {};

...
```
It is the constructor with default parameters that ensures the struct is default initialized, not the value initializations on the members. If `VULKAN_HPP_NO_STRUCT_CONSTRUCTORS` is set I expect to use the structs as PODs and I would expect them to behave as PODs (i.e. remain uninitialized if created with `BufferCreateInfo buff_info;`).

In other words: if we remove the value initialization from the struct members `BufferCreateInfo buff_info;` will still value initialize all of its members due to the default constructor that is available. Only when `VULKAN_HPP_NO_STRUCT_CONSTRUCTORS` is set and the struct is treated as an aggregate are you left with uninitialized fields, but that is something you explicitly asked for by setting `VULKAN_HPP_NO_STRUCT_CONSTRUCTORS`.

If someone wants to set `VULKAN_HPP_NO_STRUCT_CONSTRUCTORS` and value initialize a struct rather than default initialize they should use `BufferCreateInfo buff_info{};` not `BufferCreateInfo buff_info;`.

asuessenbach:
> If VULKAN_HPP_NO_STRUCT_CONSTRUCTORS is set I expect to use the structs as PODs and I would expect them to behave as PODs (i.e. remain uninitialized if created with BufferCreateInfo buff_info;).

And that's a wrong assumption. Having no (non-default-)constructors doesn't mean, entities are not default-initialized. The wrappers around the C-structs are not PODs. They are initialized, and you just have to care about the elements you're interested in.

tailhwip:
~~As the wrappers are both trivially constructable and have a standard layout without the constructors in place they are by all accounts of the specification PODs.~~ And of course without their constructor they will still be default initialized. But default initialization of non-static fundamental types in C++ is defined as being indeterminate.

In general:
```
struct Point { float x; float y; };

Point p;   // default initialization, values are indeterminate and access prior to explicit assignment is undefined bevaviour
Point p{}; // value initialization, members are initialized to 0
```

In Vulkan-Hpp:
```
BufferCreateInfo buff_info;   // value initialization, members are initialized to 0
BufferCreateInfo buff_info{}; // value initialization, members are initialized to 0
```

The language already has a built-in construct to request value initialization. I'm only asking to bring back the option to default initialize by removing the value initialization from the struct members. Note that this would only affect usage of the API with VULKAN_HPP_NO_STRUCT_CONSTRUCTORS set:

Vulkan-Hpp with constructors and no member-wise value initialization
```
BufferCreateInfo buff_info;   // value initialization, members are initialized to 0 (via constructor)
BufferCreateInfo buff_info{}; // value initialization, members are initialized to 0
```

Vulkan-Hpp without constructors and no member-wise value initialization
```
#define VULKAN_HPP_NO_STRUCT_CONSTRUCTORS

BufferCreateInfo buff_info;   // default initialization, values are indeterminate and access prior to explicit assignment is undefined bevaviour
BufferCreateInfo buff_info{}; // value initialization, members are initialized to 0
```

Also removing the value initialization from the struct members does not affect incomplete designated initializer lists, as missing properties from designated initializer lists are well-defined to be zero initialized. With or without explicit value initialization on the members in the struct definition.

Before change:
```
Extend2D e { .width = 0 }; // height is zero initialized due to explicit value initialization on struct member
```

After change:
```
Extend2D e { .width = 0 }; // height is zero initialized by the compiler
```

tailhwip:
In hindsight you are completely correct, the in-class member initializers make them non-aggregates.

It looks like our opinions on default and value initializiation differ. I'd be happy to respond to any further comments, but if you feel it is out of scope please feel free to close the PR.

asuessenbach:
Thanks a lot for this discussion.
Even though, our opinions on this specific issue are different, I really enjoyed this thread.
Please don't hesitate to initiate new issues. It is discussions like these that potentially advance Vulkan-Hpp.

PR #2254, opened 2025-08-11
CI: Reduce build times by reusing generated headers

M2-TE:
Current CI takes a very long time to complete, owing in part to the fact that the generator gets rebuilt and run with every spawned instance. Since it will always compile with C++20 and should only really be compiled in Release, it can be reused without issue within the same compiler instance. Each compiler instance just loops over C++ versions from 11 to 23 in addition to Debug and Release.

Compilers could also get their own for loop, but leaving them separated like this makes finding issues with specific compilers easier, as those are the most frequent culprits.

I've used bash syntax, so windows and macos might fail currently. What interests me first and foremost is how fast the ubuntu CI will be.

The run for ubuntu 24 also only uses clang++-17, since I want to check if the syntax for the module switch is valid and working. Will of course be changed back to use all compilers afterwards.

### DIFF
--- a/.github/workflows/ci-macos.yml
+++ b/.github/workflows/ci-macos.yml
@@ -69,6 +69,20 @@ jobs:
       run: |
         cmake -B build -G Ninja                        \
           -D VULKAN_HPP_GENERATOR_BUILD=ON             \
+          -D VULKAN_HPP_RUN_GENERATOR=OFF              \
+          -D VULKAN_HPP_SAMPLES_BUILD=OFF              \
+          -D VULKAN_HPP_TESTS_BUILD=OFF                \
+          -D VULKAN_HPP_ENABLE_CPP20_MODULES=OFF       \
+          -D VULKAN_HPP_DISPATCH_LOADER_DYNAMIC=ON     \
+          -D VULKAN_HPP_BUILD_WITH_LOCAL_VULKAN_HPP=ON \
+          -D VULKAN_HPP_PRECOMPILE=OFF                 \
+          -D CMAKE_CXX_COMPILER=${{matrix.compiler}}   \
+          -D CMAKE_CXX_STANDARD=20                     \
+          -D CMAKE_BUILD_TYPE=Debug
+        cmake --build build --parallel
+
+        cmake -B build -G Ninja --fresh                \
+          -D VULKAN_HPP_GENERATOR_BUILD=ON             \
           -D VULKAN_HPP_RUN_GENERATOR=ON               \
           -D VULKAN_HPP_SAMPLES_BUILD=OFF              \
           -D VULKAN_HPP_TESTS_BUILD=OFF                \
@@ -79,7 +93,7 @@ jobs:
           -D CMAKE_CXX_COMPILER=${{matrix.compiler}}   \
           -D CMAKE_CXX_STANDARD=20                     \
           -D CMAKE_BUILD_TYPE=Release
-        cmake --build build --parallel
+        cmake --build build --parallel --clean-first
 
     - name: Build samples and tests
       run: |
--- a/.github/workflows/ci-ubuntu-22.yml
+++ b/.github/workflows/ci-ubuntu-22.yml
@@ -71,6 +71,20 @@ jobs:
       run: |
         cmake -B build -G Ninja                        \
           -D VULKAN_HPP_GENERATOR_BUILD=ON             \
+          -D VULKAN_HPP_RUN_GENERATOR=OFF              \
+          -D VULKAN_HPP_SAMPLES_BUILD=OFF              \
+          -D VULKAN_HPP_TESTS_BUILD=OFF                \
+          -D VULKAN_HPP_ENABLE_CPP20_MODULES=OFF       \
+          -D VULKAN_HPP_DISPATCH_LOADER_DYNAMIC=ON     \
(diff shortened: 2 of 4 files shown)


asuessenbach:
Nice idea, indeed!
Let's see if it helps somewhat...

M2-TE:
Looks to have shaved off around 5 to 20 minutes per compiler instance? Hard to tell as the CI build times fluctuate a lot.

Two potential improvements still:

1. What is the reason behind `VULKAN_HPP_PRECOMPILE` being forced to `OFF`? I didn't benchmark anything, but build times should improve with it enabled
2. You reckon we can install glm, glfw and glslang once during the run? The latter in particular takes up a large chunk (around 1/3 for me) of the compile time. My idea is to compile them just once with Release on C++(11?) and then just link to them afterwards. It saves a lot of time, so I'm just waiting to hear your stance on this

@asuessenbach any idea why the windows build is failing? It seems to be an issue with the SharedHandle constructor

```
D:\a\Vulkan-Hpp\Vulkan-Hpp\samples\SharedHandles\SharedHandles.cpp(90): error C2440: 'initializing': cannot convert from 'initializer list' to 'vk::SharedHandle<vk::SurfaceKHR>'
D:\a\Vulkan-Hpp\Vulkan-Hpp\samples\SharedHandles\SharedHandles.cpp(90): note: 'vk::SharedHandle<vk::SurfaceKHR>::SharedHandle': function does not take 2 arguments
D:\a\Vulkan-Hpp\Vulkan-Hpp\vulkan/vulkan_shared.hpp(303): note: could be 'vk::SharedHandle<vk::SurfaceKHR>::SharedHandle(HandleType,vk::detail::ObjectDestroyShared<HandleType>) noexcept'
        with
        [
            HandleType=vk::SurfaceKHR
        ]
D:\a\Vulkan-Hpp\Vulkan-Hpp\vulkan/vulkan_shared.hpp(294): note: or       'vk::SharedHandle<vk::SurfaceKHR>::SharedHandle(HandleType,vk::SharedHandle<vk::Instance>,vk::SharedHandle<vk::NoDestructor>,const Dispatcher &) noexcept'
        with
        [
            HandleType=vk::SurfaceKHR
        ]
D:\a\Vulkan-Hpp\Vulkan-Hpp\vulkan/vulkan_shared.hpp(286): note: or       'vk::SharedHandle<vk::SurfaceKHR>::SharedHandle(HandleType,vk::SharedHandle<vk::Instance>,vk::detail::ObjectDestroyShared<HandleType>) noexcept'
        with
        [
            HandleType=vk::SurfaceKHR
        ]
D:\a\Vulkan-Hpp\Vulkan-Hpp\samples\SharedHandles\SharedHandles.cpp(90): note: while trying to match the argument list '(VkSurfaceKHR, vk::SharedInstance)'
```

Is it expecting `(VkSurfaceKHR, VkInstance)` instead of `(VkSurfaceKHR, vk::SharedInstance)`? Some lack of implicit casting maybe? Not sure why this fails here, I should not have affected it.

asuessenbach:
> Not sure why this fails here, I should not have affected it.

Nope, should not have affected.

Either way, could you please add some echo to the build step, like
```
            echo "================================================================================="
            echo "Building C++"$cpp_standard" in "$build_type" mode on "${{matrix.os}}" with architecture "${{matrix.architecture}}
            echo "================================================================================="
```
Would simplify "debugging" like this a little.

You're building everything directly into the "build"-directory. When looping over standards and build types, you might need to build into something like "build/$cpp_standard/$build_type"?

asuessenbach:
Ahm... instead of $cpp_standard, you should use $CXX_STANDARD, and instead of $build_type use $BUILD_TYPE?

M2-TE:
I've added the echo, should help.

> You're building everything directly into the "build"-directory. When looping over standards and build types, you might need to build into something like "build/$cpp_standard/$build_type"?

That was the first idea, but I remember there being storage space issues with this approach combined with loops in the past. That's why i opted to rebuild from a clean tree with `cmake [...] --fresh` and `cmake [...] --clean-first`.

I had accidentally set Ninja as the generator for Windows (which should still work either way), so it should now use MSBuild again. If that does not work, I will likely just remove the build folder for a guaranteed clean rebuild after each loop iteration.

M2-TE:
Tried installing the deps once before looping over the C++ versions. Seems good on ubuntu, but no idea if it will work with windows

I'm sorry for the spotty changes, I wish there was a better way for me to test CI before opening/changing PRs..

M2-TE:
Looks like they could all run in parallel, that's pretty nice. CI logs are also more concise, as they do not include non-sample/-test compilation logs.

I've disabled the installing on windows, as I don't really know how to do it without sudo. Does not matter much, as its plenty fast on there.

Should be ready now, unless you have more ideas?

asuessenbach:
Well, it's hard to tell if this really speeds up CI. The variance is simply too large.
But it looks really great now, thanks a lot for your work and dedication!

One little thing, maybe: I'm wondering if it's worth to compile the generators not only in release, but also in debug mode, in order spot potential issues there as well.

M2-TE:
I can add a step to also try compiling the generator in debug mode. Better safe than sorry!

Did you also mean to use headers created by debug mode generators? If so, I don't think they will be different, so that will likely not be worth doing.

asuessenbach:
I think, just the debug build would be fine. There's no need to generate the headers twice.

asuessenbach:
Thanks a lot for your contribution! Great work.

PR #2256, opened 2025-08-14
Refactor file generation functions

asuessenbach:
Plus some minor cleanup work.

### DIFF
--- a/vulkan/vulkan_enums.hpp
+++ b/vulkan/vulkan_enums.hpp
@@ -12,9 +12,6 @@
 // IWYU pragma: private; include "vulkan.hpp"
 
 #include <type_traits>  // for std::underlying_type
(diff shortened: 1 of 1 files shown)


M2-TE:
This might clash with #2248 since I changed the generator code a bit for `vulkan_enums.hpp`:
```cpp
#if defined( VULKAN_HPP_HAS_SPACESHIP_OPERATOR )
# include <compare>       // for std::partial_ordering (<=> operator)
#endif
```
It's fine if this change is left out since it will likely be unnecessary (was an attempted fix for MSVC bug), but just wanted to make you aware

asuessenbach:
Ah, ok. It can be removed as (for module usage) it has to be included outside of vulkan.hpp, right?

M2-TE:
> It can be removed as (for module usage) it has to be included outside of vulkan.hpp, right?

Correct

PR #2257, opened 2025-08-14
CI: Removed g++-12 compiler instance from ubuntu 22 CI

M2-TE:
g++-12 is tested by both Ubuntu 22 and Ubuntu 24, so does it really make sense to have it on both?


asuessenbach:
You're probably right. I would assume, that the same compiler version works the same on different OS versions (fingers crossed).

That is, it can be removed from one version.

PR #2273, opened 2025-09-02
Fix missing `CreateReturnType` export

otreblan:


asuessenbach:
Thanks for providing this fix.
But as that file, `vulkan.cppm`, is generated, we would need to add that to the `VulkanHppGenerator`.
It would belong into the function `VulkanHppGenerator::generateCppModuleRaiiUsings`.

asuessenbach:
By the way: why do you need that `CreateReturnType` at all? It has been gone with #2226.

otreblan:
My distro's repos has vulkan-headers 1.4.321 which still has `CreateReturnType`

PR #2276, opened 2025-09-05
C++Module: Manually export vk::SwapchainOwns

### LINKED ISSUE #2170: vk::SwapchainOwns SharedImage flag not exported in C++20 Module

M2-TE:
This flag was previously not exported, though I'm not entirely sure why it is not part of the `m_handles` or `requireData.types`. Are enums exluded?

Either way, this will resolve #2170

### DIFF
--- a/VulkanHppGenerator.cpp
+++ b/VulkanHppGenerator.cpp
@@ -5195,6 +5195,11 @@ std::string VulkanHppGenerator::generateCppModuleSharedHandleUsings() const
   sharedHandleUsings += R"(  using VULKAN_HPP_NAMESPACE::SharedHandleTraits;
 )";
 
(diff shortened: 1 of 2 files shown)


asuessenbach:
vk::SwapchainOwns is a manually added flag for the SharedHandle<vk::Image> to allow non-owning versions of that type. And it just have been missed to add.

Issue #1993, opened 2024-11-13
Comprehensive testing, CI, and CD for `vulkan.cppm`


sharadhr:
The `vulkan.cppm` module needs more rigorous automated testing, CI and CD; there is currently only one test that doesn't exercise both the entirety of the module and potential compiler bugs that could be exposed with more testing (such as #1943 which is an MSVC-specific problem). 

Of course, given the still-experimental status of C++ standard modules as a whole, any CI/CD failures shouldn't block merges, but regressions should.

Just for testing `vulkan.cppm`, I propose setting up Actions each with the latest possible compilers for maximum coverage:

- Windows 11
  - Latest MSVC
  - Latest LLVM/Clang for MSVC ABI
- macOS Sequoia
  - Latest Xcode AppleClang
  - Latest LLVM/Clang
- Linux 
  - Latest GCC
  - Latest LLVM/Clang
  
I propose restricting to only the latest possible compilers as C++  standard modules are so new. As new compilers are released these environments should be periodically updated to follow new compiler version updates. 
  
Next, I propose porting the existing tests in [`samples`](/KhronosGroup/Vulkan-Hpp/tree/main/samples) and [`tests`](/KhronosGroup/Vulkan-Hpp/tree/main/tests) as well as the _entirety_ of [Sascha Willems' Vulkan samples](/SaschaWillems/Vulkan) to use `vulkan.cppm`. 

This will provide very comprehensive code and toolchain coverage.

sharadhr:
@M2-TE now that we've merged in #2303 maybe we can work towards resolving and closing this, also now that most recent compilers are happy to support named modules. 

I need to refresh myself on the CI/CD setup, I haven't looked at it in the greatest detail in a long while.

M2-TE:
First step will be to get modules to compile, which should be doable with MSVC and Clang. I will work on that.

I've figured out how to get all tests working with modules beforehand, some of which was done in #2292. Others require some more work.

All that's missing is compiling all samples with modules. Could you look into how to best approach those?

sharadhr:
I think the broad part of this work is done; the next step will be to migrate both the normal samples and the RAII samples to the module. 

It is here that I really want to demonstrate the performance of module compilation versus pure headers versus PCHs. In fact I would like to make the `utils` library a module as well.

PR #2277, opened 2025-09-05
Tests: Enabled ArrayProxy test for C++Module, except for gcc <= 15

M2-TE:
Related to #1993

gcc 15 has an issue where it doesn't seem to respect `pragma GCC diagnostic ignored "-Winit-list-lifetime"` when importing vulkan_hpp as a module, so I've disabled the test for it. Otherwise, passing an initializer-list to construct an ArrayProxy would emit this warning.

### DIFF
--- a/tests/ArrayProxy/CMakeLists.txt
+++ b/tests/ArrayProxy/CMakeLists.txt
@@ -14,7 +14,7 @@
 
 vulkan_hpp__setup_test( NAME ArrayProxy )
 if( VULKAN_HPP_ENABLE_CPP20_MODULES )
-    # gcc 15 fails to ignore -Winit-list-lifetime when importing vulkan_hpp
+    # gcc versions <= 15.x fail to ignore -Winit-list-lifetime when importing vulkan_hpp
     if ( NOT (CMAKE_CXX_COMPILER_ID STREQUAL "GNU" AND CMAKE_CXX_COMPILER_VERSION VERSION_LESS "16") )
         vulkan_hpp__setup_test( NAME ArrayProxy CXX_MODULE )
     endif()


tests/ArrayProxy/CMakeLists.txt
```diff
@@ -14,5 +14,8 @@
 
 vulkan_hpp__setup_test( NAME ArrayProxy )
 if( VULKAN_HPP_ENABLE_CPP20_MODULES )
-    # vulkan_hpp__setup_test( NAME ArrayProxy CXX_MODULE ) # error with GCC
+    # gcc 15 fails to ignore -Winit-list-lifetime when importing vulkan_hpp
```
asuessenbach: Unfortunately, gcc 15 isn't used in any of our CI checks. I assume, you're using it locally?
But if I see it right, you're skipping that test not only for gcc 15, but for all upcoming gcc versions above 15 as well. Is that intended?

M2-TE: > Unfortunately, gcc 15 isn't used in any of our CI checks. I assume, you're using it locally?

Yup, I try to test newest gcc, as it's the only one of gcc that works well with modules right now. Sadly could not find a way to use it in the `EXT` runner though.

> but for all upcoming gcc versions above 15 as well

It was intended to preemptively prevent issues, but I think you are right. It does not make much sense and it's better to use `VERSION_LESS 16`. I will change it!

asuessenbach: I still don't get it... now you exclude all gcc versions below 16, but I thought, it's just version 15 that has a problem?

M2-TE: The versions below 15 have the issue as well, alongside other issues that do not allow them to properly run these tests in the first place. They cannot even properly use `import vulkan_hpp` without some hacks.

That's why I currently use gcc 15 as the baseline for whether the tests work for gcc as a whole.

asuessenbach: ok... then you should adjust the comment accordingly.
Or simply exclude gcc here, without version test at all.

M2-TE: Done. Sorry for the confusion!

asuessenbach:
Yeah, thanks for your contribution, again.

Issue #2272, opened 2025-09-02
What is a use case for `vk::raii::Image`?


sharadhr:
`vk::raii::Image` is an _owning_ type corresponding to `VkImage`, but I'm struggling to understand how it could be used without running into bad destruction ordering. That is, consider the following:

```cpp

struct Engine
{
vk::raii::SwapchainKHR swapchain{makeSwapchain()};
std::vector<vk::raii::Image> swapchainImages{getSwapchainImages()}; // this function creates a `std::vector<vk::raii::Image>` from each of the `vk::Image` returned by `swapchain.getImages()` 
}; 
```

Now, the destruction order is as follows:

1. The `swapchainImages` destructor is called. This invokes the destructor of each contained `vk::raii::Image`, and thus each `VkImage` is destroyed by calling [`vkDestroyImage`](https://registry.khronos.org/vulkan/specs/latest/man/html/vkDestroyImage.html).
2. The `swapchain` destructor is invoked. This calls [`vkDestroySwapchainKHR`](https://registry.khronos.org/vulkan/specs/latest/man/html/vkDestroySwapchainKHR.html), and the manual states:
   > `swapchain` and all associated `VkImage` handles are destroyed, and must not be acquired or used any more by the application.
3. Boom, double-free here, as the swapchain destructor tries to re-destroy the same `VkImage` handles. 

The RAII samples only use `vk::Image`, and I can't think of a correct of the RAII handle, as the opposite ordering will lead to the same double-free, primarily because destroying the swapchain means destroying the `VkImage`s.

sharadhr:
This is why I should RTFM—I ran into this very issue a while ago, and the [manual](https://github.com/KhronosGroup/Vulkan-Hpp/blob/88d2ac296b5fbf7e6bc5af6d323f484cea67ad3d/vk_raii_ProgrammingGuide.md?plain=1#L185) states as much: 

> Note, that you don't get `vk::raii::Image`s here, but plain `VkImage`s. They are controlled by the swap chain, and you should not destroy them.

And additionally: 

> For a depth buffer, you need an image and some device memory and bind the memory to that image. That is, you first create a `vk::raii::Image` 

And so on.

I will leave this issue in case anyone else runs into this...

sharadhr:
I want to reopen this—in fact, I'd like to re-cast the question. How would someone use `vk::raii::Image` in their engine or program? 

[...]

PR #2280, opened 2025-09-08
Expose "raw" RAII constructors

YaaZ:
Some use cases, like creating Buffers/Images via the [VMA library](https://github.com/YaaZ/VulkanMemoryAllocator-Hpp) require an ability to create RAII handles from "raw" components (e.g. C handle, allocation callbacks, dispatcher).
See the [somewhat related discussion](https://github.com/KhronosGroup/Vulkan-Hpp/issues/2272#issuecomment-3265803413).

### DIFF
--- a/VulkanHppGenerator.cpp
+++ b/VulkanHppGenerator.cpp
@@ -8898,7 +8898,7 @@ std::string VulkanHppGenerator::generateRAIIHandle( std::pair<std::string, Handl
 
     auto [singularConstructors, arrayConstructors] = generateRAIIHandleConstructors( handle );
 
-    auto [clearMembers, getConstructorSuccessCode, memberVariables, moveConstructorInitializerList, moveAssignmentInstructions, swapMembers, releaseMembers] =
+    auto [clearMembers, getConstructorSuccessCode, memberVariables, rawConstructorArguments, rawConstructorInitializerList, moveConstructorInitializerList, moveAssignmentInstructions, swapMembers, releaseMembers] =
       generateRAIIHandleDetails( handle );
 
     std::string declarations = generateRAIIHandleCommandDeclarations( handle, specialFunctions );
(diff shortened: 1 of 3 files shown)


YaaZ:
For the purposes of VMA-Hpp the raw constructor can also be made protected to avoid polluting the public API - let me know your thoughts

asuessenbach:
I don't get, why you need an additional set of constructors, that take (besides other arguments) a raw handle.
There are already such constructors available, like
```
      Image( Device const & device, VkImage image, Optional<const AllocationCallbacks> allocator = nullptr )
        : m_device( device ), m_image( image ), m_allocator( static_cast<const AllocationCallbacks *>( allocator ) ), m_dispatcher( device.getDispatcher() )
      {
      }
```

YaaZ:
First off, I was thinking about that patch and decided that it's better to propose making RAII handle fields protected instead. The change would be less invasive and classes inheriting `vk::raii` handles would be able to initialize the fields manually.

Answering your question:
Image/Buffer constructors accept `vk::raii:Device` and use the dispatcher from it - we need an owning handle, which we don't have in VMA:
VMA allocator (proposed `vma::raii::Allocator`) is created from `vk::raii:Device`, but does not own it. Allocator cannot keep a pointer to the `vk::raii:Device` because the latter can be moved. Therefore, allocator keeps a plain `vk::Device` handle and pointer to the dispatcher, just like many other RAII handles. But then, in order to create an Image/Buffer from that allocator, it needs an **owning** device handle - `vk::raii:Device`, which allocator doesn't have and has no ability to recreate a "temporary" one either.

So the proposed solution is to create RAII Image/Buffer from `vk::Device` and dispatcher directly, bypassing the owning `vk::raii:Device` handle. Again, in case of VMA it can be done by simply changing field visibility to `protected`, as Images/Buffers created by VMA would inherit `vk::raii` types.

YaaZ:
Updated proposal - protected "raw" constructors only for `vk::raii::Buffer` and `vk::raii::Image`, following [the discussion](https://github.com/KhronosGroup/Vulkan-Hpp/issues/2272#issuecomment-3305613152).

asuessenbach:
Resolves #2272.

PR #2269, opened 2025-09-01
C++Module: Fixed lack of cstdlib with import std on Clang

M2-TE:
One of the options to resolve #2266

PR #2283, opened 2025-09-11
C++Module: Fixing import std on enum header and using `export import std` when appropriate

### LINKED ISSUE #2281: `VULKAN_HPP_USE_STD_MODULE` should `export import std;`
The C++ standard library is explicitly part of the public interfaces exposed by Vulkan-Hpp, from `std::vector` and `std::string` to constructors accepting `std::array`, as well as the `<compare>` header. 

If consumers subscribe to using the `std` module to build Vulkan-Hpp, then we should implicitly `export import` it, especially to support that final case of `<compare>` not being available.

M2-TE:
Resolves #2281.

This issue pops up when using the vulkan headers and `import std` on gcc 15.
The enums header had a leftover `#include <type_traits>` that was not properly wrapped for `import std` and would redefine the symbols within.

I've also cleaned out some headers out of `vulkan.cppm` that are already being included in `vulkan.hpp`.

### DIFF
--- a/.github/workflows/ci-ubuntu-ext.yml
+++ b/.github/workflows/ci-ubuntu-ext.yml
@@ -131,6 +131,7 @@ jobs:
               -D VULKAN_HPP_DISPATCH_LOADER_DYNAMIC=ON                \
               -D VULKAN_HPP_BUILD_WITH_LOCAL_VULKAN_HPP=ON            \
               -D VULKAN_HPP_PRECOMPILE=OFF                            \
+              -D VULKAN_HPP_NO_STD_MODULE=ON                          \
               -D CMAKE_CXX_COMPILER=clang++-${{matrix.clang_version}} \
               -D CMAKE_CXX_STANDARD=$CXX_STANDARD                     \
               -D CMAKE_BUILD_TYPE=$BUILD_TYPE


snippets/CppmTemplate.hpp
```diff
@@ -5,8 +5,6 @@
 
 module;
 
-#include <vulkan/vulkan_hpp_macros.hpp>
```
sharadhr, line 8: Why was this removed? This header provides the definition for `VULKAN_HPP_STD_MODULE` and `<version>`.

M2-TE: I had removed it since it was a duplicate include (`vulkan.hpp` includes it again as well).

Since we are on the topic: What is the rationale behind defining `VULKAN_HPP_STD_MODULE` to `std.compat`? I thought that was an MSVC-only thing. I had thought of removing the need for that macro entirely, as `std` seems to be standard across all compilers now.

sharadhr: >I had removed it since it was a duplicate include (vulkan.hpp includes it again as well).

It is duplicate, but the macros are needed before we have `#include <vulkan/vulkan.hpp>`; otherwise, that feature test in line 10 will not work, and nor will many of the other configurations. This was previously discussed in the `import std;` PR: #1932. 

> What is the rationale behind defining VULKAN_HPP_STD_MODULE to std.compat

`std.compat` exposes the global-namespace C library functions, like `::fopen`, etc. It is not MSVC-specific. See: https://eel.is/c++draft/std.modules#3

M2-TE: Thanks for the clarification on std.compat! Vulkan-Hpp doesn't really need it, but i suppose people would not want to import both `std` and `std.compat` at various places.

The feature test in line 10 does not seem to work for me either way. `__cpp_lib_modules` is never defined, even though I am actively using `import std`. That aside, it should only require `#include <version>` to work, not the entire macro header, right?

It might also make sense to remove the check for `__cpp_lib_modules` when `VULKAN_HPP_ENABLE_STD_MODULE` is defined but `VULKAN_HPP_STD_MODULE` is not. Since modules are requested and no specific module is given, we should just default to what is currently `std.compat`. Or, at the very least, throw some sort of compile-time error. For me it silently didnt use `import std` until I manually defined `VULKAN_HPP_STD_MODULE`.

sharadhr: > `__cpp_lib_modules` is never defined

This implies a bug somewhere. It must be defined by the standard library when `export std` is declared and when the language mode is C++23. On MSVC the macro is correctly exposed in `<version>`, which in turn includes `<yvals_core.h>`.

M2-TE: So to summarize what I've added:
1. The dynamic enabling of `VULKAN_HPP_ENABLE_STD_MODULE` was moved into the macro header. It has access to `<version>` there and will propagate the macro to both the `vulkan.cppm` and `vulkan.hpp` headers, as well as all the headers included within them. It doesn't strictly need to be in the `vulkan.cppm` global module fragment.
2. `VULKAN_HPP_STD_MODULE` was automatically defined when it was missing with `VULKAN_HPP_ENABLE_STD_MODULE` being present. The extra check for `__cpp_lib_modules` is not really necessary in this case and also prevents issues with the module being undefined when users dont supply their own definition (which is a compiler issue, but an issue that affects people nonetheless).

@sharadhr are there any concerns I left unaddressed?

snippets/CppmTemplate.hpp
```diff
@@ -5,21 +5,12 @@
...
-#endif
-
 #include <vulkan/${api}.hpp>
 #include <vulkan/${api}_extension_inspection.hpp>
 #include <vulkan/${api}_format_traits.hpp>
 #include <vulkan/${api}_hash.hpp>
 #include <vulkan/${api}_raii.hpp>
 #include <vulkan/${api}_shared.hpp>
-#ifndef VULKAN_HPP_NO_TO_STRING
-#include <vulkan/${api}_to_string.hpp>
```
asuessenbach, line 21: Why have you removed this?

M2-TE: This is already being conditionally pulled in by `vulkan.hpp`:
https://github.com/KhronosGroup/Vulkan-Hpp/blob/4bbdbb434d3e8993948cf57a2f70421dae41d149/vulkan/vulkan.hpp#L6759-L6762

vulkan/vulkan_hpp_macros.hpp
```diff
@@ -87,7 +87,11 @@
...
 #endif
 
-#if defined( __cpp_lib_modules ) && !defined( VULKAN_HPP_STD_MODULE ) && defined( VULKAN_HPP_ENABLE_STD_MODULE )
+#if defined( __cpp_lib_modules ) && !defined( VULKAN_HPP_ENABLE_STD_MODULE )
+#  define VULKAN_HPP_ENABLE_STD_MODULE
+#endif
+
+#if !defined( VULKAN_HPP_STD_MODULE ) && defined( VULKAN_HPP_ENABLE_STD_MODULE )
 #  define VULKAN_HPP_STD_MODULE std.compat
 #endif
```
M2-TE: CI is failing on some configurations, where it cannot find `import std.compat` in `vulkan.hpp` in cases where modules are not available in general. From what I can see, this could only be enabled by `__cpp_lib_modules` being defined (aka theoretical support being available), but the toolchain obviously isn't set up for module workflows.

@sharadhr shouldn't this only be defined if `import std` is actually available or am I missing something here? It would be nice to get "automatic" usage of `import std`, even if people just use the `vulkan.hpp` header instead of the `vulkan_hpp` module.

If it gets defined even for just theoretical support, I will have to add another check for `VULKAN_HPP_CXX_MODULE` or something and setting that define within the `vulkan.cppm` global module fragment. It would allow users to "manually" request `import std` with only `vulkan.hpp`.

sharadhr: `__cpp_lib_modules` must be provided by the toolchain when C++23 is enabled. 

[libc++](https://github.com/llvm/llvm-project/blob/7936b6f1133516c427c478104606d9276da6f81b/libcxx/include/version#L509)

[libstdc++](https://gcc.gnu.org/git/?p=gcc.git;a=blob;f=libstdc%2B%2B-v3/include/bits/version.h;h=b05249857d2723c5f66aff8389e4f25cb764fdff;hb=HEAD#l2280)

M2-TE: Did you ever confirm that it would actually be provided within `vulkan.cppm`?
We test modules with C++23 in CI (and I also use modules myself with C++23), but it never automatically used `import std`. It would have caused an error, as the toolchains (including CMake) are not set up for it.

sharadhr: At this point I wonder if `VULKAN_HPP_STD_MODULE` is necessary at all. I feel we should change the usage model: if we are using `vulkan.cppm` and are in C++23, we should unconditionally opt into using and exporting the std module rather than making it configurable with `VULKAN_HPP_STD_MODULE`, probably with `#define VULKAN_HPP_ENABLE_STD_MODULE 1` in the GMF of `vulkan.cppm`.

`vulkan.cppm` with C++ headers doesn't really make sense anyway, since compilers don't like mixing headers and modules that declare the same symbols, and secondly because the C++ headers have problems with not properly exporting necessary symbols from `<compare>` etc.

M2-TE: > vulkan.cppm with C++ headers doesn't really make sense anyway

I don't agree, as they can very much be used and issues are easily circumvented (e.g. manually including `<compare>`. But I do agree that it would be nice to use `export import std` by default rather than headers.

The problem remains that even in C++23, CMake and co still lack the proper support for `import std`, so defaulting to it is not a good idea as of yet I believe.

sharadhr: > as they can very much be used and issues are easily circumvented (e.g. manually including <compare>

I strongly disagree. The module exports C++ standard library APIs (e.g. `vk::raii::CommandBuffers` is a child class of `std::vector`), and expecting consumers to additionally include headers as a workaround is, in my view, unacceptable. We should expose everything that we use, and if we expect the STL, we should have it in the module. Just like how `#include <vulkan/vulkan.hpp>` textually includes everything that the header provides, including parts of the C++ standard library.

> C++23, CMake and co still lack the proper support for import std

CMake's support being experimental is orthogonal to the compiler and standard library support being present. All three compilers now provide `std` and `std.compat`, and the standard library module can be used even with Makefiles: 

- https://0xstubs.org/using-the-c23-std-module-with-clang-18/
- https://0xstubs.org/using-the-c23-std-module-with-gcc-15/

M2-TE:
I've opted to remove the automatic defining of `VULKAN_HPP_ENABLE_STD_MODULE` for the following reasons:
1. Before the changes in this patch, it was present in the global module fragment of `vulkan.cppm`, where it seems to have never been in effect. `__cpp_lib_modules` was never being defined, even in C++23, as our CI workflows that test modules would have failed. The fact that C++23 workflows fail now that it is in the normal macro header shows this.
2. As of yet, I believe toolchains are not yet ready to automatically use `import std` when using C++23. The best example is CMake, where you have to specifically set experimental flags to even get it to work. Thus, most people would likely be unable to use the `vulkan_hpp` module without it.

We can rethink the automatic inclusion later, but it should not happen in the vulkan.hpp header or sub-includes, as it affects projects not set up to deal with `import std` or even just modules. It should also not be in the global module fragment, as it makes more sense to `export import std` as is currently being discussed in #2284.

vulkan/vulkan_hpp_macros.hpp
M2-TE: > All three compilers now provide std and std.compat, and the standard library module can be used even with Makefiles

It can be used _in theory_. It requires extra work that should not be expected of people who just want to use the module. As far as I know (and I could be wrong) most people use CMake to configure their Vulkan projects.

I believe using the vulkan module should work out-of-the-box, which project configurators like make or cmake are currently _not_ able to do.

Opt-in for modules is currently the way to go imo. Even if not, there absolutely still needs to be an option to at least opt out of using `import std` in `vulkan_hpp`.

vulkan/vulkan_hpp_macros.hpp
```diff
@@ -87,7 +87,7 @@
 #  define VULKAN_HPP_SUPPORT_SPAN
 #endif
 
-#if defined( __cpp_lib_modules ) && !defined( VULKAN_HPP_STD_MODULE ) && defined( VULKAN_HPP_ENABLE_STD_MODULE )
```
sharadhr, line 90: I also don't follow why this feature test macro was removed.

sharadhr:
@M2-TE could you in summary explain how users would use the module? There are four possible scenarios now:

- C++20, `import std` disabled
- C++20, `import std` enabled (possible with extension)
- C++23, `import std` disabled 
- C++23, `import std` enabled

I believe the third configuration ought to be impossible, as a C++23-compliant toolchain_must_ ship `std`.

vulkan/vulkan_hpp_macros.hpp
M2-TE: I don't think it serves a purpose here. If `VULKAN_HPP_ENABLE_STD_MODULE` is defined, this should simply supply a default `VULKAN_HPP_STD_MODULE` if it was not yet defined.

M2-TE:
There are two additional scenarios:
* C++20, `vulkan_hpp` in C++23, `import std` enabled
* C++20, `vulkan_hpp` in C++23, `import std` disabled

as these modules with import std can be consumed just fine by C++20 sources. Important for projects that cannot yet upgrade their entire source code to C++23.

The third configuration you showed ought to be impossible, if the C++ versions are all we consider. But people do not write their own configuration tools and instead use Make or CMake, so we cannot just look at compilers in isolation. E.g.:

* CMake 3.30 first introduced experimental `import std` support, but it was broken for quite a few configurations. It could not automatically use libstd++ for clang for example and would fail with gcc.
* CMake 4.1 was the first stable release for which `import std` works in most cases, under the condition that you still set a specific experimental flag. So not out-of-the-box yet.

Stable distros like Ubuntu have not even shipped 4.1 yet (e.g. Ubuntu 25.04 is at CMake 3.31).
There is also the concern that Vulkan-Hpp may introduce some C++23-only feature, that people may want to use.

sharadhr:
> - C++20, vulkan_hpp in C++23, import std enabled
> - C++20, vulkan_hpp in C++23, import std disabled

These use cases seem counterintuitive given the delivery machinery of Vulkan-Hpp, which is source-only. It is impossible for Vulkan-Hpp to be somehow compiled a version ahead of that on client machines, as the BMI is expected to be built on said machines anyway, since it is a header-only library.

> if the C++ versions are all we consider

Additionally, my impression is that C++ libraries are in general, binary-incompatible—that is, it doesn't make sense to write C++20 code against a C++23 library (or vice versa). From my experience, binary, source-unavailable libraries are shipped to be used with a specific toolchain version: that is, a specific compiler, language standard, and compiler flags. Naturally we don't consider this as Vulkan-Hpp is source-only. 

I don't think we should rely CMake's behaviour unless it affects internal testing. In fact even `vulkan.cppm` isn't correctly shipped in the `Vulkan-Headers` CMake config file because CMake doesn't correctly install module interfaces, and users have to manually set it up as detailed in the read-me. If they can build the BMI with C++23, then they can build their entire project with C++23 as well.

___

I intend to propose an alternative solution that removes all `import` from the headers (that is, effectively revert #1932), since most compilers have now fixed the original issue of mixing headers and `import std`, and instead unconditionally including the standard library headers, and then straightforwardly `export import std;` under C++23, like so:

## `vulkan_hpp_macros.hpp`
```cpp
#if defined( __cpp_lib_modules )
#  define VULKAN_HPP_ENABLE_STD_MODULE 1
#  define VULKAN_HPP_STD_MODULE std.compat
#endif
```

## `vulkan.cppm`
```cpp
module;
// ... other includes

export module vulkan_hpp;
#if defined(VULKAN_HPP_ENABLE_STD_MODULE) 
export import VULKAN_HPP_STD_MODULE;
#endif

namespace VULKAN_HPP_NAMESPACE { // ...
```

Consumers affected by this should upgrade compilers, as this scenario is supported in C++20 and 23, and if it isn't, it is a compiler/toolchain bug. It bears mentioning that `vulkan.cppm` is _still_ experimental, and these breaking changes are expected, especially between versions.

M2-TE:
> It is impossible for Vulkan-Hpp to be somehow compiled a version ahead of that on client machines, as the BMI is expected to be built on said machines anyway, since it is a header-only library.

It is not impossible; the only reason some user may still use C++20 for their own code could be an incompatibility with the newer standard, unrelated to the compiler that may very much be capable of compiling for C++23.

> It is impossible for Vulkan-Hpp to be somehow compiled a version ahead of that on client machines, as the BMI is expected to be built on said machines anyway, since it is a header-only library. [...] Naturally we don't consider this as Vulkan-Hpp is source-only.

Modules with `import std` are an exception. They require C++23 to be compiled, but only C++20 to be consumed as they do not expose anything that a normal C++20 library couldn't. CMake even has a way of handling this:

```cmake
target_compile_features(vulkan_hpp
  PRIVATE   cxx_std_23  # compile with C++23
  INTERFACE cxx_std_20) # consumable as C++20
```
So you still use the global compiler flags etc, except for a bumping the C++ version to enable `import std`.

> I don't think we should rely CMake's behaviour unless it affects internal testing

Then at least make it possible to opt-out to not completely ignore the general usability of the module under C++23.

> and instead unconditionally including the standard library headers, and then straightforwardly export import std;

Does this not go against the idea of modules in the first place; we would be unnecessarily increasing binary size by including headers and only doing the `import std` afterwards.

-----

I would propose to have an opt-out solution for C++23 in the following manner, akin to what you have shown:

### vulkan_hpp_macros.hpp

```cpp
#if !defined( VULKAN_HPP_STD_MODULE ) && defined( VULKAN_HPP_ENABLE_STD_MODULE )
#  define VULKAN_HPP_STD_MODULE std.compat
#endif
```

### vulkan.cppm

```cpp
module;

#include <version>
#if defined( __cpp_lib_modules ) && !defined( VULKAN_HPP_NO_STD_MODULE )
#  define VULKAN_HPP_ENABLE_STD_MODULE
#endif

#include <vulkan/vulkan.hpp>
// ... other includes (vulkan_hpp_macros.hpp already included via vulkan.hpp)

export module vulkan_hpp;
#if defined( VULKAN_HPP_ENABLE_STD_MODULE )
   export import VULKAN_HPP_STD_MODULE
#endif

namespace VULKAN_HPP_NAMESPACE { // ...
```

This way, users can:
1. Manually define `VULKAN_HPP_ENABLE_STD_MODULE` in case their compiler does not (bugged or otherwise)
2. Not define `VULKAN_HPP_STD_MODULE` and get a default (std.compat)
3. Opt-out by setting `VULKAN_HPP_NO_STD_MODULE` similar to other features provided by Vulkan-Hpp

sharadhr:
I'm happy with the opt-out! This will be more in line with the rest of Vulkan-Hpp's config options, where opt-out is the default. That said, is there a possibility to simplify `VULKAN_HPP_NO_STD_MODULE` and `VULKAN_HPP_ENABLE_STD_MODULE`? They appear to be negations of each other, could we only use one?

M2-TE:
Hm.. I don't think we can only use one. We need `VULKAN_HPP_NO_STD_MODULE` for opting out, so it can't be checked for enabling `import std`, only for disabling it. The "enabler" flag therefore needs to be separate, which is `VULKAN_HPP_ENABLE_STD_MODULE`. Though it's currently only necessary, because `__cpp_lib_modules` isn't defined in a C++23 module (seems to be the case for all 3 major compilers currently, should be a topic for a separate issue + discussion as to why). So we might be able to merge the flags once that issue is solved!

Current TODO:
* add proposed changes
* update README for the new opt-out flag
* add new flag to the CMake file

CI testing for std is something I'm already working on, so that will be a separate PR.

sharadhr:
> Does this not go against the idea of modules in the first place; we would be unnecessarily increasing binary size by including headers and only doing the import std afterwards.

Side note, a proper compiler would discard everything in the global module fragment that is not explicitly `export`ed in the module interface. However this depends on compiler: MSVC does it correctly, Clang does not.

M2-TE:
> * add proposed changes
> * update README for the new opt-out flag
> * add new flag to the CMake file

Implemented these changes mentioned before and also added the `VULKAN_HPP_NO_STD_MODULE` to CI to prevent potential issues later down the line. From what I can see, Clang does not define `__cpp_lib_modules` at all, while gcc defines it even in C++20. This also resolves #2281.

It turned out to be a little more involved than I thought at first, but it's ready for merging, so feel free to give it a final glance @sharadhr and @asuessenbach!

sharadhr:
The conditional `import std;` ought to be removed from all the headers, now that we're doing it once in the module interface file.

Also...

> Clang does not define `__cpp_lib_modules` at all

See [libc++ <version> header](https://github.com/llvm/llvm-project/blob/4c7de930b3e265f04d86425cbbbc5f78536ce4b2/libcxx/include/version#L509). Also note it's not the compiler that defines the macro, but rather the standard library, when loaded by the compiler under a certain language version, conditionally enables the feature test macro. 

Note that if you're using Clang on a Mac, AppleClang is not Clang in general, and AppleClang has disabled functionality that Clang otherwise has, especially modules. The versioning scheme has also diverged. 


> while gcc defines it even in C++20

This is a valid extension in C++20: see [MSVC STL discussion](https://github.com/microsoft/STL/issues/3945).

> The changes will also resolve https://github.com/KhronosGroup/Vulkan-Hpp/issues/2281.

Closed my PR, it makes sense to combine that with this.

M2-TE:
Thanks for the clarification on the `__cpp_lib_modules` macro!

> The conditional `import std;` ought to be removed from all the headers, now that we're doing it once in the module interface file

I think we should open a separate issue for discussion about that. At least, I'd like to address that in a separate PR.

sharadhr:
> I think we should open a separate issue for discussion about that. At least, I'd like to address that in a separate PR.

I feel it needs to be done with this PR, especially now that we have the `import export`. There's no real reason to separate it; it was written in the first place as a workaround for a compiler bug.

M2-TE:
In that case, removing the conditional `import std` in `vulkan.hpp` will mean that we are not really using `import std` at all.
`vulkan.cppm` imports it after the global module fragment, but at this point it is no longer even needed. All it would do is export it, fixing issues like the lack of `<compare>`.

I do like the current solution, where even the vulkan headers are capable of using `import std` themselves. Even if the compiler throws away everything not explicitly `export`ed, it helps reduce compile times. The duplicate import does not have an impact afaik (first `import std` in `vulkan.hpp`, then `export import std` in `vulkan.cppm`).

sharadhr:
> at this point it is no longer even needed. All it would do is export it, fixing issues like the lack of <compare>.

As I've mentioned previously a few times, `std::` symbols are part of the public interface of Vulkan-Hpp; this line is necessary for consumers of Vulkan-Hpp to correctly use it. `<compare>` is not the only example; things like `vk::raii::CommandBuffers` break without an `import std;` from the consuming code. 

I want to enforce the strong coupling of: if you're using headers, then you're always using headers; if you're using the `vulkan_hpp` module, then everything is a module. Having over-configurability also massively complicates things.

M2-TE:
> at this point it is no longer even needed.

I should clarify this line; I meant that the importing is no longer needed at this point. Exporting is still very much important, as you have noted a few times already. Sorry about the wording!

> I want to enforce the strong coupling of: if you're using headers, then you're always using headers; if you're using the vulkan_hpp module, then everything is a module. Having over-configurability also massively complicates things.

Exactly why I'm against removing the conditional includes/imports from the normal vulkan headers. 
* When using `vulkan.hpp`, std headers should be included.
* When using the `vulkan_hpp` module, std module should be imported (`vulkan.hpp`) and exported (`vulkan_hpp`).

There's nothing about what you said that I disagree with, I just seem to misunderstand why you want to remove `import std` from the vulkan headers. If we do that, we will always have case 1 where we just include std headers even with the vulkan module.

YaaZ:
@M2-TE I think a few includes in [StructsHppTemplate](https://github.com/KhronosGroup/Vulkan-Hpp/blob/938a2c36d2d3886a293c63c9a26417d6b0e2bc2d/snippets/StructsHppTemplate.hpp#L9) also worth guarding, could you take a look?

sharadhr:
@YaaZ, for some reason symbols in `cstring` and `cstdlib` (and C stuff in general) aren't exported by `std`. 

I can't find the relevant standards section, but [here is a discussion](https://cpplang.slack.com/archives/C92GZLCSE/p1725128701509549) in the Cpplang Slack when I asked about this, specifically to fix Vulkan-Hpp's module.

sharadhr:
@M2-TE It seems like compilers still aren't happy about mixing `#include <std_header>` and `import std;`, so it probably makes sense to leave the guards in... However, I would prefer if we had only _one_ import statement. As it is, this PR will cause the textual result of `vulkan.cppm` to have _multiple_ `import std.compat;`

I recommend that the `import ...` be removed from the _headers_, and be placed into `vulkan.cppm` instead, and simplify the guards to, for instance:

```cpp
#if !( defined( VULKAN_HPP_ENABLE_STD_MODULE ) && defined( VULKAN_HPP_STD_MODULE ) )
#  include <type_traits>    // for std::underlying_type
#endif
```

And at the top of `vulkan.cppm`, before `#include <vulkan/vulkan.hpp>`, we can have:

`import std;`

and within the module interface, we can instead do:

`export std;`

Is this reasonable?

YaaZ:
@sharadhr cppreference [says](https://en.cppreference.com/w/cpp/standard_library.html#Importing_modules) that `std` and `std.compat` provide [C++ headers for C library facilities](https://en.cppreference.com/w/cpp/standard_library.html#Headers) - `cstring` and `cstdlib` included.
I think we need to guard all includes unless absolutely necessary (like those providing macros).

I also agree that it looks better to have all includes guarded in one place.

sharadhr:
@YaaZ it would seem so, until we ran into issues that had to be resolved by explicitly having those includes, for instance #2269.

M2-TE:
@sharadhr sadly, using `export std` is not legal. Module names may only be used on `export module [...]` `import [...]` or `export import [...]` statements: [https://en.cppreference.com/w/cpp/language/modules.html](https://en.cppreference.com/w/cpp/language/modules.html).

So I believe we are stuck with using `import std` in the header for now

sharadhr:
Right, then we can leave it as `export import std;`.

M2-TE:
Placing `import std` into the `vulkan.cppm` GMF (before the `#includes`) and removing it from `vulkan.hpp` seems to break a few symbols like `std::exchange` and more on Clang. Can you test this on MSVC to see if this is compiler-specific?

sharadhr:
Hmm, I see something similar on MSVC:

```cpp
module;

import std;

export module Module;

export auto MyFunc()
{
	return 42ULL * std::numbers::pi;
}
```

With the warning:

```
1>R:\ConsoleApplication1\Module.ixx(3,11): warning C5202: a global module fragment can only contain preprocessor directives
```

This is only a _warning_, but the project compiles and runs nonetheless, with a consuming `main.cpp`:

```cpp
#include <iostream>

import Module;

int main()
{
    std::cout << MyFunc() << std::endl;
}
```

There is a [report on DevComm](https://developercommunity.visualstudio.com/t/Modules-Bogus-warning-C5202/10397891?sort=active) about this, but then there is also [this Reddit thread](https://www.reddit.com/r/cpp/comments/1ihbb9f/can_i_put_import_inside_the_global_module_fragment/) that claims writing `import` directly in the GMF is not allowed (but even there the wording that disallows this is... unclear).

M2-TE:
looksinside  

Either way, I suppose the key takeaway is that `import` is theoretically a preprocessing directive and should be allowed to be placed into the GMF. Considering that it doesn't seem to work yet on major compilers however, we should probably wait until we do it this way.

Is there anything left to do on this PR that I overlooked?

sharadhr:
A+ meme. 

If having multiple `import` works then let's leave it, not the most ideal but if it works it works. Clearly module support is still broken... :/

sharadhr:
And @M2-TE no other comments; I appreciate the discussion. Thanks :)

M2-TE:
Likewise, thank you very much for your help and discussions @sharadhr!

asuessenbach:
@M2-TE and @sharadhr: It's great to have you both contributing here! It really moves forward the modules usage on Vulkan-Hpp.
Thanks a lot!

@M2-TE Would you please also adjust the newly added `ci-ubuntu-ext.yml`?

sharadhr:
Perhaps it's a better idea to rebase than merge...?

M2-TE:
Doesn't it all get squashed on PR merge anyways?

asuessenbach:
Ok then, thanks a lot for your contribution!

PR #2284, opened 2025-09-11
`export import VULKAN_HPP_STD_MODULE;`

### LINKED ISSUE #2281: `VULKAN_HPP_USE_STD_MODULE` should `export import std;`

sharadhr:
Resolves #2281.

### DIFF
--- a/vulkan/vulkan.cppm
+++ b/vulkan/vulkan.cppm
@@ -28,6 +28,8 @@ module;
 
 export module vulkan_hpp;
 
+export import VULKAN_HPP_STD_MODULE;
+
 export namespace VULKAN_HPP_NAMESPACE
 {
   //=====================================


M2-TE:
You forgot to run the generator on this one ;)
Also the `export import VULKAN_HPP_STD_MODULE` should be done in `vulkan.hpp`, not the `vulkan.cppm` module. It needs the include guards `#if defined( VULKAN_HPP_ENABLE_STD_MODULE ) && defined( VULKAN_HPP_STD_MODULE )` present there, as the std module is optional.

sharadhr:
Ah, let me fix that. 

> Also the export import VULKAN_HPP_STD_MODULE should be done in vulkan.hpp, not the vulkan.cppm module. It needs the include guards #if defined( VULKAN_HPP_ENABLE_STD_MODULE ) && defined( VULKAN_HPP_STD_MODULE ) present there, as the std module is optional.

This is not allowed, and it's why I put it beneath `export module vulkan_hpp;`. `export` can only happen in the module purview and not the global module fragment, which only allows preprocessor statements (which *does* include `import`).

M2-TE:
> export can only happen in the module purview and not the global module fragment

You are correct, export import needs to happen inside the module.

Without any conditional macro guards though, you will be breaking the module for a lot of people who use it. Importing std is still very much experimental even in newest CMake (4.1), where you have to specifically go through hoops to opt-in.

PR #2285, opened 2025-09-12
CI: Added bleeding-edge ubuntu ci with clang 19/20 and newest cmake (4.1)

M2-TE:
This new CI is supposed to test newest compiler versions that are not available on the runners by default yet, hopefully catching errors before they make it into stable distro releases.
It pulls newest builds for clang and cmake from the publishers apt repos.

The plan is to also use these to test `import std` in the near future, as newer version of CMake are required (min `3.30`, though much more stable in `4.1`).

### DIFF
--- a/.github/workflows/ci-ubuntu-ext.yml
+++ b/.github/workflows/ci-ubuntu-ext.yml
@@ -1,0 +1,142 @@
+
+name: CI Ubuntu 24
+
+on:
+  pull_request:
+    branches: [ main ]
+
+jobs:
+  build:
+    runs-on: ubuntu-24.04
+
+    defaults:
+      run:
+        shell: bash
+
+    strategy:
+      matrix:
+        clang_version: [19, 20]
+    
+    steps:
+    - uses: actions/checkout@v4
+      with:
+        submodules: recursive
+
+    - name: Install libraries
+      run: sudo apt update && sudo apt install libgl-dev libxcursor-dev libxi-dev libxinerama-dev libxrandr-dev
+
+    - name: Install Ninja
+      uses: ashutoshvarma/setup-ninja@master
+      with:
+        version: 1.11.0
+    - name: Install Clang and CMake
+      run: |
+        sudo apt-get update && sudo apt-get install lsb-release gpg software-properties-common wget
+        wget https://apt.llvm.org/llvm.sh
+        chmod +x llvm.sh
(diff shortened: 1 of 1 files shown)


.github/workflows/ci-ubuntu-ext.yml
```diff
@@ -0,0 +1,142 @@
+
+name: CI Ubuntu 24
```
asuessenbach: You should at least name this a bit different.

.github/workflows/ci-ubuntu-ext.yml
```diff
@@ -0,0 +1,142 @@
...
+          -D CMAKE_CXX_COMPILER=clang++-${{matrix.clang_version}}   \
+          -D CMAKE_CXX_STANDARD=20                                  \
+          -D CMAKE_BUILD_TYPE=Release
+        cmake --build build --parallel --clean-first
+
+    - name: Build samples and tests
+      run: |
+        for CXX_STANDARD in 11 14 17 20 23; do
+          for BUILD_TYPE in Debug Release; do
+            if [ clang++-${{matrix.clang_version}} == clang++-17 ] && [ $CXX_STANDARD == 23 ]; then
```
asuessenbach: "clang++-17" can't happen here.

asuessenbach (commented):
The differences (besides the "Install Clang and CMake" section) to `ci-ubuntu.yml` are small, but I assume integrating this PR into that file would make it substantially more complex.

That is, at least the two comments below should be fixed.

.github/workflows/ci-ubuntu-ext.yml
M2-TE: Yup! I chose `CI Ubuntu EXT` for this

.github/workflows/ci-ubuntu-ext.yml
M2-TE: You are right, that's a remnant from the ubuntu 24 CI

M2-TE:
> The differences (besides the "Install Clang and CMake" section) to ci-ubuntu.yml are small, but I assume integrating this PR into that file would make it substantially more complex.

Yea, they are not "standard" runners per se, so I thought it would make sense to separate them. The only real differences are the Clang and CMake versions.

Once the runner for Ubuntu 26 comes out, this workflow can be replaced by that one.

asuessenbach:
That's a great extension of our CI!

PR #2288, opened 2025-09-16
Fix: Correct IWYU pragmas

dv-001:
This builds on #1936 and #1939, correcting typos and file paths, as well as addressing false-positive unused include warnings.

As @rwols pointed out in [this comment](https://github.com/KhronosGroup/Vulkan-Hpp/pull/1939#discussion_r2346016464), the IWYU [private](https://github.com/include-what-you-use/include-what-you-use/blob/master/docs/IWYUPragmas.md#iwyu-pragma-private) pragma uses a comma, not a semicolon. I have corrected this and also updated the paths from `"vulkan.hpp"` to `"vulkan/vulkan.hpp"`.

~In addition to this, I've added IWYU [export](https://github.com/include-what-you-use/include-what-you-use/blob/master/docs/IWYUPragmas.md#iwyu-pragma-export) pragmas to `vulkan/vulkan.hpp` so that tools like clangd no longer incorrectly flag it as being unused.~

Edit: It appears the `export` pragmas are unnecessary after correcting the `private` typos. They have been removed.

### DIFF
--- a/snippets/EnumsHppTemplate.hpp
+++ b/snippets/EnumsHppTemplate.hpp
@@ -4,7 +4,7 @@ ${licenseHeader}
 #  define VULKAN_ENUMS_HPP
 
 // include-what-you-use: make sure, vulkan.hpp is used by code-completers
-// IWYU pragma: private; include "vulkan.hpp"
+// IWYU pragma: private, include "vulkan/vulkan.hpp"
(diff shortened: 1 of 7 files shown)


rwols:
This works for me. Thanks for fixing it.

PR #2290, opened 2025-09-18
Export ResultValue related functions

YaaZ:
Another request for [VMA-Hpp](https://github.com/YaaZ/VulkanMemoryAllocator-Hpp):
VMA functions return `VkResult` codes and VMA-Hpp wraps them the same way Vulkan-Hpp do.
Now we have to copy the implementation of `createResultValueType`, `resultCheck` and `throwResultException`, as it's not exported from the `.cppm`. Maintaining this copy manually is burdensome, as sometimes error codes get deprecated and removed, breaking compilation.
It would be really nice if VMA-Hpp could use these functions directly from Vulkan-Hpp, maybe it could be helpful in other use cases where people have to deal with result codes.

I essentially implemented [this suggestion](https://github.com/KhronosGroup/Vulkan-Hpp/pull/1931#issuecomment-2273060760) to export the symbols under `vk::detail` namespace.

### DIFF
--- a/VulkanHppGenerator.cpp
+++ b/VulkanHppGenerator.cpp
@@ -5449,6 +5449,8 @@ std::string VulkanHppGenerator::generateCppModuleUsings() const
     using VULKAN_HPP_NAMESPACE::detail::DispatchLoaderStatic;
     using VULKAN_HPP_NAMESPACE::detail::getDispatchLoaderStatic;
 #endif /*VK_NO_PROTOTYPES*/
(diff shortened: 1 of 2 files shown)


asuessenbach:
You did not add `throwResultException`. Is that not needed, as it's only used internally?

YaaZ:
It's not used by VMA-Hpp directly and it also needs to be exported conditionally, so I decided to leave it for now, I think `resultCheck` should be enough.

asuessenbach:
Thanks a lot for you ongoing contribution!

PR #2291, opened 2025-09-18
Guard includes against VULKAN_HPP_STD_MODULE

YaaZ:
C++20 module with `std` fails to compile for me on MSVC with errors like:
```
MSVC\14.43.34808\include\xtr1common(26): error C2953: 'std::integral_constant': class template has already been defined
```
I traced it to the `#include <type_traits>` in `vulkan_enums.hpp`.
`#include <cstring>` in `vulkan_structs.hpp` doesn't cause problems for me, but also worth fixing.


YaaZ:
Damn, I should have checked that this was already [addressed](https://github.com/KhronosGroup/Vulkan-Hpp/pull/2283).

PR #2292, opened 2025-09-19
Tests: Enabling CTest alongside module testing with import std

### LINKED ISSUE #2331: MSVC with Ninja fails to compile SharedHandles sample
When trying to compile the `SharedHandle` sample with MSVC (2022) using `Ninja` (-G Ninja), it is not able to construct a shared handle of `vk::SurfaceKHR`:

https://github.com/KhronosGroup/Vulkan-Hpp/blob/0a7333161db0821452ad950760e48abc35ea0fa3/samples/SharedHandles/SharedHandles.cpp#L89

fails with the following output:

```bash
FAILED: [code=2] samples/SharedHandles/CMakeFiles/SharedHandles.dir/SharedHandles.cpp.obj 
C:\PROGRA~1\MICROS~2\2022\ENTERP~1\VC\Tools\MSVC\1444~1.352\bin\Hostx86\x86\cl.exe  /nologo /TP -DVK_NO_PROTOTYPES -DVK_USE_PLATFORM_WIN32_KHR -DVULKAN_HPP_DISPATCH_LOADER_DYNAMIC -DVULKAN_HPP_DISPATCH_LOADER_DYNAMIC=1 -ID:\a\Vulkan-Hpp\Vulkan-Hpp -ID:\a\Vulkan-Hpp\Vulkan-Hpp\Vulkan-Headers\include -ID:\a\Vulkan-Hpp\Vulkan-Hpp\glm -ID:\a\Vulkan-Hpp\Vulkan-Hpp\glfw\include -ID:\a\Vulkan-Hpp\Vulkan-Hpp\glslang\glslang\.. -ID:\a\Vulkan-Hpp\Vulkan-Hpp\build\include -ID:\a\Vulkan-Hpp\Vulkan-Hpp\glslang\SPIRV\.. /DWIN32 /D_WINDOWS /EHsc /Zi /Ob0 /Od /RTC1 -MDd /W4 /WX /showIncludes /Fosamples\SharedHandles\CMakeFiles\SharedHandles.dir\SharedHandles.cpp.obj /Fdsamples\SharedHandles\CMakeFiles\SharedHandles.dir\ /FS -c D:\a\Vulkan-Hpp\Vulkan-Hpp\samples\SharedHandles\SharedHandles.cpp
D:\a\Vulkan-Hpp\Vulkan-Hpp\samples\SharedHandles\SharedHandles.cpp(89): error C2440: 'initializing': cannot convert from 'initializer list' to 'vk::SharedHandle<vk::SurfaceKHR>'
D:\a\Vulkan-Hpp\Vulkan-Hpp\samples\SharedHandles\SharedHandles.cpp(89): note: 'vk::SharedHandle<vk::SurfaceKHR>::SharedHandle': function does not take 2 arguments
D:\a\Vulkan-Hpp\Vulkan-Hpp\vulkan/vulkan_shared.hpp(304): note: could be 'vk::SharedHandle<vk::SurfaceKHR>::SharedHandle(HandleType,vk::detail::ObjectDestroyShared<HandleType>) noexcept'
        with
        [
            HandleType=vk::SurfaceKHR
        ]
D:\a\Vulkan-Hpp\Vulkan-Hpp\vulkan/vulkan_shared.hpp(295): note: or       'vk::SharedHandle<vk::SurfaceKHR>::SharedHandle(HandleType,vk::SharedHandle<vk::Instance>,vk::SharedHandle<vk::NoDestructor>,const Dispatcher &) noexcept'
        with
        [
            HandleType=vk::SurfaceKHR
        ]
D:\a\Vulkan-Hpp\Vulkan-Hpp\vulkan/vulkan_shared.hpp(285): note: or       'vk::SharedHandle<vk::SurfaceKHR>::SharedHandle(HandleType,vk::SharedHandle<vk::Instance>,vk::detail::ObjectDestroyShared<HandleType>) noexcept'
        with
        [
            HandleType=vk::SurfaceKHR
        ]
D:\a\Vulkan-Hpp\Vulkan-Hpp\samples\SharedHandles\SharedHandles.cpp(89): note: while trying to match the argument list '(VkSurfaceKHR, vk::SharedInstance)'
```

M2-TE:
Builds on changes from #2248. Progress for #1993. Resolves #2331.

The tests were not well isolated and had to go through hoops to get e.g. modules with separate compilation flags.
They also did not use any of the convenient `option`s that are provided in our top-level `CMakeLists.txt`.

Running the tests via CTest gives us full control over them, avoiding current issues like compile definitions bleeding over, adding special compilation flags for certain tests (partiularly for dynamic vs static dispatch testing, it was previously impossible to run both). It will also allow to selectively enable `import std` (soon).

CTest also allows running individual tests only, making debugging easier. Also leads to easier error reading in CI logs by showing exactly which tests fail.

I've tried to logically separate the commits, but here's a summary of the changes:

* All targets in tests now use the default `Vulkan::Hpp` or `Vulkan::HppModule` (or Video variant), with no manual include directories that muddy the waters. The target setup was placed into `vulkan_hpp__setup_vulkan_targets` to get consistent path setups, no matter where it is included from. Modules would occasionally break until now due to relative paths, for example.
* Tests use e.g. `set(VULKAN_HPP_NO_CONSTRUCTORS ON)` as settings, rather than manually setting compile definitions. This required raising the minimum CMake version to 3.13 for the proper policy.
* More module tests were fixed and enabled
* CI was updated to run ctest

Note that the current setup will, unlike before, try to test both dynamic and static dispatch. I don't know how CI will handle this and I remember there being an issue with an internal git repo before. Likely a good idea to just conditionally disable static dispatch testing again before merging.

### DIFF
--- a/.github/workflows/ci-windows-clang.yml
+++ b/.github/workflows/ci-windows-clang.yml
@@ -1,5 +1,5 @@
 
-name: CI Windows Clang
+name: CI Windows ClangCL-Ninja
 
 on:
   pull_request:
--- a/.github/workflows/ci-windows-ext.yml
+++ b/.github/workflows/ci-windows-ext.yml
@@ -1,5 +1,5 @@
 
-name: CI Windows Clang
+name: CI Windows MSVC-Ninja
 
 on:
   pull_request:
--- a/.github/workflows/ci-windows-msvc.yml
+++ b/.github/workflows/ci-windows-msvc.yml
@@ -1,5 +1,5 @@
 
-name: CI Windows MSVC
+name: CI Windows MSVC-MSBuild
 
 on:
   pull_request:


asuessenbach:
Where have all the tests gone?
In VS I just have the `UniqueHandle` test, that's all.

M2-TE:
`UniqueHandle` is the only test that I have not converted to CTest, as it requires lots of dependency juggling and is not really standalone.

The tests are handled by the CTest executable, should be installed alongside CMake. Does VS not have CMake tools integration?
If you want to use it manually, cd into your build folder and do `ctest -j` and it should work.

M2-TE:
@asuessenbach if you are used to the workflow of having `add_subdirectory` for testing and want to keep it that way, let me know! I will redo the PR without CTest integration. I just thought to attempt the conversion as CTest seems pretty standard across CMake projects.

asuessenbach:
Having CTests would be cool.
But not having the test files available in the VS solution feels wrong. Can't we have both, CTest and files listed in VS?
If that's not (easily) possible, I'd prefer to not use CTest for now.

M2-TE:
I will try to find a solution to have both.
As for CTest in VS, you can view all the tests with the `Test Explorer` (Test > Test Explorer).
Under the same menu in `Test`, you can find `Run CTests for VulkanHppGenerator` if you just want to run all tests (it seems to not run them in parallel though).

M2-TE:
Most of the tests can now handle modules and `import std`. Remains to be seen if CI is able to handle it (currently only enabled for `windows` and `ubuntu-ext`).

All tests can be built as before via subdirectory (default behavior) or optionally be built via CTest instead as via `VULKAN_HPP_TESTS_CTEST`. Can you check if this works as expected for your workflow @asuessenbach?

Default setup of the `std` module is now handled in `vulkan.hpp` directly, so that the module no longer needs to include the `vulkan_hpp_macro.hpp` header. Module fails to build at current head due to lack of this definition, so give it a quick read if it's okay @sharadhr!

sharadhr:
> so that the module no longer needs to include the `vulkan_hpp_macro.hpp` header

It should; we still have `VULKAN_HPP_STRINGIFY` for the warnings! But this wasn't removed anyway...

Also, I'm not quite able to parse this statement:

> Module fails to build at current head due to lack of this definition

In which scenario does the module fail to build?

tests/StructureChain/StructureChain.cpp
```diff
@@ -115,25 +110,25 @@ int main( int /*argc*/, char ** /*argv*/ )
 
     // unlink a struct from a StructureChain
     sc7.unlink<vk::PhysicalDeviceMaintenance3Properties>();
-    assert( !sc7.isLinked<vk::PhysicalDeviceMaintenance3Properties>() );
+    void( !sc7.isLinked<vk::PhysicalDeviceMaintenance3Properties>() );
 
     // some invalid unlink calls
     // clang-format off
-    //sc7.unlink<vk::PhysicalDeviceMaintenance3Properties>();  // assertion fires on trying to unlink some already
+    //sc7.unlink<vk::PhysicalDeviceMaintenance3Properties>();  // voidion fires on trying to unlink some already
```
asuessenbach: voidion?

tests/StructureChain/StructureChain.cpp
```diff
@@ -115,25 +110,25 @@ int main( int /*argc*/, char ** /*argv*/ )
...
     sc7.relink<vk::PhysicalDeviceMaintenance3Properties>();
-    assert( sc7.isLinked<vk::PhysicalDeviceMaintenance3Properties>() );
+    void( sc7.isLinked<vk::PhysicalDeviceMaintenance3Properties>() );
 
     // invalid re-linking
     // clang-format off
     //sc7.relink<vk::PhysicalDeviceProperties2>();
     //sc1.relink<vk::PhysicalDeviceMaintenance3Properties>();
-    //sc1.relink<vk::PhysicalDeviceIDProperties>();  // assertion fires on trying to relink some structure that hasn't been unlinked
+    //sc1.relink<vk::PhysicalDeviceIDProperties>();  // voidion fires on trying to relink some structure that hasn't been unlinked
```
asuessenbach: voidion :)

M2-TE:
@sharadhr the error I get is:

```
In file included from Vulkan-Hpp/vulkan/vulkan.cppm:28:
Vulkan-Hpp/vulkan/vulkan.hpp:41:8: fatal error: module 'VULKAN_HPP_STD_MODULE' not found
   41 | import VULKAN_HPP_STD_MODULE;
```

but you are right, it **does** include the macro header (was I drunk?). It is once again an issue with the 
```cpp
defined( __cpp_modules ) && defined( __cpp_lib_modules )
```

so this could probably just be shortened to check for `VULKAN_HPP_CXX_MODULE` instead (defined at top of `vulkan.cppm`). What do you think?

@asuessenbach do you know of a way to test static dispatch in CI? It obviously won't be able to find Vulkan packages, so it likely isn't possible (same as before).

tests/StructureChain/StructureChain.cpp
M2-TE: woops..

snippets/includes.hpp
```diff
@@ -28,6 +28,9 @@
 #  include <cassert>
 #  include <cstring>
 #  include <cstdlib>
+#  if !defined( VULKAN_HPP_STD_MODULE )
```
sharadhr: > the error I get is... what do you think?

Let me explain how the macro works: `VULKAN_HPP_CXX_MODULE` tells the preprocessor that 'this header is now part of the named module, preprocess this as such', which means 'don't `#include` the standard library but `import` it instead'. So your change here and further below where you removed the `&& defined( __cpp_modules ) && defined( __cpp_lib_modules )` doesn't quite align with this. The named module should only ever be available if the entire stack (compiler, standard library, CMake, generator) supports it, and if it is hacked in anyway, then it will fail to compile (it does not for me, on MSVC 19.44).

Please revert both changes.

___

If you're compiling with Clang and want the standard library module, you need to have `-stdlib=libc++`.

snippets/MacrosHppTemplate.hpp
```diff
@@ -82,10 +82,6 @@
 #  define VULKAN_HPP_SUPPORT_SPAN
 #endif
 
-#if !defined( VULKAN_HPP_STD_MODULE ) && defined( __cpp_modules ) && defined( __cpp_lib_modules )
```
sharadhr: See [other comment](https://github.com/KhronosGroup/Vulkan-Hpp/pull/2292/files#r2432058468).

I put the `std.compat` definition in the macro header to appropriately separate concerns. It is a macro definition that is used by the named module. Plus, `vulkan.cppm` needs the macro file anyway—we use at least two other macros, enough that we should have the macro file up top anyway.

snippets/CppmTemplate.hpp
```diff
@@ -8,10 +8,9 @@ module;
 
 #if !defined( VULKAN_HPP_CXX_MODULE_EXPERIMENTAL_WARNING )
 #  define VULKAN_HPP_CXX_MODULE_EXPERIMENTAL_WARNING \
-  "The Vulkan-Hpp C++ named module is experimental. " \
-  "It is subject to change without prior notice.\n" \
-  "To silence this warning, define the VULKAN_HPP_CXX_MODULE_EXPERIMENTAL_WARNING macro.\n\n" \
-  "For feedback, go to: https://github.com/KhronosGroup/Vulkan-Hpp/issues"
+  "\n\tThe Vulkan-Hpp C++ named module is experimental. It is subject to change without prior notice.\n" \
```
sharadhr, line 11: Why the newline and tab stop? Also, I wanted to have two newlines for the feedback line so as to clearly separate out the new paragraph.

sharadhr (changes requested):

snippets/includes.hpp
M2-TE: Yes I'll revert those changes, as they don't make much sense.

We now don't really need to "check" for std module anymore though, so checking for `__cpp_modules` and `__cpp_lib_modules` can be replaced by a simple check for `VULKAN_HPP_CXX_MODULE`. As that one tells the header, as you put it, "this header is now part of the named module, preprocess this as such". It gets rid of the entire standard library headache entirely.

See the latest commit for what that would look like!

snippets/MacrosHppTemplate.hpp
M2-TE: Yes I agree, it will be put in the macro header again. The question now is just what it should look like.

.github/workflows/ci-ubuntu-ext.yml
```diff
@@ -121,16 +121,20 @@ jobs:
...
+              -D CMAKE_EXPERIMENTAL_CXX_IMPORT_STD=d0edc3af-4c50-42ea-a356-e2862fe7a444 \
               -D VULKAN_HPP_GENERATOR_BUILD=OFF                       \
               -D VULKAN_HPP_RUN_GENERATOR=OFF                         \
               -D VULKAN_HPP_SAMPLES_BUILD=ON                          \
               -D VULKAN_HPP_TESTS_BUILD=ON                            \
+              -D VULKAN_HPP_TESTS_CTEST=ON                            \
               -D VULKAN_HPP_DISPATCH_LOADER_DYNAMIC=ON                \
               -D VULKAN_HPP_BUILD_WITH_LOCAL_VULKAN_HPP=ON            \
               -D VULKAN_HPP_PRECOMPILE=OFF                            \
               -D CMAKE_CXX_COMPILER=clang++-${{matrix.clang_version}} \
```
sharadhr, line 139: I would also add `-D CMAKE_CXX_FLAGS=stdlib=libc++`, just for this case.

snippets/CppmTemplate.hpp
M2-TE: The compiler output made it a little unreadable before, where the empty newline made it seem like they were separate warnings and the start of the warning that we defined was hard to spot. The tabline was inserted to be in line with other compiler warning messages following the actual warning.

.github/workflows/ci-windows.yml
```diff
@@ -102,15 +102,19 @@ jobs:
             echo "================================================================================="
 
             cmake -B build --fresh                            \
```
sharadhr: I would add in `-G Ninja`. CMake's `module std` support only works with Ninja.

snippets/includes.hpp
sharadhr: > so checking for __cpp_modules and __cpp_lib_modules can be replaced by a simple check for VULKAN_HPP_CXX_MODULE

Not quite. These guards are our expectation that these macros need to have been defined by the toolchain—which proves a complete implementation—before we define the standard library module. If the user experiences a compile failure, they will follow the definition (or lack thereof) of `VULKAN_HPP_STD_MODULE`, find it in the macros file, and these guards will immediately tell them that their toolchain is in some way incomplete and at that point we can say 'sorry, we can't do anything here, ask your toolchain to fix it'.

We do it for the other feature tests like `<span>`, `<format>`, and `<expected>`; I feel it is appropriate to have the feature macro guard for this, too. 

As I also said:
> if it is hacked in anyway, then it will fail to compile

snippets/CppmTemplate.hpp
sharadhr: Fair enough.

snippets/includes.hpp
M2-TE: It compiles perfectly fine, as the support is definetely there. `module std` support is now a hard baseline requirement for the vulkan module, unlike the feature tests for `<span>` and such; those are optional.

Rather than blocking modules for compiler + standard library combinations that may theoretically fail (current), it may be better to only block when it **actually** doesn't compile (proposed).

.github/workflows/ci-windows.yml
M2-TE: Oh I see. I tested it on a VM earlier and perhaps VS automatically added the `-G Ninja` flag.
I would love to see CI output without it set first!

snippets/includes.hpp
sharadhr: > It compiles perfectly fine, as the support is definetely there

I don't quite follow—you mentioned earlier that you had a compile failure, but it now compiles correctly?

M2-TE: It compiles with the PR changes, which just check for `VULKAN_HPP_CXX_MODULE` to define the std module:
```cpp
// vulkan_macros.hpp
#if !defined( VULKAN_HPP_STD_MODULE ) && defined( VULKAN_HPP_CXX_MODULE )
#  define VULKAN_HPP_STD_MODULE std.compat
#endif
```

sharadhr: That's interesting. Were you not running into the problem described in #2301?

M2-TE: Hm, I worded that badly I believe. I do run into the linked issue (Clang + gcc standard lib on current main branch). That is precisely why I eased the std module definition guards like shown, now that std is a requirement and no longer optional. With that, it works

sharadhr: That's also interesting. CMake supposedly doesn't support [Clang with GNU libstdc++](https://gitlab.kitware.com/ben.boeckel/cmake/-/blob/cxxmodules-docs/Help/manual/cmake-cxxmodules.7.rst?ref_type=heads&plain=1#L82). 

As you mentioned, I'd like to see a CI run to see how it works...

.github/workflows/ci-windows.yml
sharadhr: On a local machine, configuring **without** `-G Ninja` doesn't print the 'developer warning' that you'd otherwise get when enabling the standard library experimentalism. The cache variable `VULKAN_HPP_BUILD_CXX_MODULE` is also unavailable, as are all the module targets.

snippets/includes.hpp
M2-TE: I cannot find it, but I could swear there was an update to CMake that eased this requirement.
But yes let us see, what CI tells us

M2-TE:
```
D:\a\Vulkan-Hpp\Vulkan-Hpp\tests\ExtensionInspection\ExtensionInspection.cpp(82,100): warning C4996: 'vk::EXTDebugReportExtensionName': The VK_EXT_debug_report extension has been deprecated by VK_EXT_debug_utils.
D:\a\Vulkan-Hpp\Vulkan-Hpp\tests\ExtensionInspection\ExtensionInspection.cpp(86,98): warning C4996: 'vk::AMDNegativeViewportHeightExtensionName': The VK_AMD_negative_viewport_height extension has been obsoleted by VK_KHR_maintenance1.
```

Warnings pop up for both `VK_EXT_debug_report` and `VK_AMD_negative_viewport_height`.
It tests for whether these are deprecated, but throws the warning regardless, simply because the enum is used, e.g.:

```cpp
  static_assert( vk::isObsoletedExtension( vk::AMDNegativeViewportHeightExtensionName ), "static assert test failed" );
  static_assert( vk::getExtensionObsoletedBy( vk::AMDNegativeViewportHeightExtensionName ) == vk::KHRMaintenance1ExtensionName,
                 "static assert test failed" );
```

The guard for these deprecation tests is also a bit peculiar, why the restrictions?

```cpp
#if ( 201907 <= __cpp_constexpr ) && ( !defined( __GNUC__ ) || ( 110400 < GCC_VERSION ) )
```

.github/workflows/ci-windows.yml
M2-TE: When you use `-G Ninja`, do you have to set a configuration to use in ctest via `-C Release`?
MSBuild is multi-config, so it is necessary for it. Not sure about Ninja on Windows.

.github/workflows/ci-ubuntu-ext.yml
M2-TE: Done. This is necessary, as runners do not have access to the gcc 15+ std library, just the newer clang compiler.

snippets/includes.hpp
M2-TE: It does indeed work if you set this up in a Ubuntu 24.04 VM, where you manually compile gcc 15 alongside the gcc std library. CI does not have access to this though (only the newer clang compiler version), so it's not really test-able yet in CI.

M2-TE:
Current state of my descent into CI debugging insanity:

- Clang++-20: CMake does not provide the `[[ TARGET __CMAKE::CXX23 ]]` target.
- Clang++-19: CMake does provide the  `[[ TARGET __CMAKE::CXX23 ]]` target, but has issues with the standard library (using `-stblib=libc++`) as seen here: https://github.com/M2-TE/Vulkan-Hpp/actions/runs/18653373179/job/53176398437. Seems to affect compilation of `utils.hpp`.

I believe this is because it defaults to an older standard library version, as clang is installed in the runners. Overriding the specific libc++ version seems like a pain: https://releases.llvm.org/14.0.0/projects/libcxx/docs/UsingLibcxx.html

Identical setup in ubuntu:24.04 docker setup works just fine with modules, so not exactly sure what is different.

M2-TE:
I think it's best to postpone the CI testing of modules. I will work on this as a separate PR, so things don't get too messy. All the tests will now work with a properly set up toolchain that can handle modules and `import std`.

Some other things added:
- Properly set the build type on windows via `--config $BUILD_TYPE`
- Fixed the extension inspection test on windows by ignoring the deprecation warning via pragma

With this, I would say the PR is pretty much complete. The concerns discussed with @sharadhr related to CI testing should be handled in a separate PR imo. On windows, building with Ninja seems to still have an issue or two in CI anyways.

sharadhr:
Side note: I realised `#warning` is [in the C++23 standard](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2022/p2437r1.pdf), but MSVC hasn't implemented it yet (sigh).

M2-TE:
Brought the PR up-to-date and managed to get module testing working.
C++Modules are being tested via ~Ninja+Clang on Windows~ and Ninja+Clang on ubuntu-ext. Testing on MSVC would required a separate runner that uses Ninja (this PR already ballooned enough as is).

M2-TE:
PR is ready for review, so please let me know what you think @sharadhr.

sharadhr:
A cursory glance shows no glaring issues. That singular build failure should probably be resolved by rebasing atop `main`. However I am surprised it did not fail on any other platform, especially Windows + MSVC, which _does_ support `import std`. There might be a missing test case scenario; see [this action failure](https://github.com/YaaZ/VulkanMemoryAllocator-Hpp/actions/runs/19757991963/job/56613233167?pr=69). (I've been thinking about using that CI setup, which does seem very comprehensive in terms of platform coverage, and yet quite simple in its implementation). 

I do recall there was discussion of splitting up this PR: one to set up CTest, and another to set up the module tests.

M2-TE:
Well at least now we know that the module tests work ;)

But yes I remember us talking about splitting this up. The changes here are quite interlinked (CTest and Modules), so if it is still desired to split it into two separate PRs, I would simply exclude the module-related things from CI. Is that acceptable or should I tear these apart on a deeper level?

> I am surprised it did not fail on any other platform

I only "activated" the ubuntu-ext one so far. Windows Clang has too old of a CMake version, same for MSVC. Since you mentioned wanting to adopt those other CI workflows, I simply focused on having one single working test to improve the current.. lack of module tests.

I do quite like those workflows too, so I would vote to adopt those as well! Hopefully after merging this though, upkeep is a bit of work with the merge conflicts..

sharadhr:
> Windows Clang has too old of a CMake version, same for MSVC

This is surprising! Visual Studio bundles CMake 4.1. Or we can always install the upstream CMake with the same `get-cmake` workflow.

M2-TE:
Yea I was suprised as well. Check the GitHub hosted runners: [Windows 2022](https://github.com/actions/runner-images/blob/main/images/windows/Windows2022-Readme.md) and [Windows 2025](https://github.com/actions/runner-images/blob/main/images/windows/Windows2025-Readme.md).

We can probably just use a baseline like
```yml
      - name: Setup CMake and Ninja
        if: matrix.env.modules
        uses: lukka/get-cmake@latest
        with:
          cmakeVersion: 4.1.2 # (or latest?)
          ninjaVersion: latest
```
Should be a better solution than the `get-cmake` thing I've done before.

sharadhr:
>  I would simply exclude the module-related things from CI. Is that acceptable or should I tear these apart on a deeper level?

I think just having the module stuff separate would be sufficient. Although I am decidedly not the final arbiter, and maybe @asuessenbach should also input.

asuessenbach:
It seems, you're running this PR on a slightly outdated Vulkan-Headers version.
Would be great, if you'd rebase it accordingly.

asuessenbach:
> > I would simply exclude the module-related things from CI. Is that acceptable or should I tear these apart on a deeper level?
>
> I think just having the module stuff separate would be sufficient. Although I am decidedly not the final arbiter, and maybe @asuessenbach should also input.

Agreed, "simply excluding the module-related things from CI" is probably the right thing to do.

M2-TE:
> Agreed, "simply excluding the module-related things from CI" is probably the right thing to do.

Done!

> It seems, you're running this PR on a slightly outdated Vulkan-Headers version

It should be on the newest version, as I've kept it up to date with `main` and it is currently at `2fa2034`. Could you check it again or tell me which commit hash it should be using @asuessenbach?

asuessenbach:
> It should be on the newest version

You're right. Don't know what I've seen before. Sorry, please.

M2-TE:
I had accidentally removed the ctest calls for Windows CI before.. sorry about that!

tests/CMakeLists.txt
```diff
@@ -12,48 +12,77 @@
...
-if( CMAKE_CXX_STANDARD GREATER_EQUAL 20 )
-	add_subdirectory( ExtensionInspection )
+# without ctest, can only run these tests with VULKAN_HPP_DISPATCH_LOADER_DYNAMIC disabled
+# CI cannot run these tests either way, as it does not have access to the Vulkan package
+if( NOT VULKAN_HPP_DISPATCH_LOADER_DYNAMIC )
+	vulkan_hpp__setup_test( NAME DeviceFunctions )
+	vulkan_hpp__setup_test( NAME DispatchLoaderStatic )
+endif()
+# without ctest, can only run these tests with VULKAN_HPP_DISPATCH_LOADER_DYNAMIC enabled
+if( NOT( NOT VULKAN_HPP_TESTS_CTEST AND NOT VULKAN_HPP_DISPATCH_LOADER_DYNAMIC ))
```
asuessenbach: Isn't this check equivalent to
`if( VULKAN_HPP_TESTS_CTEST OR VULKAN_HPP_DISPATCH_LOADER_DYNAMIC)`

tests/DesignatedInitializers/DesignatedInitializers.cpp
```diff
@@ -62,27 +62,9 @@ int main( int /*argc*/, char ** /*argv*/ )
...
-  vk::ApplicationInfo ai1( appName );
-  vk::ApplicationInfo ai2( appName, appVersion );
-  vk::ApplicationInfo ai3( appName, appVersion, engineName );
-  vk::ApplicationInfo ai4( appName, appVersion, engineName, engineVersion );
-  vk::ApplicationInfo ai5( appName, appVersion, engineName, engineVersion, vk::ApiVersion12 );
-
-  // a structure in namespace vk:: can be copied from the corresponding vulkan C-struct
-  VkApplicationInfo   vai;
-  vk::ApplicationInfo ai6( vai );
-
```
asuessenbach, line 77: Where are those tests gone?

tests/CMakeLists.txt
```diff
@@ -12,48 +12,77 @@
...
-	endif()
-endif()
-add_subdirectory( CppType )
+vulkan_hpp__setup_test( NAME ArrayProxy )
+vulkan_hpp__setup_test( NAME ArrayProxyNoTemporaries )
+vulkan_hpp__setup_test( NAME ArrayWrapper )
+vulkan_hpp__setup_test( NAME CppType )
 if( CMAKE_CXX_STANDARD GREATER_EQUAL 20 )
-	add_subdirectory( DesignatedInitializers )
+	vulkan_hpp__setup_test( NAME DesignatedInitializers )
```
asuessenbach, line 20: Maybe the name of this test is not as good as it should... Anyways, it also contained tests with C++ less than 20.

tests/CMakeLists.txt
M2-TE: Indeed it is! Let me change it to that. Not sure how that boolean sausage came to be.

tests/DesignatedInitializers/DesignatedInitializers.cpp
M2-TE: I deemed these to be somewhat pointless, since this test will always have `VULKAN_HPP_NO_STRUCT_CONSTRUCTORS` defined. This will never get compiled. These sorts of constructors get tested in the other tests aplenty!

tests/CMakeLists.txt
M2-TE: If you are referring to the DesignatedInitializers test, aren't designated initializers a C++20 feature? The name is also quite fitting, or is this test supposed to test other things besides?

M2-TE:
> In all the tests, you've changed the asserts to void

The asserts do not get compiled in release mode, which makes some tests have a much more narrow coverage in said release mode. The `void` makes sure they get evaluated. This includes things like testing operators and constructors.

tests/DesignatedInitializers/DesignatedInitializers.cpp
asuessenbach: Indeed, you're right.

tests/CMakeLists.txt
asuessenbach: Yes, it's the DesignatedInitializers test. And it did include some tests that (as noted in the above comment) were never used.
That is: it's ok to have it with C++20 and above only.

asuessenbach:
> The `void` makes sure they get evaluated.

That's right. But it doesn't fire on you when the check fails (in debug mode).

So, maybe, we'd need some real runtime check other than just `assert`?

M2-TE:
Are the test's compiled binaries intended to be executed aswell? I've only ever seen them being compiled for testing. In the case of running them, a real runtime check would indeed be better than assert or void. For compile-time checks, `static_assert` would still be evaluated properly even in release mode.

asuessenbach:
> Are the test's compiled binaries intended to be executed as well?

Well, some of them (those with `assert`s in ;)) are meant to be executed as well. At least every now and then, I'm running them. In debug mode, usually. And they need to be runtime checks.

sharadhr:
Could we also add the `CMAKE_EXPERIMENTAL_CXX_IMPORT_STD` variable to the Windows/MSVC tests?

M2-TE:
@asuessenbach I'll write an assert function that may `throw std::runtime_exception` and can be used in place of assert in both debug and release mode. This function can be added to the relevant tests where you want asserts to happen.

sharadhr:
On a side note, it might be worthwhile to move to a proper testing mechanism like GoogleTest in future, that will allow us to express concepts such as 'can/cannot accept temporary' which cannot necessarily be written conveniently with just `assert`.

M2-TE:
@sharadhr should I not leave `CMAKE_EXPERIMENTAL_CXX_IMPORT_STD` out for now, since the module portion should not be part of the CI for this PR?

@asuessenbach I've added a `release_assert` function to the tests. These are simply copied as to not have any dependencies per test and therefore also kept simple. Maybe we should expand the tests in the future to use CMake's [try_run](https://cmake.org/cmake/help/latest/command/try_run.html) to also run these tests.

tests/ArrayProxy/ArrayProxy.cpp
```diff
@@ -35,6 +35,11 @@ import vulkan;
 #  include <vulkan/vulkan.hpp>
 #endif
 
+template<typename T> void release_assert( const T &condition )
```
asuessenbach: Why do you make this a templated function? I think, the condition is always a boolean.

And instead of having that function in every test, we could add a little header with it.

And, finally, I had envisioned it to be a bit more verbose, like
```
#define CHECK( condition )                                                                                 \
  do                                                                                                       \
  {                                                                                                        \
    if ( !( condition ) )                                                                                  \
    {                                                                                                      \
      std::cout << "ERROR: (" << #condition << ") failed in " << __FILE__ << ":" << __LINE__ << std::endl; \
      std::exit( -1 );                                                                                     \
    }                                                                                                      \
  } while ( false )

```

M2-TE: It is templated since e.g. Flags test has some `vk::MemoryHeapFlags` that weren't getting accepted otherwise.

I had decided against a header to leave all tests standalone, but that does not make much sense in retrospect, so I'll create one.

As for the assert function, I thought `std::abort` printed the usual __FILE__ etc strings. Since that is not the case, I will create a macro that mirrors `assert` as much as possible, similar to what you have. Here's what I would use:
```cpp
#define release_assert( condition ) \
  if ( !( x ) ) \
  { \
    std::cout << __FILE__ << ':' << __LINE__ << ": Assertion \'" << #condition << "\' failed."<< std::endl; \
    std::abort(); \
  }
```

Or do you prefer your proposed macro?

asuessenbach: No, that macro looks good as well.
Maybe just wrap it with `do {...} while ( false )` to catch missing `;`s

M2-TE: Oh so that's what that is for, good to know. Will do!

sharadhr:
> should I not leave CMAKE_EXPERIMENTAL_CXX_IMPORT_STD out for now, since the module portion should not be part of the CI for this PR?

Right, yes. That should be the case. I didn't notice your latest updates where you had removed it from all of the other CI runs as well.

tests/ArrayProxy/ArrayProxy.cpp
asuessenbach: Yeah, looks good now.
Please just add the standard copyright and license message to the top of this new `test_macros.hpp`.

And, as all that `assert`/`release_assert` stuff essentially is independent of the actual goal of this PR, it would be great if you would cut that out into a separate PR.
But if you think that's too much effort, it can stay here.

M2-TE: No big deal, I've reverted it to what it was before.

asuessenbach: Well, but before, there were `assert`s, no `void`s?

M2-TE: Indeed, my mistake. I've re-added `cassert` headers back as well. Will be cut out with the release assertion PR.

asuessenbach:
Alright now. I both of you, @M2-TE  and @sharadhr are satisfied with this PR, I'd be happy to merge it.

tests/DispatchLoaderStatic/DispatchLoaderStatic.cpp
```diff
@@ -15,12 +15,16 @@
...
 //                     Compile test on DispatchLoaderStatic functions
 
-#undef VULKAN_HPP_DISPATCH_LOADER_DYNAMIC
-#define VULKAN_HPP_DISPATCH_LOADER_DYNAMIC 0
 
-#include <iostream>
-#include <map>
-#include <vulkan/vulkan.hpp>
+#ifdef VULKAN_HPP_USE_CXX_MODULE
+import vulkan_hpp;
```
sharadhr:
```suggestion
import vulkan;
```

tests/FunctionCalls/FunctionCalls.cpp
```diff
@@ -27,9 +27,20 @@
 // unknown compiler... just ignore the warnings for yourselves ;)
 #endif
 
-#include <vulkan/vulkan.hpp>
 
-VULKAN_HPP_DEFAULT_DISPATCH_LOADER_DYNAMIC_STORAGE
+#ifdef VULKAN_HPP_USE_CXX_MODULE
+#  include <vulkan/vulkan.h>
+import vulkan_hpp;
```
sharadhr:
```suggestion
import vulkan;
```

tests/FunctionCallsRAII/FunctionCallsRAII.cpp
```diff
@@ -27,7 +27,13 @@
 // unknown compiler... just ignore the warnings for yourselves ;)
 #endif
 
-#include <vulkan/vulkan_raii.hpp>
+
+#ifdef VULKAN_HPP_USE_CXX_MODULE
+#  include <cstdint>
+import vulkan_hpp;
```
sharadhr:
```suggestion
import vulkan;
```

tests/Video/Video.cpp
```diff
@@ -26,7 +26,13 @@
 // unknown compiler... just ignore the warnings for yourselves ;)
 #endif
 
-#include <vulkan/vulkan_video.hpp>
+
+#ifdef VULKAN_HPP_USE_CXX_MODULE
+import vulkan_video_hpp;
```
sharadhr:
```suggestion
import vulkan;
```

M2-TE:
@sharadhr are you sure those should to be separated out? It is safe behind that `VULKAN_HPP_USE_CXX_MODULE` flag for normal testing.

Quite a lot of work was done in this PR that fixed and/or made module testing possible (CI thing aside) even in the CMake and the tests themselves. It was the motivation behind using CTest to begin with and were designed together.

sharadhr:
Hmm, if they're that closely tied together, then maybe we might as well _enable_ and run those tests. It would catch any typos, bugs, etc here. Naturally this would mean putting the experimental CMake import std variable back in.

M2-TE:
I would still opt to implement module testing in CI as a separate PR, since the stuff here only changed the tests themselves and not the functionality of the module.

Enabling CI tests would be sort of intrusive as well, since e.g. the Windows runners would need their base image to be adjusted to get access to the newest CMake.

sharadhr:
That feels a bit incongruent.

The module testing changes are so closely tied that they can't be separated out, but they're not so close that they can be enabled in another PR, seems a very fine line. 

If the work here was done explicitly to allow testing the module then I feel we should also enable the tests, it doesn't make sense to split those up. 

I'd really rather get the tests in CI sooner than later—it's an extra ~5 lines to set up the experimental CMake variable and pass it into the command-line invocations.

M2-TE:
In that case, I've copied the CMake+Ninja base from the ubuntu-ext runner over to the windows-clang runner.
Those two should now be able to test modules.

.github/workflows/ci-windows-msvc.yml
```diff
@@ -89,9 +89,11 @@ jobs:
             -D VULKAN_HPP_PRECOMPILE=OFF                    \
             -D CMAKE_CXX_STANDARD=$CXX_STANDARD
```
sharadhr, line 90:
```suggestion
            -D CMAKE_CXX_STANDARD=$CXX_STANDARD
            -D CMAKE_EXPERIMENTAL_CXX_IMPORT_STD=d0edc3af-4c50-42ea-a356-e2862fe7a444
            -D VULKAN_HPP_BUILD_CXX_MODULE=$CXX_MODULES
```

.github/workflows/ci-windows-msvc.yml
sharadhr, line 78:
```suggestion
        for CXX_STANDARD in 11 14 17 20 23; do
          CXX_MODULES=ON
            if [ $CXX_STANDARD -lt 23 ]; then
              CXX_MODULES=OFF
            fi
```

M2-TE:
Before I push this onto CI, could you locally test whether this works without using Ninja on Windows? The MSVC runners use MSBuild atm.

> Hmm, [CMake doesn't like Clang-CL](https://gitlab.kitware.com/cmake/cmake/-/issues/25731) for modules.

Indeed, it does not run the module tests yet:
https://github.com/M2-TE/Vulkan-Hpp/actions/runs/20129925080/job/57768605787?pr=13

sharadhr:
Running locally on MSVC. Note that CMake doesn't support MSBuild for `import std;`, and only Ninja is allowed. 

Full command-line output follows below. 

### Configuration
```plaintext
> cmake -B build --fresh -GNinja -D VULKAN_HPP_GENERATOR_BUILD=OFF -D VULKAN_HPP_RUN_GENERATOR=OFF -D VULKAN_HPP_SAMPLES_BUILD=ON -D VULKAN_HPP_TESTS_BUILD=ON -D VULKAN_HPP_DISPATCH_LOADER_DYNAMIC=ON -D VULKAN_HPP_BUILD_WITH_LOCAL_VULKAN_HPP=ON -D VULKAN_HPP_PRECOMPILE=OFF -D CMAKE_CXX_STANDARD=23 -D CMAKE_EXPERIMENTAL_CXX_IMPORT_STD="d0edc3af-4c50-42ea-a356-e2862fe7a444" -D VULKAN_HPP_BUILD_CXX_MODULE=ON -D VULKAN_HPP_TESTS_CTEST=ON
-- The CXX compiler identification is MSVC 19.50.35720.0
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: C:/Program Files/Microsoft Visual Studio/18/Enterprise/VC/Tools/MSVC/14.50.35717/bin/Hostx64/x64/cl.exe - skipped
-- Detecting CXX compile features
CMake Warning (dev) at C:/Program Files/CMake/share/cmake-4.2/Modules/Compiler/CMakeCommonCompilerMacros.cmake:248 (cmake_language):
  CMake's support for `import std;` in C++23 and newer is experimental.  It
  is meant only for experimentation and feedback to CMake developers.
Call Stack (most recent call first):
  C:/Program Files/CMake/share/cmake-4.2/Modules/CMakeDetermineCompilerSupport.cmake:113 (cmake_create_cxx_import_std)
  C:/Program Files/CMake/share/cmake-4.2/Modules/CMakeTestCXXCompiler.cmake:83 (CMAKE_DETERMINE_COMPILER_SUPPORT)
  CMakeLists.txt:30 (project)
This warning is for project developers.  Use -Wno-dev to suppress it.

-- Detecting CXX compile features - done
CMake Deprecation Warning at glm/CMakeLists.txt:1 (cmake_minimum_required):
  Compatibility with CMake < 3.10 will be removed from a future version of
  CMake.

  Update the VERSION argument <min> value.  Or, use the <min>...<max> syntax
  to tell CMake that the project requires at least <min> but has been updated
  to work with policies introduced by <max> or earlier.


CMake Deprecation Warning at glm/CMakeLists.txt:2 (cmake_policy):
  Compatibility with CMake < 3.10 will be removed from a future version of
  CMake.

  Update the VERSION argument <min> value.  Or, use the <min>...<max> syntax
  to tell CMake that the project requires at least <min> but has been updated
  to work with policies introduced by <max> or earlier.


-- GLM: Version 0.9.9.9
-- The C compiler identification is MSVC 19.50.35720.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: C:/Program Files/Microsoft Visual Studio/18/Enterprise/VC/Tools/MSVC/14.50.35717/bin/Hostx64/x64/cl.exe - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD - Failed
-- Looking for pthread_create in pthreads
-- Looking for pthread_create in pthreads - not found
-- Looking for pthread_create in pthread
-- Looking for pthread_create in pthread - not found
-- Found Threads: TRUE
-- Could NOT find Doxygen (missing: DOXYGEN_EXECUTABLE)
-- Including Win32 support
-- spirv-tools not linked - illegal SPIRV may be generated for HLSL
-- Found Python3: C:/Program Files/Python312/python3.exe (found version "3.12.10") found components: Interpreter
-- Configuring done (3.6s)
-- Generating done (0.6s)
-- Build files have been written to: R:/Vulkan-Hpp/build
```

### Build
```plaintext
> cmake --build build --config Release --clean-first
[1/1] Cleaning all built files...
Cleaning... 292 files.
[1/302] Scanning "C:\Program Files\Microsoft Visual Studio\18\Enterprise\VC\Tools\MSVC\14.50.35717\modules\std.compat.ixx" for CXX dependencies
std.compat.ixx
[2/302] Scanning "C:\Program Files\Microsoft Visual Studio\18\Enterprise\VC\Tools\MSVC\14.50.35717\modules\std.ixx" for CXX dependencies
std.ixx
[3/302] Scanning R:\Vulkan-Hpp\vulkan\vulkan_video.cppm for CXX dependencies
vulkan_video.cppm
[13/302] Scanning R:\Vulkan-Hpp\vulkan\vulkan.cppm for CXX dependencies
vulkan.cppm
[43/302] Building CXX object CMakeFiles\__cmake_cxx23.dir\C_\Program_File...Visual_Studio\18\Enterprise\VC\Tools\MSVC\14.50.35717\modules\std.ixx.obj
std.ixx
[44/302] Building CXX object CMakeFiles\__cmake_cxx23.dir\C_\Program_File...Studio\18\Enterprise\VC\Tools\MSVC\14.50.35717\modules\std.compat.ixx.obj
std.compat.ixx
[74/302] Building CXX object CMakeFiles\VulkanHppModule.dir\vulkan\vulkan_video.cppm.obj
vulkan_video.cppm
R:\Vulkan-Hpp\vulkan\vulkan_video.cppm(19): warning:
        The Vulkan-Hpp C++ named module is experimental. It is subject to change without prior notice.
        To silence this warning, define the VULKAN_HPP_CXX_MODULE_EXPERIMENTAL_WARNING macro.
        For feedback, go to: https://github.com/KhronosGroup/Vulkan-Hpp/issues
[142/302] Building CXX object CMakeFiles\VulkanHppModule.dir\vulkan\vulkan.cppm.obj
vulkan.cppm
R:\Vulkan-Hpp\vulkan\vulkan.cppm(20): warning:
        The Vulkan-Hpp C++ named module is experimental. It is subject to change without prior notice.
        To silence this warning, define the VULKAN_HPP_CXX_MODULE_EXPERIMENTAL_WARNING macro.
        For feedback, go to: https://github.com/KhronosGroup/Vulkan-Hpp/issues
[302/302] Linking CXX executable RAII_Samples\Template\RAII_Template.exe
```

### Test

```plaintext
> ctest -j --output-on-failure --test-dir build
Test project R:/Vulkan-Hpp/build
      Start 39: MODULE_StructureChain
      Start 30: MODULE_EnableBetaExtensions
      Start 29: MODULE_DispatchLoaderDynamic
      Start 31: MODULE_Flags
      Start 37: MODULE_NoExceptions
      Start 27: MODULE_CppType
      Start 25: MODULE_StridedArrayProxy
      Start 33: MODULE_FunctionCallsRAII
      Start 38: MODULE_NoExceptionsRAII
      Start 28: MODULE_DesignatedInitializers
      Start 35: MODULE_Hash
      Start 36: MODULE_NoDefaultDispatcher
      Start 26: MODULE_ArrayProxyNoTemporaries
      Start 24: MODULE_ArrayProxy
      Start 32: MODULE_FunctionCalls
      Start 34: MODULE_HandlesMoveExchange
 1/41 Test #27: MODULE_CppType .............................   Passed   42.19 sec
      Start 40: MODULE_UniqueHandleDefaultArguments
 2/41 Test #24: MODULE_ArrayProxy ..........................   Passed   42.18 sec
      Start 41: MODULE_Video
 3/41 Test #39: MODULE_StructureChain ......................***Failed   42.39 sec
Internal cmake changing into directory: R:/Vulkan-Hpp/build/tests/tests/MODULE_StructureChain
======== CMake output     ======
Configuring done (0.0s)
CMake Warning (dev) in CMakeLists.txt:
  CMake's support for `import std;` in C++23 and newer is experimental.  It
  is meant only for experimentation and feedback to CMake developers.
This warning is for project developers.  Use -Wno-dev to suppress it.

Generating done (0.1s)
Build files have been written to: R:/Vulkan-Hpp/build/tests/tests/MODULE_StructureChain
======== End CMake output ======
Change Dir: 'R:/Vulkan-Hpp/build/tests/tests/MODULE_StructureChain'

Run Clean Command: C:/PROGRA~1/MIB055~1/18/ENTERP~1/Common7/IDE/COMMON~1/MICROS~1/CMake/Ninja/ninja.exe clean
[1/1] Cleaning all built files...
Cleaning... 26 files.

Run Build Command(s): C:/PROGRA~1/MIB055~1/18/ENTERP~1/Common7/IDE/COMMON~1/MICROS~1/CMake/Ninja/ninja.exe
[1/16] Scanning "C:\Program Files\Microsoft Visual Studio\18\Enterprise\VC\Tools\MSVC\14.50.35717\modules\std.compat.ixx" for CXX dependencies
std.compat.ixx
[2/16] Scanning R:\Vulkan-Hpp\tests\StructureChain\StructureChain.cpp for CXX dependencies
[3/16] Scanning R:\Vulkan-Hpp\vulkan\vulkan_video.cppm for CXX dependencies
vulkan_video.cppm
[4/16] Scanning R:\Vulkan-Hpp\vulkan\vulkan.cppm for CXX dependencies
vulkan.cppm
[5/16] Scanning "C:\Program Files\Microsoft Visual Studio\18\Enterprise\VC\Tools\MSVC\14.50.35717\modules\std.ixx" for CXX dependencies
std.ixx
[6/16] Generating CXX dyndep file CMakeFiles\__cmake_cxx23.dir\CXX.dd
[7/16] Generating CXX dyndep file CMakeFiles\VulkanHppModule.dir\CXX.dd
[8/16] Generating CXX dyndep file CMakeFiles\MODULE_StructureChain.dir\CXX.dd
[9/16] Building CXX object CMakeFiles\__cmake_cxx23.dir\C_\Program_Files\Microsoft_Visual_Studio\18\Enterprise\VC\Tools\MSVC\14.50.35717\modules\std.ixx.obj
std.ixx
[10/16] Building CXX object CMakeFiles\__cmake_cxx23.dir\C_\Program_Files\Microsoft_Visual_Studio\18\Enterprise\VC\Tools\MSVC\14.50.35717\modules\std.compat.ixx.obj
std.compat.ixx
[11/16] Linking CXX static library __cmake_cxx23.lib
[12/16] Building CXX object CMakeFiles\VulkanHppModule.dir\R_\Vulkan-Hpp\vulkan\vulkan_video.cppm.obj
vulkan_video.cppm
R:\Vulkan-Hpp\vulkan\vulkan_video.cppm(19): warning:
        The Vulkan-Hpp C++ named module is experimental. It is subject to change without prior notice.
        To silence this warning, define the VULKAN_HPP_CXX_MODULE_EXPERIMENTAL_WARNING macro.
        For feedback, go to: https://github.com/KhronosGroup/Vulkan-Hpp/issues
[13/16] Building CXX object CMakeFiles\VulkanHppModule.dir\R_\Vulkan-Hpp\vulkan\vulkan.cppm.obj
vulkan.cppm
R:\Vulkan-Hpp\vulkan\vulkan.cppm(20): warning:
        The Vulkan-Hpp C++ named module is experimental. It is subject to change without prior notice.
        To silence this warning, define the VULKAN_HPP_CXX_MODULE_EXPERIMENTAL_WARNING macro.
        For feedback, go to: https://github.com/KhronosGroup/Vulkan-Hpp/issues
[14/16] Linking CXX static library VulkanHppModule.lib
[15/16] Building CXX object CMakeFiles\MODULE_StructureChain.dir\StructureChain.cpp.obj
FAILED: CMakeFiles/MODULE_StructureChain.dir/StructureChain.cpp.obj
C:\PROGRA~1\MIB055~1\18\ENTERP~1\VC\Tools\MSVC\1450~1.357\bin\Hostx64\x64\cl.exe  /nologo /TP -DVK_NO_PROTOTYPES -DVULKAN_HPP_DISPATCH_LOADER_DYNAMIC -DVULKAN_HPP_USE_CXX_MODULE -IR:\Vulkan-Hpp -IR:\Vulkan-Hpp\Vulkan-Headers\include /DWIN32 /D_WINDOWS /EHsc /MDd /Zi /Ob0 /Od /RTC1 -std:c++latest /W4 /WX /showIncludes @CMakeFiles\MODULE_StructureChain.dir\StructureChain.cpp.obj.modmap /FoCMakeFiles\MODULE_StructureChain.dir\StructureChain.cpp.obj /FdCMakeFiles\MODULE_StructureChain.dir\ /FS -c R:\Vulkan-Hpp\tests\StructureChain\StructureChain.cpp
R:\Vulkan-Hpp\tests\StructureChain\StructureChain.cpp(91): error C3643: 'vk::StructureChain<vk::PhysicalDeviceProperties2,vk::PhysicalDeviceIDProperties>': cannot decompose type with non-static data-members in both 'std::tuple<vk::PhysicalDeviceProperties2,vk::PhysicalDeviceIDProperties>' and 'std::tuple<vk::PhysicalDeviceIDProperties>'
C:\Program Files\Microsoft Visual Studio\18\Enterprise\VC\Tools\MSVC\14.50.35717\include\tuple(823): note: see declaration of 'std::tuple<vk::PhysicalDeviceProperties2,vk::PhysicalDeviceIDProperties>::_Myfirst'
C:\Program Files\Microsoft Visual Studio\18\Enterprise\VC\Tools\MSVC\14.50.35717\include\tuple(823): note: see declaration of 'std::tuple<vk::PhysicalDeviceIDProperties>::_Myfirst'
R:\Vulkan-Hpp\tests\StructureChain\StructureChain.cpp(91): error C3448: the number of identifiers must match the number of array elements or members in a structured binding declaration
R:\Vulkan-Hpp\tests\StructureChain\StructureChain.cpp(92): error C3643: 'vk::StructureChain<vk::PhysicalDeviceProperties2,vk::PhysicalDeviceMaintenance3Properties>': cannot decompose type with non-static data-members in both 'std::tuple<vk::PhysicalDeviceProperties2,vk::PhysicalDeviceMaintenance3Properties>' and 'std::tuple<vk::PhysicalDeviceMaintenance3Properties>'
C:\Program Files\Microsoft Visual Studio\18\Enterprise\VC\Tools\MSVC\14.50.35717\include\tuple(823): note: see declaration of 'std::tuple<vk::PhysicalDeviceProperties2,vk::PhysicalDeviceMaintenance3Properties>::_Myfirst'
C:\Program Files\Microsoft Visual Studio\18\Enterprise\VC\Tools\MSVC\14.50.35717\include\tuple(823): note: see declaration of 'std::tuple<vk::PhysicalDeviceMaintenance3Properties>::_Myfirst'
R:\Vulkan-Hpp\tests\StructureChain\StructureChain.cpp(92): error C3448: the number of identifiers must match the number of array elements or members in a structured binding declaration
ninja: build stopped: subcommand failed.


      Start  6: ExtensionInspection
 4/41 Test #25: MODULE_StridedArrayProxy ...................   Passed   42.38 sec
      Start 15: Handles
 5/41 Test #30: MODULE_EnableBetaExtensions ................   Passed   42.60 sec
      Start  4: CppType
 6/41 Test #37: MODULE_NoExceptions ........................   Passed   42.74 sec
      Start  8: DispatchLoaderDynamicSharedLibraryClient
 7/41 Test #38: MODULE_NoExceptionsRAII ....................   Passed   43.59 sec
      Start 14: FunctionCallsRAII
 8/41 Test #36: MODULE_NoDefaultDispatcher .................   Passed   43.60 sec
 9/41 Test #34: MODULE_HandlesMoveExchange .................   Passed   43.59 sec
      Start 17: Hash
      Start 19: NoExceptionsRAII
10/41 Test #28: MODULE_DesignatedInitializers ..............   Passed   43.64 sec
      Start 18: NoDefaultDispatcher
11/41 Test #32: MODULE_FunctionCalls .......................   Passed   44.00 sec
      Start  2: ArrayProxyNoTemporaries
12/41 Test #29: MODULE_DispatchLoaderDynamic ...............   Passed   44.23 sec
      Start 11: EnableBetaExtensions
13/41 Test #26: MODULE_ArrayProxyNoTemporaries .............   Passed   44.32 sec
      Start 16: HandlesMoveExchange
14/41 Test #31: MODULE_Flags ...............................   Passed   44.39 sec
      Start  5: DesignatedInitializers
15/41 Test #33: MODULE_FunctionCallsRAII ...................   Passed   44.54 sec
      Start 13: FormatTraits
16/41 Test #35: MODULE_Hash ................................   Passed   47.52 sec
      Start 21: StructureChain
17/41 Test  #2: ArrayProxyNoTemporaries ....................   Passed   10.78 sec
      Start 12: Flags
18/41 Test  #4: CppType ....................................   Passed   12.32 sec
      Start 10: FunctionCalls
19/41 Test  #5: DesignatedInitializers .....................   Passed   10.65 sec
      Start  9: NoExceptions
20/41 Test #11: EnableBetaExtensions .......................   Passed   10.79 sec
      Start  7: DispatchLoaderDynamic
21/41 Test #15: Handles ....................................   Passed   12.75 sec
      Start  1: ArrayProxy
22/41 Test #19: NoExceptionsRAII ...........................   Passed   11.84 sec
      Start  3: ArrayWrapper
23/41 Test #16: HandlesMoveExchange ........................   Passed   11.24 sec
      Start 23: Video
24/41 Test  #8: DispatchLoaderDynamicSharedLibraryClient ...   Passed   12.88 sec
      Start 20: StridedArrayProxy
25/41 Test #17: Hash .......................................   Passed   12.08 sec
      Start 22: UniqueHandleDefaultArguments
26/41 Test #13: FormatTraits ...............................   Passed   11.13 sec
27/41 Test #18: NoDefaultDispatcher ........................   Passed   12.08 sec
28/41 Test #14: FunctionCallsRAII ..........................   Passed   12.20 sec
29/41 Test  #6: ExtensionInspection ........................   Passed   16.84 sec
30/41 Test #21: StructureChain .............................   Passed   12.06 sec
31/41 Test #12: Flags ......................................   Passed    8.78 sec
32/41 Test  #9: NoExceptions ...............................   Passed    8.84 sec
33/41 Test  #1: ArrayProxy .................................   Passed    8.82 sec
34/41 Test  #7: DispatchLoaderDynamic ......................   Passed    8.92 sec
35/41 Test #10: FunctionCalls ..............................   Passed    9.12 sec
36/41 Test  #3: ArrayWrapper ...............................   Passed    8.67 sec
37/41 Test #22: UniqueHandleDefaultArguments ...............   Passed    8.49 sec
38/41 Test #23: Video ......................................   Passed    8.66 sec
39/41 Test #20: StridedArrayProxy ..........................   Passed    8.61 sec
40/41 Test #40: MODULE_UniqueHandleDefaultArguments ........   Passed   29.29 sec
41/41 Test #41: MODULE_Video ...............................   Passed   29.34 sec

98% tests passed, 1 tests failed out of 41

Total Test time (real) =  71.57 sec

The following tests FAILED:
         39 - MODULE_StructureChain (Failed)
Errors while running CTest
```

So, the tests (mostly) work, but we need to special-case the generation of modules to switch to Ninja or additionally use Ninja.

M2-TE:
That test failure on `StructureChain` is something to look into, but we have some options for the CI setup:
1. Use `Ninja` when in C++23 mode
2. Explicitly use `Ninja Multi-Config` in current MSVC runner, should be no functional difference compared to MSBuild
3. New MSVC runner that only compiles C++23 to test for modules

I think 2 might be the best solution here and just use `Ninja Multi-Config`. Thoughts?

sharadhr:
> I think 2 might be the best solution here and just use Ninja Multi-Config. Thoughts?

@asuessenbach needs to input on this. I believe he works on Windows with VS and MSBuild rather than Ninja (as do many consumers of Vulkan-Hpp), and I reckon he'd prefer if MSBuild was still tested. 

So probably (1) or (3).

asuessenbach:
I agree with @sharadhr, option (2) is not that good, even it it "should" behave the very same as when using MSBuild.

That is, I'd prefer option (3), as that test actually is something new and deserves some additional run.
At some point in the hopefully to too distant future, when CMake and/or MSBuild can handle all that stuff, we can combine those two runners into one again.

M2-TE:
There's now a new `ci-windows-ext.yml` runner that only runs the module tests (samples disabled) in C++23.
Due to issues with MSVC, I had to temporarily disable two module tests:
* `FunctionCalls`: Internal compiler error regarding the `getMemoryProperties()` function, seems to have something to do with dynamic dispatching.
* `StructureChain`: MSVC cannot deduce the types in the structured binding section of the test.

sharadhr:
Should we look at fixing those tests if possible? 

In the case of the ICE we should probably report to [MSVC Developer Community](https://developercommunity.visualstudio.com/cpp).

M2-TE:
Looking into these tests would be good (I don't have a Windows VM on hand atm, I'll try to set one up). There are other disabled module tests worth looking into, but they require changes to the headers or module itself, rather than the tests. Likely best left to individual PRs for each test failure.

What I find interesting is that your run of the tests did not ICE at `FunctionCalls`. I've attached the log of my failed runner below:

<details>

<summary>ci-windows-ext.yml excerpt</summary>

```
32/40 Test #32: MODULE_FunctionCalls .......................***Failed   41.66 sec
Internal cmake changing into directory: D:/a/Vulkan-Hpp/Vulkan-Hpp/build/tests/tests/MODULE_FunctionCalls
======== CMake output     ======
The CXX compiler identification is MSVC 19.44.35221.0
Detecting CXX compiler ABI info
Detecting CXX compiler ABI info - done
Check for working CXX compiler: C:/Program Files/Microsoft Visual Studio/2022/Enterprise/VC/Tools/MSVC/14.44.35207/bin/Hostx64/x64/cl.exe - skipped
Detecting CXX compile features
CMake Warning (dev) at D:/a/_temp/-1721498437/cmake-4.1.2-windows-x86_64/share/cmake-4.1/Modules/Compiler/CMakeCommonCompilerMacros.cmake:248 (cmake_language):
  CMake's support for `import std;` in C++23 and newer is experimental.  It
  is meant only for experimentation and feedback to CMake developers.
Call Stack (most recent call first):
  D:/a/_temp/-1721498437/cmake-4.1.2-windows-x86_64/share/cmake-4.1/Modules/CMakeDetermineCompilerSupport.cmake:113 (cmake_create_cxx_import_std)
  D:/a/_temp/-1721498437/cmake-4.1.2-windows-x86_64/share/cmake-4.1/Modules/CMakeTestCXXCompiler.cmake:83 (CMAKE_DETERMINE_COMPILER_SUPPORT)
  CMakeLists.txt:16 (project)
This warning is for project developers.  Use -Wno-dev to suppress it.

Detecting CXX compile features - done
Configuring done (1.0s)
Generating done (0.0s)
Build files have been written to: D:/a/Vulkan-Hpp/Vulkan-Hpp/build/tests/tests/MODULE_FunctionCalls
======== End CMake output ======
Change Dir: 'D:/a/Vulkan-Hpp/Vulkan-Hpp/build/tests/tests/MODULE_FunctionCalls'

Run Clean Command: D:/a/_temp/-1721498437/ninja.exe clean
[1/1] Cleaning all built files...
Cleaning... 0 files.

Run Build Command(s): D:/a/_temp/-1721498437/ninja.exe
[1/16] Scanning "C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Tools\MSVC\14.44.35207\modules\std.compat.ixx" for CXX dependencies
std.compat.ixx
[2/16] Scanning D:\a\Vulkan-Hpp\Vulkan-Hpp\tests\FunctionCalls\FunctionCalls.cpp for CXX dependencies
[3/16] Scanning "C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Tools\MSVC\14.44.35207\modules\std.ixx" for CXX dependencies
std.ixx
[4/16] Scanning D:\a\Vulkan-Hpp\Vulkan-Hpp\vulkan\vulkan_video.cppm for CXX dependencies
vulkan_video.cppm
[5/16] Generating CXX dyndep file CMakeFiles\__cmake_cxx23.dir\CXX.dd
[6/16] Scanning D:\a\Vulkan-Hpp\Vulkan-Hpp\vulkan\vulkan.cppm for CXX dependencies
vulkan.cppm
[7/16] Generating CXX dyndep file CMakeFiles\VulkanHppModule.dir\CXX.dd
[8/16] Generating CXX dyndep file CMakeFiles\MODULE_FunctionCalls.dir\CXX.dd
[9/16] Building CXX object CMakeFiles\__cmake_cxx23.dir\C_\Program_Files\Microsoft_Visual_Studio\2022\Enterprise\VC\Tools\MSVC\14.44.35207\modules\std.ixx.obj
std.ixx
[10/16] Building CXX object CMakeFiles\__cmake_cxx23.dir\C_\Program_Files\Microsoft_Visual_Studio\2022\Enterprise\VC\Tools\MSVC\14.44.35207\modules\std.compat.ixx.obj
std.compat.ixx
[11/16] Linking CXX static library __cmake_cxx23.lib
[12/16] Building CXX object CMakeFiles\VulkanHppModule.dir\D_\a\Vulkan-Hpp\Vulkan-Hpp\vulkan\vulkan_video.cppm.obj
vulkan_video.cppm
D:\a\Vulkan-Hpp\Vulkan-Hpp\vulkan\vulkan_video.cppm(19): warning: 
	The Vulkan-Hpp C++ named module is experimental. It is subject to change without prior notice.
	To silence this warning, define the VULKAN_HPP_CXX_MODULE_EXPERIMENTAL_WARNING macro.
	For feedback, go to: https://github.com/KhronosGroup/Vulkan-Hpp/issues
[13/16] Building CXX object CMakeFiles\VulkanHppModule.dir\D_\a\Vulkan-Hpp\Vulkan-Hpp\vulkan\vulkan.cppm.obj
vulkan.cppm
D:\a\Vulkan-Hpp\Vulkan-Hpp\vulkan\vulkan.cppm(20): warning: 
	The Vulkan-Hpp C++ named module is experimental. It is subject to change without prior notice.
	To silence this warning, define the VULKAN_HPP_CXX_MODULE_EXPERIMENTAL_WARNING macro.
	For feedback, go to: https://github.com/KhronosGroup/Vulkan-Hpp/issues
[14/16] Linking CXX static library VulkanHppModule.lib
[15/16] Building CXX object CMakeFiles\MODULE_FunctionCalls.dir\FunctionCalls.cpp.obj
FAILED: [code=3221225477] CMakeFiles/MODULE_FunctionCalls.dir/FunctionCalls.cpp.obj 
C:\PROGRA~1\MICROS~2\2022\ENTERP~1\VC\Tools\MSVC\1444~1.352\bin\Hostx64\x64\cl.exe  /nologo /TP -DVK_NO_PROTOTYPES -DVULKAN_HPP_DISPATCH_LOADER_DYNAMIC -DVULKAN_HPP_USE_CXX_MODULE -ID:\a\Vulkan-Hpp\Vulkan-Hpp -ID:\a\Vulkan-Hpp\Vulkan-Hpp\Vulkan-Headers\include /DWIN32 /D_WINDOWS /EHsc /MDd /Zi /Ob0 /Od /RTC1 -std:c++latest /W4 /WX /showIncludes @CMakeFiles\MODULE_FunctionCalls.dir\FunctionCalls.cpp.obj.modmap /FoCMakeFiles\MODULE_FunctionCalls.dir\FunctionCalls.cpp.obj /FdCMakeFiles\MODULE_FunctionCalls.dir\ /FS -c D:\a\Vulkan-Hpp\Vulkan-Hpp\tests\FunctionCalls\FunctionCalls.cpp
D:\a\Vulkan-Hpp\Vulkan-Hpp\vulkan\vulkan_funcs.hpp(405): fatal error C1001: Internal compiler error.
(compiler file 'msc1.cpp', line 1589)
 To work around this problem, try simplifying or changing the program near the locations listed above.
If possible please provide a repro here: https://developercommunity.visualstudio.com/ 
Please choose the Technical Support command on the Visual C++ 
 Help menu, or open the Technical Support help file for more information
D:\a\Vulkan-Hpp\Vulkan-Hpp\vulkan\vulkan_funcs.hpp(405): note: the template instantiation context (the oldest one first) is
D:\a\Vulkan-Hpp\Vulkan-Hpp\tests\FunctionCalls\FunctionCalls.cpp(162): note: see reference to function template instantiation 'vk::PhysicalDeviceMemoryProperties vk::PhysicalDevice::getMemoryProperties<vk::detail::DispatchLoaderDynamic,true>(const Dispatch &) noexcept const' being compiled
        with
        [
            Dispatch=vk::detail::DispatchLoaderDynamic
        ]
INTERNAL COMPILER ERROR in 'C:\PROGRA~1\MICROS~2\2022\ENTERP~1\VC\Tools\MSVC\1444~1.352\bin\Hostx64\x64\cl.exe'
    Please choose the Technical Support command on the Visual C++
    Help menu, or open the Technical Support help file for more information
ninja: build stopped: subcommand failed.
Access violation
CMake Error: Generator: build tool execution failed, command was: D:/a/_temp/-1721498437/ninja.exe

Generator: build tool execution failed, command was: D:/a/_temp/-1721498437/ninja.exe
```

</details>

sharadhr:
Immediately, I see a version difference; I am using `cl.exe` version 19.50 (comes with VS 2026). From what I understand, a _lot_ of ICEs were fixed in this release.

M2-TE:
Can you try gating something in CMake via your MSVC version?
There the [MSVC_VERSION](https://cmake.org/cmake/help/latest/variable/MSVC_VERSION.html#variable:MSVC_VERSION) variable, but the documentation does not state anything about versions 19.0+.

Perhaps `CMAKE_CXX_COMPILER_VERSION` will work better here, if it properly reports a version we can use to hide that test with.

sharadhr:
> the documentation does not state anything about versions 19.0+

That's the last line: 

> 1950-1959 VS 18.0 (v145 toolset)

M2-TE:
I can't be the only one who would get confused by that versioning scheme, right..?
Anyways, the test is now properly version-gated.

sharadhr:
> I can't be the only one who would get confused by that versioning scheme, right..?

Well there are many simultaneous versions:

1. The Visual Studio _IDE_ version, which is 18.0
2. The MSVC and MSBuild _platform toolset_, which is v14.5 (or v145); these include the C++ standard library
3. The `cl.exe` _compiler_ version, is reported in the banner when executing `cl.exe`, which is v19.50 (or `1950`, as reported by the `_MSC_VER` macro)
4. The Windows SDK version, which since a while now has been reported with a five digit number.

tests/Flags/Flags.cpp
```diff
@@ -107,7 +107,9 @@ int main( int /*argc*/, char ** /*argv*/ )
   mhf0 ^= mhf3;
   assert( mhf0 == ( vk::MemoryHeapFlagBits::eMultiInstance ) );
 
-  vkmhf = static_cast<VkMemoryHeapFlags>(mhf0);
+  assert( mhf0 );
```
asuessenbach: Surplus assert here?

M2-TE: Good catch, snuck in during the merge

M2-TE:
What should be done about the `StructureChain` test? It is currently disabled for MSVC C++ module tests only and requires changes that are not related to the tests to make it work. IMO this should be addressed in a separate PR, but what do you think @asuessenbach?

asuessenbach:
> What should be done about the StructureChain test? It is currently disabled for MSVC C++ module tests only and requires changes that are not related to the tests to make it work. IMO this should be addressed in a separate PR, but what do you think @asuessenbach?

Yes, handling that in a separate PR would be fine. We just should not forget about it.

.github/workflows/ci-windows-ext.yml
```diff
@@ -0,0 +1,121 @@
+
+name: CI Windows Clang
```
sharadhr:
```suggestion
name: CI Windows MSVC-Ninja
```

.github/workflows/ci-windows-msvc.yml
sharadhr:
```suggestion
name: CI Windows MSVC-MSBuild
```

.github/workflows/ci-windows-clang.yml
sharadhr:
```suggestion
name: CI Windows ClangCL-Ninja
```

sharadhr (changes requested):
A few minor changes, just to make things more clear in the actions page.

M2-TE:
> We just should not forget about it

Indeed, all the disabled tests alongside their reason for being disabled can be viewed in `tests/CMakeLists.txt`.
There are a few things which will require fixes on the header/module side for some other tests as well.

sharadhr:
Great to see all the tests passing. I have no further additions. @M2-TE, thank you so much for your effort and collaborative spirit on this.

M2-TE:
Thank you as well @sharadhr for your continued input! No further additions from my side either.

asuessenbach:
@M2-TE, @sharadhr It was a great pleasure to see how this PR grew and flourished thanks to your collaboration here. It's great to have both of you contributing.

PR #2305, opened 2025-09-29
Extend test DesignatedInitializers to use a struct with deprectated members.

### LINKED ISSUE #2299: Missing designated field initializers with `vk::DeviceCreateInfo`
Using designated initializers with `VULKAN_HPP_NO_CONSTRUCTORS` results in warnings when dealing with members that have the `[deprecated]` tag. You either trigger a deprecation warning by setting them or you trigger a missing-designated-field-initializers warning by not setting them. Below is an example of how to trigger it.
```cpp
vk::DeviceCreateInfo info_device {
    .pNext = &required_features,
    .queueCreateInfoCount = (uint32_t)info_queues.size(),
    .pQueueCreateInfos = info_queues.data(),
    .enabledExtensionCount = (uint32_t)extensions.size(),
    .ppEnabledExtensionNames = extensions.data(),
};
```
-> `warning: missing field 'enabledLayerCount' initializer [-Wmissing-designated-field-initializers]`

The reason behind it is the lack of default assignment in the member declarations for the deprecated members (which makes sense, default assignment would trigger the deprecation warning too):
```cpp
// vulkan_structs.hpp -> DeviceCreateInfo
  public:
    StructureType                 sType                = StructureType::eDeviceCreateInfo;
    const void *                  pNext                = {};
    DeviceCreateFlags             flags                = {};
    uint32_t                      queueCreateInfoCount = {};
    const DeviceQueueCreateInfo * pQueueCreateInfos    = {};
    VULKAN_HPP_DEPRECATED( "ignored" ) uint32_t enabledLayerCount;
    VULKAN_HPP_DEPRECATED( "ignored" ) const char * const * ppEnabledLayerNames;
    uint32_t                       enabledExtensionCount   = {};
    const char * const *           ppEnabledExtensionNames = {};
    const PhysicalDeviceFeatures * pEnabledFeatures        = {};
```

setting `#pragma clang diagnostic ignored "-Wmissing-designated-field-initializers"` for the struct in `vulkan_structs.hpp` seems to have no effect.

Should we really use the deprecation tag for this struct? Or better yet, could we maybe even get rid of the members entirely?

asuessenbach:
Resolves #2299.

### DIFF
--- a/VulkanHppGenerator.cpp
+++ b/VulkanHppGenerator.cpp
@@ -12265,7 +12265,7 @@ std::tuple<std::string, std::string, std::string, std::string>
         members += " : " + member.bitCount;  // except for bitfield members, where no default member initializatin
                                              // is supported (up to C++20)
       }
-      else if ( member.deprecated.empty() )
+      else
       {
         members += " = ";
         auto enumIt = m_enums.find( member.type.type );
--- a/tests/DesignatedInitializers/DesignatedInitializers.cpp
+++ b/tests/DesignatedInitializers/DesignatedInitializers.cpp
@@ -17,11 +17,12 @@
 
 #include <cstdint>
 #include <iostream>
+#include <vector>
 #ifdef VULKAN_HPP_USE_CXX_MODULE
-# include <vulkan/vulkan_hpp_macros.hpp>
-  import vulkan_hpp;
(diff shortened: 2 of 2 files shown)


M2-TE:
Oh I didn't even see this. Thanks a lot for the fix!

PR #2336, opened 2025-10-30
CI: Properly set build type for Multi-Config MSVC builds

### LINKED ISSUE #2333: Windows runners never actually build in Release
From what I can tell, Windows runners have always used `CMAKE_BUILD_TYPE=$BUILD_MODE` to switch between `Debug` and `Release`.
As per the CMake documentation for [CMAKE_BUILD_TYPE](https://cmake.org/cmake/help/latest/variable/CMAKE_BUILD_TYPE.html) though, this is only relevant for single-config builds (Ninja, Make), whereas MSBuild is multi-config and therefore ignores this flag.

It instead needs to use the more explicit `--config $BUILD_MODE` to configure this for the _build_ (--build), instead of during _configuration_.

M2-TE:
Resolves #2333


asuessenbach:
Thanks a lot for you contribution!

PR #2340, opened 2025-11-02
Do not `export import std` as a macro

### LINKED ISSUE #2339: Exported module names cannot be macros
We currently define the `std` module as a macro, and `export import` it. This will no longer be allowed as of [P3034R1: Module Declarations Shouldn’t be Macros](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2024/p3034r1.html):

> No [identifier](https://eel.is/c++draft/lex.name#nt:identifier) in the [pp-module-name](https://eel.is/c++draft/cpp#nt:pp-module-name) or [pp-module-partition](https://eel.is/c++draft/cpp#nt:pp-module-partition) shall currently be defined as an object-like macro[.](https://eel.is/c++draft/cpp#module-2.sentence-1)

We need to fix the following to explicitly write out the module name, `std.compat`: https://github.com/KhronosGroup/Vulkan-Hpp/blob/4bc2057ad8342c361f31c8d5a84928aea355afc8/vulkan/vulkan.cppm#L36

sharadhr:
Resolves #2339.

### DIFF
--- a/snippets/CppmTemplate.hpp
+++ b/snippets/CppmTemplate.hpp
@@ -13,12 +13,7 @@ module;
   "To silence this warning, define the VULKAN_HPP_CXX_MODULE_EXPERIMENTAL_WARNING macro.\n\n" \
   "For feedback, go to: https://github.com/KhronosGroup/Vulkan-Hpp/issues"
 
-
-#  if defined(__clang__) || defined(__GNUC__) || defined(__GNUG__)
-_Pragma(VULKAN_HPP_STRINGIFY(GCC warning VULKAN_HPP_CXX_MODULE_EXPERIMENTAL_WARNING))
-#  elif defined(_MSC_VER)
-_Pragma(VULKAN_HPP_STRINGIFY(message(__FILE__ "(" VULKAN_HPP_STRINGIFY(__LINE__) "): warning: " VULKAN_HPP_CXX_MODULE_EXPERIMENTAL_WARNING)))
-#  endif
+  VULKAN_HPP_COMPILE_WARNING( VULKAN_HPP_CXX_MODULE_EXPERIMENTAL_WARNING )
 #endif
 
 #include <vulkan/${api}.hpp>
--- a/snippets/MacrosHppTemplate.hpp
+++ b/snippets/MacrosHppTemplate.hpp
@@ -30,6 +30,20 @@ ${licenseHeader}
 #  include <ciso646>
 #endif
 
+#define VULKAN_HPP_STRINGIFY2( text ) #text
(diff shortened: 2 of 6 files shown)


M2-TE:
I reckon `VULKAN_HPP_STD_MODULE` could then be removed from `vulkan_hpp_macros.hpp` as well, right?

sharadhr:
We use it `vulkan.hpp`for the `import` statement, so probably not.

M2-TE:
Hm, that's true.. could you test whether e.g. importing std in vulkan.cppm and std.compat in vulkan.hpp causes issues?

Actually, GCC 15.2.1 seems to be unable to handle `import VULKAN_HPP_STD_MODULE` alongside `export import std.compat`.
Clang doesn't mind, but GCC seems to postpone parsing the macro to _after_ module mapping and doesn't notice that the same module is being imported in `vulkan.cppm` and `vulkan.hpp`... At the very least, I'm getting duplicate definitions.

Should we just get rid of the macro and do an explicit `import std` in `vulkan.hpp` as well?

sharadhr:
> Should we just get rid of the macro and do an explicit import std in vulkan.hpp as well?

I'm not too fussed either way. Other things like [ReSharper and CLion](https://youtrack.jetbrains.com/issue/RSCPP-37063/Imported-module-name-could-came-from-macro-expansion) break because of the macro definition of the module, so perhaps we should just remove it. Although I'd really like the feature-testing checks... Not sure how to work those in.

M2-TE:
What about another `#pragma warning` that emits a warning if those feature tests fail?
This would make me quite happy, as it means Clang modules would work without a fuss as well (feature tests fail on there, but modules compile just fine).

Could replace the current module stuff in the macro header, so something like:

```cpp
#if defined( VULKAN_HPP_CXX_MODULE ) && ( !defined( __cpp_modules ) || !defined( __cpp_lib_modules ))
    // emit warning about failing feature tests
#endif
```

sharadhr:
Happy with that. Maybe elevate the pragma to `#error`? But that would mean Clang not working, which would be a pain. _De facto_ modules and the standard module works, so I think emitting a warning is good enough, yes.

I'd like to De Morgan's Law-ify the check:

```
#if defined( VULKAN_HPP_CXX_MODULE ) && !( defined( __cpp_modules ) && defined( __cpp_lib_modules ))
```

M2-TE:
"De Morgan's Law" gave me some ptsd.. I like that solution though.

asuessenbach:
Looks good again.
If both of you, @sharadhr and @M2-TE, are satisfied with this PR, I'll merge it.

M2-TE:
LGTM!

sharadhr:
LGTM as well.

PR #2303, opened 2025-09-27
Simplify named module handling and macros

sharadhr:
Obviates and resolves #2301.  

This fundamentally reworks `module vulkan_hpp;`.

It **requires and exports** `module std;` unconditionally. The module is blocked from being set up in CMake if users do not have an appropriate toolchain by virtue of the `__CMAKE::CXX23` target, which is discussed [in this Slack thread](https://cpplang.slack.com/archives/C5Y2GDECX/p1758882551809159).

PR #2349, opened 2025-11-11
Merge Vulkan Video module into main module

### LINKED ISSUE #2310: Update/merge Vulkan-Video C++ named module with main C++ named module
The Vulkan-Video C++ named module was created a while ago and hasn't really been maintained since: https://github.com/KhronosGroup/Vulkan-Hpp/blob/07a9e8911c91bf923c5fd790bd68db0af4efc982/vulkan/vulkan_video.cppm#L20

We should spend some time updating this to coincide with updates for the main named module. 

Some questions:

1. Do we keep the module separate?
2. Do we merge it into `module vulkan_hpp`, with no separation whatsoever?
3. How about a [module partition](https://learn.microsoft.com/en-gb/cpp/cpp/modules-cpp?view=msvc-170#module-partitions), e.g. `module vulkan_hpp:video`?

CC: @asuessenbach @M2-TE

### LINKED ISSUE #2311: C++20 module name
The module is currently set to `vulkan_hpp`, which doesn't really adhere to any of the, frankly still developing, naming schemes for modules.
Since it is still experimental, I would like to propose changing the module name and wanted to hear some thoughts on it.

1. `vulkan.hpp`: Module names are officially allowed to include dots, often eluding to paths. A module in `foo/faa.cppm` could be exported as `export module foo.faa` to prevent module name clashes and also "categorize" it. This does not really apply here, as we can cram everything into a single module, so it can instead be used to signify the kind of vulkan headers that are being imported. Namely hinting at the `hpp` portion.
2. `vulkan`: Choosing plain old `vulkan` might be the nicest option as C does not have any modules at all. It would be very clear intent (importing C++ headers) and also concise. Also suits the `vulkan` namespace that everything is in.

I personally would love to simply have `import vulkan` as the concise option, particularly if we may want to merge the video module into the main one as per #2310.

sharadhr:
Resolves #2310 and resolves #2311.

### DIFF
--- a/README.md
+++ b/README.md
@@ -797,7 +797,7 @@ Some functions might provide information that depends on the vulkan version. As 
 
 #### Overview <a id='named_module_overview'>
 
-Vulkan-Hpp provides a [C++ named module](https://en.cppreference.com/w/cpp/language/modules), `vulkan_hpp` in [`vulkan.cppm`](vulkan/vulkan.cppm).
+Vulkan-Hpp provides a [C++ named module](https://en.cppreference.com/w/cpp/language/modules), `vulkan` in [`vulkan.cppm`](vulkan/vulkan.cppm).
 C++ modules are intended to supersede header files. Modules have potential to drastically improve compilation times for large projects, as declarations and definitions may be easily shared across translation units without repeatedly parsing headers.
 Vulkan-Hpp has some extremely long headers (e.g. [`vulkan_structs.hpp`](vulkan/vulkan_structs.hpp)), and the C++ module is likely to shorten compile times for projects currently using it.
 
@@ -882,7 +882,7 @@ target_link_libraries( VulkanHppModule PUBLIC Vulkan::Headers )
 Finally, supply the macro `VULKAN_HPP_DEFAULT_DISPATCH_LOADER_DYNAMIC_STORAGE` exactly once in your source code, just as in the non-module case. In order to have that macro available, include [`vulkan_hpp_macros.hpp`](vulkan/vulkan_hpp_macros.hpp), a lightweight header providing all Vulkan-Hpp related macros and defines. And as explained above, you need to initialize that dispatcher in two or three steps:
 
 ```cpp
-import vulkan_hpp;
+import vulkan;
 
 #include <vulkan/vulkan_hpp_macros.hpp>
 


M2-TE:
The import name in the tests needs to be changed as well.
Might also be good to update the readme:

https://github.com/KhronosGroup/Vulkan-Hpp/blob/63a3afa9b290d3265780ddcf950bc0b75b8fb601/README.md?plain=1#L800

https://github.com/KhronosGroup/Vulkan-Hpp/blob/63a3afa9b290d3265780ddcf950bc0b75b8fb601/README.md?plain=1#L885

M2-TE:
How about we update the minimal example in readme for `import vulkan` with dynamic dispatching as well?
Notably, we fix the import/include order (breaks modules on gcc otherwise) and provide some more info.
The previously used macros were also not really necessary in the example:
```cpp
// required to access macros (otherwise, use vk::detail::DispatchLoaderDynamic directly)
// import after all other #includes
import vulkan;

// when vulkan_hpp_macros.hpp was included

auto main2(int argc, char* const argv[]) -> int
{
    // initialize minimal set of function pointers
    VULKAN_HPP_DEFAULT_DISPATCHER.init();

    auto appInfo = vk::ApplicationInfo("My App", 1, "My Engine", 1, vk::makeApiVersion(1, 0, 0, 0));
    // ...
}
```

sharadhr:
Thanks for the insight. I've been meaning to update the readmes for a while, including the guidance on different compilers, build systems, etc (the module requires latest _everything_).

asuessenbach:
I'm sorry, a new test has materialized: NoDefaultDispatcher.
Would you please adjust that as well, to import vulkan, instead of vulkan_hpp?

sharadhr:
@asuessenbach this is ready. 

@M2-TE I have made the minimum necessary edits to the readme (module name only). All other changes I intend to work in as part of #2358.

theHamsta:
Hi! This looks good to me. I enabled individual tests like in https://github.com/KhronosGroup/Vulkan-Hpp/pull/2292 and they worked fine with the corresponding changes in C++23 mode with CMake's experimental `import std`-support and clang-21.

Would it be possible to enable at least one configuration in CI that compiles at least one test with C++ modules enabled. It does not need to be all of https://github.com/KhronosGroup/Vulkan-Hpp/pull/2292. We could also rebase https://github.com/KhronosGroup/Vulkan-Hpp/pull/2292 on this to have a CI for these changes.

Is C++ modules support bound to C++23 `import std` support (seems to be used in the header when c++ modules are enabled)? Please pardon me ignorance. I just came back to this project after a longer time.

We also would need to update a C++20 modules test in the Vulkan-Header repository after merging this change.

sharadhr:
Hi @theHamsta! Long time, no see. 

> Would it be possible to enable at least one configuration in CI that compiles at least one test with C++ modules enabled.

Yes, that makes sense, to avoid embarrassments like #2362 slipping in.

As for #2292, I feel it has fallen afoul of scope creep and really needs to be broken up, into at least two PRs. One for testing `module std`, another for CTest. 

> Is C++ modules support bound to C++23 import std support (seems to be used in the header when c++ modules are enabled)?

For now, yes. Proper module support across all three toolchains (GCC, Clang, MSVC) is still quite experimental, and we're demanding the latest toolchains anyway, so might as well ask for the latest standard as well. It also massively simplifies how the module is used by the end user, and also simplifies the CMake build logic. As for C++20, all three compiler vendors have [pledged to backport `import std`](https://github.com/microsoft/STL/issues/3945). Some build systems support this out-of-the-box (xmake, build2), some others don't (CMake). End-users can still manually set it up, we ship the module interface file anyway.

> We also would need to update a C++20 modules test in the Vulkan-Header repository after merging this change.

That test has been disabled, but after I write a CI test for this, we can re-enable that so we get modules compiling properly.

M2-TE:
Edit: Looks like @sharadhr was just a second faster.. but yes, perhaps #2292 should be broken into several smaller PRs.

> Would it be possible to enable at least one configuration in CI that compiles at least one test with C++ modules enabled. It does not need to be all of https://github.com/KhronosGroup/Vulkan-Hpp/pull/2292. We could also rebase https://github.com/KhronosGroup/Vulkan-Hpp/pull/2292 on this to have a CI for these changes.

I had benched that PR for a little while until I figured out how to get CI working with modules. As mentioned in #2362, @YaaZ has a working workflow for this, so I'll check that out and try to incorporate it.

> Is C++ modules support bound to C++23 `import std` support 

Yes and no. You could use modules in C++20, but that would mean doing half things by compiling modules and still including headers. We decided to make `import std` it a hard requirement for the "true module way", which was implemented in #2303.

> We also would need to update a C++20 modules test in the Vulkan-Header repository after merging this change.

Indeed, but that will require more thorough testing, so that it does not block releases again.

theHamsta:
> Hi @theHamsta! Long time, no see.
> 

Since this is breaking change, I would prefer @asuessenbach to merge. He will be back next week.

> I had benched that PR for a little while until I figured out how to get CI working with modules. As mentioned in https://github.com/KhronosGroup/Vulkan-Hpp/pull/2362, @YaaZ has a working workflow for this, so I'll check that out and try to incorporate it.

So maybe a good strategy would then be to leave this PR as is and have a minimal follow-up that enables CI for modules. Just building some module tests for some configuration might be enough as a first step.

asuessenbach:
I know, I should have asked this quite a while ago...
But could you split this PR in two, one to rename the module from vulkan_hpp to vulkan, and one to actually merge the video module into main?

sharadhr:
I can, but at least one of the PRs will run into a conflict and will have to be correctly rebased, one on top of another.

asuessenbach:
> I can, but at least one of the PRs will run into a conflict and will have to be correctly rebased, one on top of another.

Yes, that's to be expected.
But I think, merging in video and renaming vulkan_hpp to vulkan are two separate things that earn their own PR, each.
Moreover, the renaming part would need some mentioning in the [Breaking Changes](https://github.com/KhronosGroup/Vulkan-Hpp#breaking-changes-) section of the readme.

sharadhr:
@asuessenbach I will close this; I have opened #2372 and #2373 instead.

sharadhr:
@asuessenbach I will close this; I have opened #2372 and #2373 instead.

PR #2352, opened 2025-11-11
Replace stragglers of assert and static_assert with the relevant macros

M2-TE:
Both `assert` and `static_assert` are defined in `vulkan_hpp_macros.hpp` and are mostly used throughout the headers. There were just a few places that manually used them directly without the macro. Redefining the macro to a custom function would therefore not replace all of them.


asuessenbach:
Thanks again for this nice little clean-up.

PR #2355, opened 2025-11-11
Provide other constructors for ObjectDestroy and ObjectFree

### LINKED ISSUE #2354: VULKAN_HPP_NO_DEFAULT_DISPATCHER is broken
There are places where function parameters have default assignments, followed by a dispatcher with `VULKAN_HPP_DEFAULT_DISPATCHER_ASSIGNMENT`.
This breaks when `VULKAN_HPP_NO_DEFAULT_DISPATCHER` is defined with errors like:
```
vulkan/vulkan.hpp:6637:39: error: missing default argument on parameter 'dispatch'
```

M2-TE:
Resolves #2354.

When `VULKAN_HPP_NO_DEFAULT_DISPATCHER` is used, both `ObjectDestroy` and `ObjectFree` will not have valid constructors, as the optional parameters always need to come last:

https://github.com/KhronosGroup/Vulkan-Hpp/blob/9423ad7f699c17740b79e63c083502fb6deddb84/vulkan/vulkan.hpp#L6634-L6642

A simple solution is to provide a set of two constructors each, which use overloads to make `allocationCallbacks` optional, even when `dispatcher` isn't. These will only be active while `VULKAN_HPP_NO_DEFAULT_DISPATCHER` is used.

I used `#if` for conditional compilation, in fear of breaking something otherwise. It _should_ however be fine to _only_ provide the two new constructors, so please let me know if that is preferred.

### DIFF
--- a/snippets/ObjectDestroy.hpp
+++ b/snippets/ObjectDestroy.hpp
@@ -4,14 +4,34 @@ class ObjectDestroy
 public:
   ObjectDestroy() = default;
 
+// dispatcher parameter cannot be placed after allocationCallbacks when default dispatcher is disabled
+#if defined( VULKAN_HPP_NO_DEFAULT_DISPATCHER )
   ObjectDestroy( OwnerType                           owner,
-                 Optional<const AllocationCallbacks> allocationCallbacks = nullptr,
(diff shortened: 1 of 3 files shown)


M2-TE:
There are a few breakages aside from the reported case, so I'll put together a new test for this macro.

PR #2357, opened 2025-11-11
Export vk::raii::isVulkanRAIIHandleType

YaaZ:
`vk::isVulkanHandleType` is exported, but `vk::raii::isVulkanRAIIHandleType` is not


asuessenbach:
Nice catch.

PR #2358, opened 2025-11-11
Rework the readme

### LINKED ISSUE #2463: Cpp20Modules example link not working
https://github.com/KhronosGroup/Vulkan-Hpp/blob/main/README.md#usage-with-cmake-

At the bottom of the "Usage with CMake" text, it says

> An example is provided in tests/Cpp20Modules/Cpp20Modules.cpp.

However, the link leads to a "404 - page not found". 
Where can i find the example? Best regards!

sharadhr:
EDIT for posterity:

Resolves #2463


asuessenbach:
Is there a reason you want to rework the readme?

sharadhr:
Hi @asuessenbach! I've been meaning to do this for a while, actually. I have a list of things that I found a little less than ideal with the current readme:

- Be clear on what this repository is actually for (development of the Vulkan-Hpp generator); the actual generated headers can be downloaded from Vulkan-Headers.
- More instructions on installation/usage from the Vulkan SDK, vcpkg, and Conan.
- Rearranging some of the information to flow a bit better; e.g. start with basic `Info` structs, then move on to specific Vulkan types like `Device`, `Image`, etc
  - Some of the information is a bit scattershot; e.g. there are two sections for RAII best practices; these could be combined.
  - Subsume the [RAII programming guide](vk_raii_ProgrammingGuide.md) into the readme, or split the readme up altogether.
  - Also group together the language-side facilities that Vulkan-Hpp uses, e.g. `[[nodiscard]]`, `std::expected`, `std::optional`, etc
- Remove/edit deprecated information, especially pertaining to `vulkan.cppm`
- Miscellaneous grammar, capitalisation, and formatting (e.g. as demonstrated, making all the configuration options demarcated ` ``in backticks`` ` makes it easy to avoid the `\_`stuff in the ToC. 
  - Adhere to some Common Markdown best practices, e.g. spaces around lists, `-` instead of `*`, etc. 
  - Also, the HTML hyperlinks are no longer necessary; as far as I understand, the Common Markdown specification provides a means to automatically generate a table of contents.
  
If you're OK with this, I can proceed with making these changes, naturally subject to your review at the end. I hope to improve the documentation at the end of this exercise.

asuessenbach:
You're more than welcome to rework the readme!
I'm pretty sure, any change will be an improvement.

M2-TE:
How about adding CMake's `FetchContent` to the "Installation and Usage" section?
Something akin to
```cmake
include(FetchContent)
FetchContent_Declare(vulkan-hpp
    GIT_REPOSITORY "https://github.com/KhronosGroup/Vulkan-Hpp.git"
    GIT_TAG "v1.4.336" # release tag, commit hash or branch name
    GIT_SHALLOW ON)
FetchContent_MakeAvailable(vulkan-hpp)
```

M2-TE:
For the "Build Instructions", it is likely best to separate header generation and sample/test building. At the very least, the build section should state that building is purely for those tasks and that headers can be used by just including them.

For merely using the headers (because e.g. they haven't landed in Vulkan-Headers yet), quickly showing how to link against `Vulkan::Hpp` in CMake would be sufficient for those users. Or showing minimal `include_directories(...)` usage.

sharadhr:
Both good points @M2-TE. I am focusing mostly on `Usage` for now (since that is where I see the biggest shortfall in documentation), but I will add these in to Building when I can.

sharadhr:
I'm much happier about the Usage document; probably a few final updates left for the last few sections. At this point, I'd like to ask you guys, @M2-TE and @asuessenbach to review what has been done; I have a _lot_ of `TODO`s that you might want to search for and comment on. I suggest reading the document in a proper, top-down manner to ensure it flows correctly. I'd also love to note any missing features that have been recently added.

___

After that, I intend to work on migrating the RAII guide to `Handles.md`, and finally revise the building and installation sections.

docs/Building.md
```diff
@@ -0,0 +1,46 @@
+# Building
+
+## Minimum requirements
+
+The generator, samples, and tests requires a C++11 compiler. The following compilers are known to work, and tested during releases:
```
M2-TE: The generator actually requires a C++20 compiler: https://github.com/KhronosGroup/Vulkan-Hpp/blob/67ec7c93ba68da2edb2e48f884747bb2db72f15e/CMakeLists.txt#L116

docs/Building.md
```diff
@@ -0,0 +1,46 @@
...
+
+> [!NOTE]
+> Compiling the C++ named module requires the most recent compiler toolchains possible.
+> This is detailed in the corresponding [section](./Usage.md/#c20-named-module).
+
+## Build steps
+
+1. Install dependencies:
+   - Install CMake and git; ensure both are available from a shell.
+   - Install the LunarG Vulkan SDK.
```
M2-TE: The SDK should not be needed for either the generators or the tests/samples. The headers and xml registry are fetched from the `Vulkan-Headers` submodule.
It might be worth noting here that, in order to build samples, you need additional deps on Ubuntu:
https://github.com/KhronosGroup/Vulkan-Hpp/blob/67ec7c93ba68da2edb2e48f884747bb2db72f15e/.github/workflows/build.yml#L79

docs/Building.md
```diff
@@ -0,0 +1,46 @@
...
+       `-DVULKAN_HPP_RUN_GENERATOR=ON` option to the CMake command line.
+5. The generated project may be opened with an IDE like Visual Studio or CLion, or the various targets may be built directly from a shell:
+
+    ```cmake --build build --parallel```
+
+## Additional features
+
+### Formatting
+
+If a `clang-format` executable is found by CMake, the define `CLANG_FORMAT_EXECUTABLE` is set accordingly. In that case, the generated headers are formatted using the `.clang-format` file located in the root directory of this project; otherwise, the formatting is left as hard-coded in the generator.
```
M2-TE: likely better to say "variable `CLANG_FORMAT_EXECUTABLE`", rather than "define".

docs/Building.md
```diff
@@ -0,0 +1,46 @@
...
+
+Use `clang-format` version 21.1.0 to format the generated files.
```
M2-TE: This line comes a bit out of nowhere after the previous paragraph; did you mean to say that you should use this version for PRs? It should probably say either that, or that it is the recommended version in general.

docs/Handles.md
```diff
@@ -0,0 +1,90 @@
...
+
+For each function which constructs a Vulkan handle of type `vk::Type` Vulkan-Hpp provides a second version which returns a `vk::UniqueType`. E.g. for `vk::Device::createBuffer` there is `vk::Device::createBufferUnique` and for `vk::allocateCommandBuffers` there is `vk::allocateCommandBuffersUnique`.
+
+Note that using `vk::UniqueHandle` comes at a cost since most deleters have to store the `vk::AllocationCallbacks` and parent handle used for construction because they are required for automatic destruction.
+
+## `vk::SharedHandle`
+
+Vulkan-Hpp provides a `vk::SharedHandle<Type>` interface. For each Vulkan handle type `vk::Type` there is a shared handle `vk::SharedType` which will delete the underlying Vulkan resource upon destruction, e.g. `vk::SharedBuffer` is the shared handle for `vk::Buffer`.
+
+Unlike `vk::UniqueHandle`, `vk::SharedHandle` takes shared ownership of the resource as well as its parent. This means that the parent handle will not be destroyed until all child resources are deleted. This is useful for resources that are shared between multiple threads or objects.
```
M2-TE: Since you were on the topic of multithreading, is it worth mentioning that shared handles are not thread safe? Handling mutexes (mutices?) is up to the user.

docs/Usage.md
```diff
@@ -0,0 +1,966 @@
...
+
+```c++
+vk::Image image = device.createImage({{}, vk::ImageType::e2D, vk::Format::eR8G8B8A8Unorm,
+                                     { width, height, 1 },
+                                     1, 1, vk::SampleCountFlagBits::e1,
+                                     vk::ImageTiling::eOptimal, vk::ImageUsageFlagBits::eColorAttachment,
+                                     vk::SharingMode::eExclusive, 0, nullptr, vk::ImageLayout::eUndefined});
+```
+
+#### Designated initialisers
```
M2-TE:
```suggestion
#### Designated initializers
```

docs/Building.md
```diff
@@ -0,0 +1,46 @@
...
+1. Install dependencies:
+   - Install CMake and git; ensure both are available from a shell.
+   - Install the LunarG Vulkan SDK.
+   - Optionally install `clang-format` with version at least 11.0 so that any generated headers can be correctly formatted.
+2. In a shell, clone the repository:
+
+    ```git clone --recurse-submodules https://github.com/KhronosGroup/Vulkan-Hpp.git```
+3. Change the shell's working directory to the newly created `Vulkan-Hpp` directory.
+4. Create a build environment with CMake:
+    ```cmake -DVULKAN_HPP_SAMPLES_BUILD=ON -DVULKAN_HPP_SAMPLES_BUILD_WITH_LOCAL_VULKAN_HPP=ON -DVULKAN_HPP_TESTS_BUILD=ON -DVULKAN_HPP_TESTS_BUILD_WITH_LOCAL_VULKAN_HPP=ON -B build```
```
M2-TE:
```suggestion
    ```cmake -DVULKAN_HPP_SAMPLES_BUILD=ON -DVULKAN_HPP_TESTS_BUILD=ON -B build```
```

I think you meant VULKAN_HPP_BUILD_WITH_LOCAL_VULKAN_HPP, as it does not have separate versions for tests/samples. Either way, you can probably leave it out as it defaults to ON.

docs/Building.md
M2-TE:
```suggestion
If a `clang-format` executable is found by CMake, the variable `CLANG_FORMAT_EXECUTABLE` is set accordingly. In that case, the generated headers are formatted using the `.clang-format` file located in the root directory of this project; otherwise, the formatting is left as hard-coded in the generator.
```

likely better to say "variable CLANG_FORMAT_EXECUTABLE", rather than "define".

docs/Configuration.md
```diff
@@ -0,0 +1,181 @@
+# Configuration Options
+
+Vulkan-Hpp offers several options to configure its behaviour and features, and these are exposed as compiler macros. The following macros may be defined in a build system, or in the command-line invocation of a compiler.
```
M2-TE: Is it worth mentioning that these configuration options apply to the header usage and not the generator?

docs/Usage.md
```diff
@@ -0,0 +1,966 @@
...
+
+>[!NOTE]
+> This feature requires a compiler supporting at least C++20, and must be **enabled** with the `VULKAN_HPP_NO_CONSTRUCTORS` or `VULKAN_HPP_NO_STRUCT_CONSTRUCTORS` macro.
+> This will disable the above-mentioned generated constructors for structs.
+
+C++20 and later versions support [designated initializers](https://en.cppreference.com/w/cpp/language/aggregate_initialization.html#Designated_initializers) for aggregate types.
+Designated initialisers allow for more expressive code: member variable names are mentioned immediately in plain-text, without needing any static analysis.
```
M2-TE:
```suggestion
Designated initializers allow for more expressive code: member variable names are mentioned immediately in plain-text, without needing any static analysis.
```

docs/Usage.md
```diff
@@ -0,0 +1,966 @@
...
+```c++
+// initialize the vk::ApplicationInfo structure
+vk::ApplicationInfo applicationInfo{ .pApplicationName   = AppName,
+                                     .applicationVersion = 1,
+                                     .pEngineName        = EngineName,
+                                     .engineVersion      = 1,
+                                     .apiVersion         = vk::ApiVersion11 };
+
+// initialize vk::InstanceCreateInfo
+vk::InstanceCreateInfo instanceCreateInfo{ .pApplicationInfo = &applicationInfo };
```
M2-TE, line 163:
```suggestion
vk::InstanceCreateInfo instanceCreateInfo{ .pApplicationInfo = &applicationInfo };

// initialize vk::InstanceCreateInfo directly within vk::createInstance
instance = vk::createInstance({ .pApplicationInfo = &applicationInfo });
```

This is a style that I have recently started using, which basically inlines your init structs into the respective function calls. It removes the need to have an additional variable name in the scope and, since we know what struct the function expects, is not obfuscated.

For me personally at least, it helped my code readability, especially in the more "busy" sections with lots of init structs. Pretty neat and concise too. Perhaps worth mentioning this style here?

docs/Usage.md
```diff
@@ -0,0 +1,966 @@
...
+// Query just vk::MemoryRequirements2KHR
+vk::MemoryRequirements2KHR memoryRequirements = device.getBufferMemoryRequirements2KHR({});
+```
+
+### Handles and functions
+
+<!-- TODO: maybe move this to ./Handles.md? -->
+
+Vulkan-Hpp provides a `vk::` class for each Vulkan handle, such as `vk::Instance`, `vk::Device`, `vk::Buffer`, and so on.
+These classes are binary-compatible with their corresponding C handles; they can be cast back and forth without any overhead.
```
M2-TE: Isn't this true only for 64 bit programs? It might be good to mention and/or link to `VULKAN_HPP_TYPESAFE_CONVERSION` here.

docs/Handles.md
```diff
@@ -0,0 +1,90 @@
...
+
+Other than the `vk::Handles`, all those handle wrapper classes need to hold additional data, and thus are not binary identical with the vulkan C-handles.
+
+As the `vk::UniqueHandles` and the `vk::SharedHandles` use the same dispatcher as the `vk::Handles`, they can be easily mixed-and-matched. The `vk::raii::Handles` use some slightly different dispatchers and thus are not compatible with the other handles! That is, for the `vk-Handles`, the `vk::UniqueHandles`, and the `vk::SharedHandles`, you need to instantiate a global dispatcher as described in <https://github.com/KhronosGroup/Vulkan-Hpp#extensions--per-device-function-pointers>. For the `vk::raii-Handles`, this is not needed, as they maintain their own dispatchers. The big advantage here is when you have multiple devices: the functions called via the `vk::raii-Handles` always call the device specific functions.
```
M2-TE:
```suggestion
As the `vk::UniqueHandles` and the `vk::SharedHandles` use the same dispatcher as the `vk::Handles`, they can be easily mixed-and-matched. The `vk::raii::Handles` use some slightly different dispatchers and thus are not compatible with the other handles! That is, for the `vk-Handles`, the `vk::UniqueHandles`, and the `vk::SharedHandles`, you need to instantiate a global dispatcher as described in [Usage](./Usage.md#extensions--per-device-function-pointers). For the `vk::raii-Handles`, this is not needed, as they maintain their own dispatchers. The big advantage here is when you have multiple devices: the functions called via the `vk::raii-Handles` always call the device specific functions.
```

The link here needs to be updated. I'm not sure whether a relative url like this works, but it should.

docs/Usage.md
```diff
@@ -0,0 +1,966 @@
...
+
+>[!NOTE]
+> Note that these handles do **not** support RAII; their lifetimes need to be manually managed with create-destroy pairs, just like in the C API.
+> For automatic resource management, refer to [Handles](./Handles.md).
+
+#### Passing structs to functions
+
+<!-- TODO: this section seems poorly worded and titled. -->
+
+Vulkan-Hpp generates references for pointers to structs. This conversion allows passing temporary structs to functions which can result in shorter code. In case the input is optional and thus accepting a null pointer, the parameter type will be `vk::Optional<T> const&`. This type accepts either a reference to `T` or `nullptr` as input and thus allows optional temporary structs.
```
M2-TE: This part reads in a bit of a confusing way (as you pointed out in TODO as I just noticed). A more concise way of saying these things would be "structs are passed to functions as constant references, therefore allowing temporary structs" or something akin to that. Mentioning stuff like `vk::ArrayProxy` that Vulkan-Hpp provides would probably also fit this category?

For the C++ example below, it might be nice to also see the C++ approach that is similar to C, i.e. making a vk::ImageSubresource struct an passing it, but then showing that you can also create it as a temporary struct.

This could perhaps be combined with the designated initializer stuff as well (or, at least mentioned here), since creating anonymous structs with those is another useful thing to learn about for passing structs.

docs/Usage.md
```diff
@@ -0,0 +1,966 @@
...
+The C API will be transparent to this, but the corresponding Vulkan-Hpp function signatures may change, and this will require updates to user code.
+These changes will be noted in [Breaking Changes](../README.md/#breaking-changes).
+
+#### Return value transformation with exceptions disabled
+
+As noted above, exceptions may be disabled by defining the `VULKAN_HPP_NO_EXCEPTIONS` macro.
+In this case `vk::ResultValue<SomeType>::type` is a struct containing `vk::Result` and `SomeType`.
+This struct may be unpacked as needed.
+
+<!-- TODO: This shader module example needs to do more so that it's a bit more indicative of how to handle errors. -->
```
M2-TE: I think it can be simplified using only a single shader module, just for readability. It is very dependant on the object at hand what the error handling should look like, so a comment as you have made it in the example below is sufficient imo.

docs/Usage.md
```diff
@@ -0,0 +1,966 @@
...
+
+   ```c++
+   // No exceptions, no return value transformation
+   vk::ShaderModuleCreateInfo createInfo(...);
+   vk::ShaderModule shader1;
+   vk::Result result = device.createShaderModule(&createInfo, allocator, &shader1);
+   if (result.result != vk::Result::eSuccess)
+   {
+     // handle error code;
+     // physicalDevice.destroyDevice(device, allocator); // cleanup; destroy device
```
M2-TE: This feels a bit out of scope for this example. You could mention cleaning up resources in general instead.

docs/Usage.md
```diff
@@ -0,0 +1,966 @@
...
+>[!NOTE]
+> This feature requires a compiler and standard library supporting at least C++23, and must be explicitly **enabled** with the `VULKAN_HPP_USE_STD_EXPECTED` macro.
+
+When `std::expected` is available, `vk::ResultValue<T>::type` is defined as `std::expected<T, vk::Result>`.
+The result can then be monadically handled with `and_then`, `transform`, and other member functions of `std::expected`.
+
+### Feature and property enumerations
+
+Related to error-handing, Vulkan enumerations require more careful handling of return values.
+The following code is usually typical:
```
M2-TE:
```suggestion
The following code is quite common:
```

usually typically normally, when the stars align

docs/Usage.md
```diff
@@ -0,0 +1,966 @@
...
+
+   - Use your own dynamic loader, which needs to provide a templated function `getProcAddress`. Refer to the implementation of `vk::detail::DynamicLoader` in `vulkan.hpp` for an example.
+
+     <!-- TODO: Add a skeleton implementation of `struct YourDynamicLoader`? -->
+     ```c++
+     YourDynamicLoader ydl;
+     VULKAN_HPP_DEFAULT_DISPATCHER.init(ydl);
+     ```
+
+     > [!NOTE]
```
M2-TE: The !NOTE does not work within a markup enumeration I believe. Having it formatted as a quote with `>` is probably sufficient.

docs/Usage.md
```diff
@@ -0,0 +1,966 @@
...
+       assert(!physicalDevices.empty());
+       vk::Device device = physicalDevices[0].createDevice({}, nullptr);
+       VULKAN_HPP_DEFAULT_DISPATCHER.init(device);
+   ```
+
+After step 2, the dispatcher is fully functional.
+Adding the third step can potentially result in more efficient code.
+If you intend to use multiple devices, this last step can be omitted, and delegated to the driver instead.
+
+<!-- TODO: what cases? -->
```
M2-TE, line 688: I was wondering this as well, I have yet to see a use case for it.

docs/Usage.md
```diff
@@ -0,0 +1,966 @@
...
+[`vulkan_hash.hpp`](../vulkan/vulkan_hash.hpp) specialises `std::hash` for `vk::Handle` bindings; with C++14, this is extended to the structure bindings too.
+Define `VULKAN_HPP_HASH_COMBINE` to customise the hash-combining algorithm for structure elements.
+
+### C++20 named module
+
+> [!WARNING]
+> The Vulkan-Hpp C++ named module is still **experimental** and its interface and usage may change often and **without prior notice** to support better usability and integration.
+> We strongly suggest using the **latest** possible toolchains and build systems, as older compilers may have incomplete support, or may emit internal compiler errors (ICEs).
+
+Vulkan-Hpp provides a [C++ named module](https://clang.llvm.org/docs/StandardCPlusPlusModules.html#standard-c-named-modules), `vulkan` in [`vulkan.cppm`](../vulkan/vulkan.cppm).
```
M2-TE:
```suggestion
Vulkan-Hpp provides a [C++ named module](https://clang.llvm.org/docs/StandardCPlusPlusModules.html#standard-c-named-modules), `vulkan` with [`vulkan.cppm`](../vulkan/vulkan.cppm) and [`vulkan_video.cppm`](../vulkan/vulkan_video.cppm).
```

docs/Usage.md
```diff
@@ -0,0 +1,966 @@
...
+C++ modules are intended to supersede header files.
+Modules tend to considerably improve compile times, as declarations and definitions may be easily shared across translation units without repeatedly parsing headers.
+This is particularly applicable to Vulkan-Hpp, as the generated headers are very large.
+
+Users who are able to upgrade to a recent toolchain (detailed below) should try to use the module instead of the headers, and report bugs and issues.
+
+<!-- TODO: leave this here or move to `Building`? There's already a small warning there. -->
```
M2-TE: I would leave it there, as it asks users to test the module if possible.

docs/Usage.md
```diff
@@ -0,0 +1,966 @@
...
+#### Compiler support
+
+The named module has been tested with the following toolchains:
+
+- MSVC 19.44 or later (as reported by the banner in `cl.exe`)
```
M2-TE:
```suggestion
- MSVC 19.44 or later
```
should specified in detail below (refer to the build.yaml)

docs/Usage.md
```diff
@@ -0,0 +1,966 @@
...
+- Clang 21.1.8 or later
+- GCC 15.1 or later
+- CMake version 4.2.1 or later
+
```
M2-TE, line 807: Here it should link to our build.yaml for an example setup of each platform (currently excluding only gcc).

docs/Usage.md
```diff
@@ -0,0 +1,966 @@
...
+#### CMake usage
+
+CMake is recommended for use with the Vulkan-Hpp named module, as it provides a convenient platform-agnostic way to configure your project.
+Version **3.30** or later is required for the Vulkan-Hpp module, which uses the standard library module feature.
+Refer to the [CMake documentation](https://cmake.org/cmake/help/latest/manual/cmake-cxxmodules.7.html) on the topic.
+
+CMake provides the [FindVulkan module](https://cmake.org/cmake/help/latest/module/FindVulkan.html), which may be used to source the Vulkan SDK and Vulkan headers on your system.
+
+When invoking CMake, make sure to provide the following UUID to enable CMake's experimental support for the C++ standard library module.
+This may also be set in a [CMake preset file](https://cmake.org/cmake/help/latest/manual/cmake-presets.7.html#configure-preset), in the `cacheVariables` key for a configure preset.
```
M2-TE:
```suggestion
This may also be set before your `project()` call, or in a [CMake preset file](https://cmake.org/cmake/help/latest/manual/cmake-presets.7.html#configure-preset) in the `cacheVariables` key for a configure preset.
Note that the UUID may change across CMake versions. To find it for your specific version, check out the correct release tag and look for `CMAKE_EXPERIMENTAL_CXX_IMPORT_STD` in [Help/dev/experimental.rst](https://gitlab.kitware.com/cmake/cmake/-/blob/master/Help/dev/experimental.rst).
```

docs/Usage.md
```diff
@@ -0,0 +1,966 @@
...
+
+```bash
+cmake -DCMAKE_EXPERIMENTAL_CXX_IMPORT_STD=d0edc3af-4c50-42ea-a356-e2862fe7a444 ...
+```
+
+A complete example `vulkan_cxx_module.cmake` file for a project using the Vulkan-Hpp named module is provided below.
+<!-- TODO: @M2-TE, any improvements here? Should we leave this out until CMake lands full support for `module std`? -->
```
M2-TE: Definetely leave it here to give people a starting point. There is SO MUCH conflicting information about how to set up modules, it's good to have something up-to-date that is proven to work here.

docs/Usage.md
```diff
@@ -0,0 +1,966 @@
...
+# Configure the module target: C++23 standard, link to Vulkan headers, enable C++ standard library module
+target_compile_features( VulkanHppModule PUBLIC cxx_std_23 )
+target_link_libraries( VulkanHppModule PUBLIC Vulkan::Vulkan )
+set_target_properties( VulkanHppModule PROPERTIES
+  CXX_MODULE_STD ON
+)
+
+# link Vulkan C++ module into user project
+add_executable( YourProject main.cpp Engine.cpp ... )
+target_link_libraries( YourProject PRIVATE VulkanHppModule )
```
M2-TE:
```suggestion
# this UUID is still valid as of CMake 4.2.1
# remove this variable if you are setting the UUID via commandline
set( CMAKE_EXPERIMENTAL_CXX_IMPORT_STD d0edc3af-4c50-42ea-a356-e2862fe7a444 )

# UUID had to be set before setting up the project
project( vulkan_hpp_modules_example LANGUAGES CXX )

# find Vulkan SDK (modules were first made available in 1.3.256)
find_package( Vulkan 1.3.256 REQUIRED )

# set up Vulkan C++ module as a static library
add_library( Vulkan-HppModule STATIC )
add_library( Vulkan::HppModule ALIAS Vulkan-HppModule )
target_sources( Vulkan-HppModule PRIVATE
  FILE_SET CXX_MODULES
  TYPE CXX_MODULES
  BASE_DIRS ${Vulkan_INCLUDE_DIR}
  FILES
    ${Vulkan_INCLUDE_DIR}/vulkan/vulkan.cppm
    ${Vulkan_INCLUDE_DIR}/vulkan/vulkan_video.cppm
)

# configure the module target
target_compile_features( Vulkan-HppModule
  PRIVATE cxx_std_23 # (currently) requires C++23 for import std
  INTERFACE cxx_std_20 # can still be consumed as only a C++20 module
)
if( TRUE ) # if you want to use the dynamic dispatcher
  target_link_libraries( Vulkan-HppModule PUBLIC Vulkan::Headers )
  target_compile_definitions( Vulkan-HppModule PUBLIC
    VULKAN_HPP_DISPATCH_LOADER_DYNAMIC
  )
else() # otherwise, use Vulkan::Vulkan to link to vulkan-1
  target_link_libraries( Vulkan-HppModule PUBLIC Vulkan::Vulkan )
endif()
set_target_properties( Vulkan-HppModule PROPERTIES
  CXX_SCAN_FOR_MODULES ON
  CXX_MODULE_STD ON
)

# link Vulkan C++ module into user project
add_executable( YourProject main.cpp Engine.cpp ... )
target_link_libraries( YourProject PRIVATE Vulkan::HppModule )
```

docs/Usage.md
```diff
@@ -0,0 +1,966 @@
...
+```
+
+Configuring the named module is straightforward; add any required Vulkan-Hpp feature macros listed in [Configuration](./Configuration.md) (or any C macros) to `target_compile_definitions`.
+Ensure the macros are `PUBLIC` so they can be propagated to user code.
+For instance:
+
+```cmake
+# Disable exceptions, disable smart handles, disable constructors for structs
+target_compile_definitions( VulkanHppModule PUBLIC
```
M2-TE:
```suggestion
target_compile_definitions( Vulkan-HppModule PUBLIC
```
To match the module target name

docs/Usage.md
```diff
@@ -0,0 +1,966 @@
...
+)
+# ...
+target_link_libraries( YourProject PRIVATE VulkanHppModule )
+```
+
+Furthermore, you may also prefer linking `VulkanHppModule` to just the `Vulkan::Headers` target with the `PUBLIC` scope instead of `Vulkan::Vulkan`, so that the `vulkan-1` library is not linked in, and the Vulkan headers are available to your consuming project, as detailed further below.
+
+```cmake
+target_link_libraries( VulkanHppModule PUBLIC Vulkan::Headers )
+```
```
M2-TE:
```suggestion
```
This was added in the main example, since it is a very common use case.

docs/Usage.md
```diff
@@ -0,0 +1,966 @@
...
+
+Finally, supply the macro `VULKAN_HPP_DEFAULT_DISPATCH_LOADER_DYNAMIC_STORAGE` exactly once in your source code, just as in the non-module case.
```
M2-TE:
```suggestion
If you are using the dynamic dispatcher, do not forget to supply the macro `VULKAN_HPP_DEFAULT_DISPATCH_LOADER_DYNAMIC_STORAGE` exactly once in your source code, just as in the non-module case.
```
Matching the example

docs/Usage.md
```diff
@@ -0,0 +1,966 @@
...
+In order to have that macro available, include [`vulkan_hpp_macros.hpp`](../vulkan/vulkan_hpp_macros.hpp), a lightweight header providing all Vulkan-Hpp related macros and defines.
+
+<!-- TODO: is this example correct? -->
+```cpp
+import vulkan;
+
+#include <vulkan/vulkan_hpp_macros.hpp>
```
M2-TE:
```suggestion
#include <vulkan/vulkan_hpp_macros.hpp>
import vulkan;
```
includes gotta come before imports, else some compilers.. not naming names.. will get very obscure errors.

docs/Usage.md
```diff
@@ -0,0 +1,966 @@
...
+    VULKAN_HPP_DEFAULT_DISPATCHER.init();
+#endif
+
+  auto appInfo = vk::ApplicationInfo("My App", 1, "My Engine", 1, vk::makeApiVersion(1, 0, 0, 0));
+  // ...
+}
+```
+
+<!-- TODO: this test is gone. @M2-TE? -->
+<!-- An example is provided in [`tests/Cpp20Modules/Cpp20Modules.cpp`](tests/Cpp20Modules/Cpp20Modules.cpp). -->
```
M2-TE: Yup, since most tests now have a module version. This test was previously the only test for modules.

docs/Usage.md
```diff
@@ -0,0 +1,966 @@
...
+.\main.exe
+```
+
+For Clang, run the following:
+
+```shell
+clang++ -std=c++23 <path-to-vulkan-hpp>/vulkan.cppm -precompile -o vulkan.pcm
+clang++ -std=c++23 -fprebuilt-module-path=. main.cpp vulkan.pcm -o main
+./main
+``` -->
```
M2-TE:
```suggestion
```
This is indeed very outdated, we do not need to keep this part.

docs/Usage.md
```diff
@@ -0,0 +1,966 @@
...
+
+More information about module compilation may be found in compiler documentation:
+
+- [Clang](https://clang.llvm.org/docs/StandardCPlusPlusModules.html)
+- [GCC](https://gcc.gnu.org/onlinedocs/gcc/C_002b_002b-Modules.html)
+- [MSVC](https://learn.microsoft.com/en-gb/cpp/cpp/modules-cpp?view=msvc-170)
+
+## Miscellaneous features and notes
+
+### Extension Inspection
```
M2-TE, line 981: It might be worth noting that this very much impacts compilation time.
E.g. Clang in Release mode can take around 2 extra minutes to compile. This is observable in our CI as well.

docs/Usage.md
```diff
@@ -0,0 +1,966 @@
...
+| `VULKAN_HPP_CONSTEXPR_20 std::string getExtensionPromotedTo( std::string const & extension )` | The extension or Vulkan version the given extension is promoted to. |
+| `VULKAN_HPP_CONSTEXPR_20 bool isDeprecatedExtension( std::string const & extension )` | `true` if the given extension is deprecated by some other extension or Vulkan version. |
+| `VULKAN_HPP_CONSTEXPR_20 bool isDeviceExtension( std::string const & extension )` | `true` if the given extension is a device extension. |
+| `VULKAN_HPP_CONSTEXPR_20 bool isInstanceExtension( std::string const & extension )` | `true` if the given extension is an instance extension. |
+| `VULKAN_HPP_CONSTEXPR_20 bool isObsoletedExtension( std::string const & extension )` | `true` if the given extension is obsoleted by some other extension or Vulkan version. |
+| `VULKAN_HPP_CONSTEXPR_20 bool isPromotedExtension( std::string const & extension )` | `true` if the given extension is promoted to some other extension or Vulkan version. |
+
+<!-- TODO: probably need to be updated after recent CI changes. -->
```
M2-TE, line 1015: Yea this section should be rewritten a little. Samples have additional RAII ones and some relevant flags should be mentioned here (cmake stuff to enable samples, use local headers, precompile headers).

Tests should get the same treatment and show a quick example of how to run them, e.g. via 

```bash
ctest -j --output-on-failure --test-dir build -C Release
```

docs/Usage.md
```diff
@@ -0,0 +1,966 @@
...
+## Samples and Tests
+
+When you configure your project using CMake, you can enable SAMPLES_BUILD to add some sample projects to your solution. Most of them are ports from the LunarG samples, but there are some more, like CreateDebugUtilsMessenger, InstanceVersion, PhysicalDeviceDisplayProperties, PhysicalDeviceExtensions, PhysicalDeviceFeatures, PhysicalDeviceGroups, PhysicalDeviceMemoryProperties, PhysicalDeviceProperties, PhysicalDeviceQueueFamilyProperties, and RayTracing. All those samples should just compile and run.
+When you configure your project using CMake, you can enable TESTS_BUILD to add some test projects to your solution. Those tests are just compilation tests and are not required to run.
+
+<!-- TODO: Should we move this to `Configuration.md`? -->
```
M2-TE: I think it's best placed here, rather than in `configuration.md`.
Might be nice to separate these flags better, e.g. group the ones that improve or de..prove? compilation time. Extension inspection should be added here too, with another warning that it very much increases time taken.

docs/Usage.md
```diff
@@ -0,0 +1,966 @@
...
+There are a few configurable options to disable features that may not be required, and thus reduce compile times:
+
+- `VULKAN_HPP_NO_SPACESHIP_OPERATOR`, which removes `operator<=>` for structures (available with C++20)
+- `VULKAN_HPP_NO_TO_STRING`, which removes the various `vk::to_string` functions on enums and bitmasks
+- `VULKAN_HPP_USE_REFLECT`, this one needs to be defined to use the reflection function on structures. It's very slow to compile, though!
+- `VULKAN_HPP_USE_STD_EXPECTED`, this one needs to be defined to use `std::expected` for functions that return a value or a `vk::Result`.
+
+## Strict aliasing
+
+As Vulkan-Hpp often needs to switch between C++ vk-types and corresponding bit-identical C-types using `reinterpret_cast`, it is highly recommended to use the compile option `-fno-strict-aliasing` to prevent potentially breaking compiler optimizations.
```
M2-TE, line 1023: I actually wondered about this quite recently, is it really all that necessary?
Compiler optimizations regarding aliasing can be quite noticable when used right. Are there any documented cases of this leading to breakage that I could take a look at?

README.md
```diff
@@ -1,1133 +1,108 @@
...
+## Overview
+
+Vulkan-Hpp provides header-only C++ bindings for the Vulkan C API to improve the developer experience with Vulkan without introducing run-time CPU costs.
+It adds features like type safety for enumerations and bit-fields, STL container support, exception support, and several varieties of RAII-capable types.
+
+This repository contains the [generators](VulkanHppGenerator.hpp) for Vulkan-Hpp, which accept the [XML specification of Vulkan](https://raw.githubusercontent.com/KhronosGroup/Vulkan-Docs/refs/heads/main/xml/vk.xml) and emit C++ bindings, which are documented further below.
+
+## CI/CD status
+
+<!-- TODO: This table is now broken with the recent changes to CI/CD. -->
```
M2-TE, line 27: Leaving a comment here so I do not forget about this. I'll look into a nicer way to show what exactly we cover.

Might allow us to extend this with information such as compiler versions, C++ version ranges, etc.

docs/Usage.md
```diff
@@ -0,0 +1,966 @@
...
+To get just the base structure, without chaining, the other getter function provided does not need a template argument for the structure to get:
+
+```c++
+// Query just vk::MemoryRequirements2KHR
+vk::MemoryRequirements2KHR memoryRequirements = device.getBufferMemoryRequirements2KHR({});
+```
+
+### Handles and functions
+
+<!-- TODO: maybe move this to ./Handles.md? -->
```
M2-TE: It should still have its own section here in usage I believe. A link to ./Handles.md might be good though?

docs/Usage.md
```diff
@@ -0,0 +1,966 @@
...
+```
+
+### Extensions and per-device function pointers
+
+The default Vulkan loader, `vulkan-1` exposes only the Vulkan core functions and a limited number of extensions.
+To use Vulkan-Hpp with extensions, a mechanism must be provided such that _all_ Vulkan operations have corresponding stubs, or can be dispatched to function pointers.
+Vulkan-Hpp provides a per-function dispatch mechanism where each function representing a Vulkan operation accepts as its last parameter a dispatch class.
+This class must provide a callable type for each Vulkan function that is required.
+
+<!-- TODO: What is this code sample exactly?  -->
```
M2-TE: Isn't it just a duplicate of the following section?
Instead of `VULKAN_HPP_DEFAULT_DISPATCHER` it uses vk::detail::DispatchLoaderDynamic directly, using `dld` and such, rather than `defaultDispatchLoaderDynamic`.

The information about vulkan-1 is still useful though, so perhaps best merged with the dynamic dispatch section below it?

docs/Usage.md
```diff
@@ -0,0 +1,966 @@
...
+
+   - Let Vulkan-Hpp do all the work with `vk::detail::DynamicLoader`:
+
+     ```c++
+     VULKAN_HPP_DEFAULT_DISPATCHER.init();
+     ```
+
+   - Use your own dynamic loader, which needs to provide a templated function `getProcAddress`. Refer to the implementation of `vk::detail::DynamicLoader` in `vulkan.hpp` for an example.
+
+     <!-- TODO: Add a skeleton implementation of `struct YourDynamicLoader`? -->
```
M2-TE, line 651: That would be a nice idea. Probably just a stripped down version from what we have in vulkan.hpp is sufficient.

docs/Usage.md
M2-TE: This matches our own cmake more and covers common use cases with dynamic dispatcher better.
Should be a good copy-paste template for people trying to get module compilation started.

docs/Usage.md
```diff
@@ -0,0 +1,966 @@
...
+#if VULKAN_HPP_DISPATCH_LOADER_DYNAMIC == 1
+VULKAN_HPP_DEFAULT_DISPATCH_LOADER_DYNAMIC_STORAGE
+#endif
+
+auto main(int argc, char* const argv[]) -> int
+{
+#if ( VULKAN_HPP_DISPATCH_LOADER_DYNAMIC == 1 )
+    // initialize minimal set of function pointers
+    VULKAN_HPP_DEFAULT_DISPATCHER.init();
+#endif
```
M2-TE:
```suggestion
VULKAN_HPP_DEFAULT_DISPATCH_LOADER_DYNAMIC_STORAGE

auto main(int argc, char* const argv[]) -> int
{
  // initialize minimal set of function pointers
  VULKAN_HPP_DEFAULT_DISPATCHER.init();
```

Probably best to just use the dynamic dispatcher stuff directly. It's what the example is for, so no need to confuse readers with additional `#ifdef`s.

docs/Usage.md
```diff
@@ -0,0 +1,966 @@
...
+
+  auto appInfo = vk::ApplicationInfo("My App", 1, "My Engine", 1, vk::makeApiVersion(1, 0, 0, 0));
+  // ...
```
M2-TE:
```suggestion
  // ...
  // load function pointers with created instance
  VULKAN_HPP_DEFAULT_DISPATCHER.init(instance);
  // OPTIONAL: load function pointers specifically for a vk::device
  VULKAN_HPP_DEFAULT_DISPATCHER.init(device);
```
Can't hurt to add the other two inits here as well, right?

docs/Usage.md
```diff
@@ -0,0 +1,966 @@
...
+
+[`vulkan_extension_inspection.hpp`](../vulkan/vulkan_extension_inspection.hpp) defines functions to inspect Vulkan extensions on the machine.
+With C++20 and above, some of these functions are marked `constexpr`.
+
+In Vulkan, each extension is identified by a string holding its name.
+Note that there exists a macro with that name for each extension.
+Some functions might provide information that depends on the vulkan version.
+As all functions here operate only on strings, the Vulkan versions are encoded by a string prefixed with "VK_VERSION_", followed by the major and the minor version, separated by an underscore, like this: `VK_VERSION_1_0`.
+
+<!-- TODO: the descriptions in this table need to be reworded a bit, it's a little hard to follow. -->
```
M2-TE, line 995: Maybe just use function names and parameters, then add a separate column for return values?
I'm not sure how much more readable that is exactly, but things like the constexpr macro can be stripped out (add another column with checkmarks for which funcs can be constexpr?).

M2-TE (commented):
This is a really great PR, fantastic work!
I've added some comments/suggestions where relevant.

> and finally revise the building and installation sections

woops, I also added some comments regarding those sections. If it's too early, just mark them as resolved and I'll revisit them once you have finished the section!

sharadhr:
> If it's too early, just mark them as resolved

No worries, I can collapse them and work on your suggestions for the usage doc first. Thanks for being so comprehensive :)

docs/Usage.md
sharadhr: Sadge. I use British English...

docs/Usage.md
sharadhr: I am thinking of rewriting it to account for `import std` rather than removing it altogether.

docs/Usage.md
sharadhr: I was hoping you could provide some input, as the CI changes sort of ran away from me after the first few changes to support modules a few months ago.

docs/Usage.md
M2-TE: I disagree with my original comment as well in retrospect. This would be a good opportunity to show how to build with `import std` in C++20, as CMake does not provide this yet.

docs/Usage.md
sharadhr: I'm going to leave this here for now until the maintainer can have a look.

docs/Usage.md
sharadhr: Added in platform-specific compilation commands; this is now done.

sharadhr:
Ready to review. The RAII guide will be a separate PR to land after this is merged.

M2-TE:
Did you mean to push the `.gitignore`?
I think it would be nice to add one, but I reckon it wasn't meant to show up here.

The `.clang-format` was also likely pushed on accident. Worth adding to the `.gitignore` too ;)

docs/Usage.md
M2-TE: Will be tracking this in #2465

sharadhr:
Ah. Fixed!

README.md
```diff
@@ -1,1133 +1,108 @@
...
+| Platform | Build Status |
+|:--------|:------------|
+| **Windows**  | [![CI Windows Clang](https://github.com/KhronosGroup/Vulkan-Hpp/actions/workflows/ci-windows-clang.yml/badge.svg)](https://github.com/KhronosGroup/Vulkan-Hpp/actions/workflows/ci-windows-clang.yml) <br> [![CI Windows MSVC](https://github.com/KhronosGroup/Vulkan-Hpp/actions/workflows/ci-windows-msvc.yml/badge.svg)](https://github.com/KhronosGroup/Vulkan-Hpp/actions/workflows/ci-windows-msvc.yml) |
+| **Linux**    | [![CI Ubuntu 22](https://github.com/KhronosGroup/Vulkan-Hpp/actions/workflows/ci-ubuntu-22.yml/badge.svg)](https://github.com/KhronosGroup/Vulkan-Hpp/actions/workflows/ci-ubuntu-22.yml) <br> [![CI Ubuntu 24](https://github.com/KhronosGroup/Vulkan-Hpp/actions/workflows/ci-ubuntu-24.yml/badge.svg)](https://github.com/KhronosGroup/Vulkan-Hpp/actions/workflows/ci-ubuntu-24.yml) |
+| **macOS**    | [![CI MacOS](https://github.com/KhronosGroup/Vulkan-Hpp/actions/workflows/ci-macos.yml/badge.svg)](https://github.com/KhronosGroup/Vulkan-Hpp/actions/workflows/ci-macos.yml) |
+
+- [Overview](#overview)
+- [CI/CD status](#cicd-status)
+- [Installation and usage](#installation-and-usage)
```
M2-TE:
```suggestion
- [Installation and usage](#installation-and-usage)
```

Probably don't need to refer to sections that are above this

README.md
```diff
@@ -1,1133 +1,108 @@
...
+We try to keep the API for all flavours of Vulkan-Hpp constant or backwards compatible. However, we may introduce unavoidable breaking changes, usually to fix bugs. Following is a list of those changes, arranged by version.
+
+### v1.4.334
+
+The `vulkan_hpp` C++ named module has been renamed to `vulkan`.
+
+### v1.4.324
+
+With PR [#2226](https://github.com/KhronosGroup/Vulkan-Hpp/pull/2226), the return type of `vk::raii::Device::acquireNextImage2KHR` and `vk::raii::SwapchainKHR::acquireNextImage` has changed from `std::pair<vk::Result,uint32_t>` to the equivalent `vk::ResultValue<uint32_t>`
+
```
M2-TE:
```suggestion

### v1.4.329

With PR [#2303](https://github.com/KhronosGroup/Vulkan-Hpp/pull/2303), `import std` was made mandatory when using the named module.

```

Happened a while back, but should still be of interest to some readers

README.md
```diff
@@ -1,1133 +1,108 @@
...
+## Installation and usage
+
+### Vulkan SDK
+
+Vulkan-Hpp has been part of the [LunarG Vulkan SDK](https://www.lunarg.com/vulkan-sdk/) since version 1.0.24; this remains the recommended installation method.
+If you need a more recent version than the SDK supports, then Vulkan-Hpp is also distributed as part of the [Khronos Group Vulkan-Headers repository](https://github.com/KhronosGroup/Vulkan-Headers).
+
+### vcpkg
+
+As above, Vulkan-Hpp is provided in the [`vulkan-headers` port](https://vcpkg.io/en/package/vulkan-headers) in [vcpkg](https://github.com/Microsoft/vcpkg).
```
M2-TE:
```suggestion
As above, Vulkan-Hpp is provided in the [`vulkan-headers`](https://vcpkg.io/en/package/vulkan-headers) port in [vcpkg](https://github.com/Microsoft/vcpkg).
```

The formatting looks odd when combining the package name in backticks and another word without

README.md
```diff
@@ -1,1133 +1,108 @@
...
+Otherwise, you can also install the [`vulkan-sdk-components` port](https://vcpkg.io/en/package/vulkan-sdk-components), which is a vcpkg equivalent to the LunarG-distributed Vulkan SDK.
```
M2-TE:
```suggestion
Otherwise, you can also install the [`vulkan-sdk-components`](https://vcpkg.io/en/package/vulkan-sdk-components) port, which is a vcpkg equivalent to the LunarG-distributed Vulkan SDK.
```

README.md
```diff
@@ -1,1133 +1,108 @@
...
+./bootstrap-vcpkg.sh
+./vcpkg integrate install
+./vcpkg install vulkan-headers
+```
+
+For support, [create an issue or pull request](https://github.com/Microsoft/vcpkg) on the vcpkg repository.
+
+### Conan
+
+Similar to vcpkg, Vulkan-Hpp can be installed as part of the [`vulkan-headers` recipe](https://conan.io/center/recipes/vulkan-headers).
```
M2-TE:
```suggestion
Similar to vcpkg, Vulkan-Hpp can be installed as part of the [`vulkan-headers`](https://conan.io/center/recipes/vulkan-headers) recipe.
```

README.md
```diff
@@ -1,1133 +1,108 @@
...
+
+### v1.4.334
+
+The `vulkan_hpp` C++ named module has been renamed to `vulkan`.
+
+### v1.4.324
+
+With PR [#2226](https://github.com/KhronosGroup/Vulkan-Hpp/pull/2226), the return type of `vk::raii::Device::acquireNextImage2KHR` and `vk::raii::SwapchainKHR::acquireNextImage` has changed from `std::pair<vk::Result,uint32_t>` to the equivalent `vk::ResultValue<uint32_t>`
+
+## Documentation
```
M2-TE: The documentation section can probably be placed higher. It is quite concise and is often exactly what people are looking for.

Perhaps place it above the `installation and usage`section?

docs/Building.md
```diff
@@ -0,0 +1,53 @@
+# Building
+
+## Minimum requirements
+
+The generator requires a toolchain that supports C++20; the samples and tests may be compiled with C++11. The following compilers are known to work, and tested during releases:
```
M2-TE:
```suggestion
The generator requires a toolchain that supports C++20; the samples and tests may be compiled using at least C++11 with higher versions enabling wider test coverage at up to C++23. The following compilers are known to work, and tested during releases:
```

To let people know that there is a benefit to running tests with higher C++ versions

README.md
M2-TE, line 87:
```suggestion

### v1.4.329

With PR [#2303](https://github.com/KhronosGroup/Vulkan-Hpp/pull/2303), `import std` was made mandatory when using the C++ named module.

```

docs/Building.md
```diff
@@ -0,0 +1,53 @@
...
+   - Additional dependencies are required to build the samples on Linux.
+     On Ubuntu, install them with:
+
+     ```shell
+     sudo apt update && sudo apt install install lsb-release gpg software-properties-common wget
+     ```
+
+2. In a shell, clone the repository:
+
+    ```git clone --recurse-submodules https://github.com/KhronosGroup/Vulkan-Hpp.git```
```
M2-TE:
```suggestion
    ```shell
    git clone --recurse-submodules https://github.com/KhronosGroup/Vulkan-Hpp.git
    ```
```

This makes the field copy-able via click for some convenience

docs/Building.md
```diff
@@ -0,0 +1,53 @@
...
+3. Change the shell's working directory to the newly created `Vulkan-Hpp` directory.
```
M2-TE, line 32:
```suggestion
3. Change the shell's working directory to the newly created `Vulkan-Hpp` directory.
    ```shell
    cd Vulkan-Hpp
    ```
```
Just for some additional convenience

docs/Building.md
```diff
@@ -0,0 +1,53 @@
...
+4. Create a build environment with CMake:
+    ```cmake -DVULKAN_HPP_SAMPLES_BUILD=ON -DVULKAN_HPP_TESTS_BUILD=ON -B build```
```
M2-TE:
```suggestion
    ```shell
    cmake -B build --preset samples-tests
    ```
```

Might as well use the presets here, its nice and concise

docs/Building.md
```diff
@@ -0,0 +1,53 @@
...
+
+    You may need to specify a generator via `-G`; for a full list of generators execute `cmake -G`.
+
+    To regenerate `vulkan.hpp` from the `vk.xml` XML registry file, add the
+       `-DVULKAN_HPP_RUN_GENERATOR=ON` option to the CMake command line.
```
M2-TE:
```suggestion
    To compile samples and tests using freshly generated headers from the `vk.xml` XML registry:
    ```shell
    cmake -B build --preset generator-samples-tests
    ```
```

docs/Building.md
```diff
@@ -0,0 +1,53 @@
...
+5. The generated project may be opened with an IDE like Visual Studio or CLion, or the various targets may be built directly from a shell:
+
+    ```cmake --build build --parallel```
```
M2-TE:
```suggestion
5. To actually start building, run the following:
    ```shell
    cmake --build build --parallel
    ```
    The tests are run via:
    ```shell
    ctest -j --output-on-failure --test-dir build
    ```
```

I don't think IDEs should be mentioned here, since we have been doing all this via CLI, so that might be a bit confusing. Most IDEs will just do all this stuff automatically and users only have to select a preset.

docs/Configuration.md
```diff
@@ -0,0 +1,182 @@
...
+You can use your own default dispatcher by setting `VULKAN_HPP_DEFAULT_DISPATCHER` to an object that provides the same API. If you explicitly set `VULKAN_HPP_DEFAULT_DISPATCHER`, you need to set `VULKAN_HPP_DEFAULT_DISPATCHER_TYPE` accordingly as well.
+
+## `VULKAN_HPP_DEFAULT_DISPATCHER_TYPE`
+
+This names the default dispatcher type, as specified by `VULKAN_HPP_DEFAULT_DISPATCHER`. Per default, it is `VULKAN_HPP_DISPATCH_LOADER_DYNAMIC_TYPE` or `VULKAN_HPP_DISPATCH_LOADER_STATIC_TYPE`, depending on `VULKAN_HPP_DISPATCH_LOADER_DYNAMIC` being `1` or not `1`, respectively. If you explicitly set `VULKAN_HPP_DEFAULT_DISPATCHER`, you need to set `VULKAN_HPP_DEFAULT_DISPATCHER_TYPE` accordingly as well.
+
+## `VULKAN_HPP_DEFAULT_DISPATCH_LOADER_DYNAMIC_STORAGE`
+
+If you have not defined your own `VULKAN_HPP_DEFAULT_DISPATCHER`, and have `VULKAN_HPP_DISPATCH_LOADER_DYNAMIC` defined to be `1` (the default), you need to have the macro `VULKAN_HPP_DEFAULT_DISPATCH_LOADER_DYNAMIC_STORAGE` exactly once in any of your source files to provide storage for that default dispatcher. `VULKAN_HPP_STORAGE_API` then controls the import/export status of that default dispatcher.
```
M2-TE, line 58:
```suggestion
If you have not defined your own `VULKAN_HPP_DEFAULT_DISPATCHER`, and have `VULKAN_HPP_DISPATCH_LOADER_DYNAMIC` defined to be `1` (the default), you need to have the macro `VULKAN_HPP_DEFAULT_DISPATCH_LOADER_DYNAMIC_STORAGE` exactly once in any of your source files to provide storage for that default dispatcher. `VULKAN_HPP_STORAGE_API` then controls the import/export status of that default dispatcher. Note that the C++ named module does not need to declare additional storage this way.
```

docs/Usage.md
```diff
@@ -0,0 +1,982 @@
...
+<!-- TODO: Should we move this to `Configuration.md`? -->
+## Compile time issues
+
+As mentioned above, Vulkan-Hpp is a large header, and compilation times may increase drastically.
+There are a few configurable options to disable features that may not be required, and thus reduce compile times:
+
+- `VULKAN_HPP_NO_SPACESHIP_OPERATOR`, which removes `operator<=>` for structures (available with C++20)
+- `VULKAN_HPP_NO_TO_STRING`, which removes the various `vk::to_string` functions on enums and bitmasks
+- `VULKAN_HPP_USE_REFLECT`, this one needs to be defined to use the reflection function on structures. It's very slow to compile, though!
+- `VULKAN_HPP_USE_STD_EXPECTED`, this one needs to be defined to use `std::expected` for functions that return a value or a `vk::Result`.
```
M2-TE: How is this related to compile time issues?

docs/Usage.md
```diff
@@ -0,0 +1,982 @@
...
+> Structured bindings and if-initialiser statements requires a compiler supporting at least C++17.
+
+#### `std::expected`
+
+<!-- TODO: Need an example. Perhaps use the same shader module. -->
+>[!NOTE]
+> This feature requires a compiler and standard library supporting at least C++23, and must be explicitly **enabled** with the `VULKAN_HPP_USE_STD_EXPECTED` macro.
+
+When `std::expected` is available, `vk::ResultValue<T>::type` is defined as `std::expected<T, vk::Result>`.
+The result can then be monadically handled with `and_then`, `transform`, and other member functions of `std::expected`.
```
M2-TE, line 576: Needs an example still!
(please, I might want to use std::expected myself)

docs/Usage.md
```diff
@@ -0,0 +1,982 @@
...
+  target_link_libraries( Vulkan-HppModule PUBLIC Vulkan::Vulkan )
+endif()
+set_target_properties( Vulkan-HppModule PROPERTIES
+  CXX_SCAN_FOR_MODULES ON
+  CXX_MODULE_STD ON
+)
+
+# link Vulkan C++ module into user project
+add_executable( YourProject main.cpp ... )
+target_link_libraries( YourProject PRIVATE Vulkan-HppModule )
```
M2-TE:
```suggestion
cmake_minimum_required( VERSION 3.30 )

# this UUID is still valid as of CMake 4.2.1
set( CMAKE_EXPERIMENTAL_CXX_IMPORT_STD d0edc3af-4c50-42ea-a356-e2862fe7a444 ) # remove this line if you are setting the UUID at the command-line or in a preset

# UUID has to be set before setting up the project
project( vulkan_hpp_modules_example LANGUAGES CXX )

# While modules were first made available in 1.3.256,
# they have changed drastically since then and this example assumes version 1.4.344
find_package( Vulkan 1.4.344 QUIET )
if ( Vulkan_FOUND )
    # set up Vulkan C++ module as a static library
    add_library( Vulkan-HppModule STATIC )
    add_library( Vulkan::HppModule ALIAS Vulkan-HppModule )
    target_sources( Vulkan-HppModule PUBLIC
        FILE_SET vulkan_modules
        TYPE CXX_MODULES
        BASE_DIRS ${Vulkan_INCLUDE_DIR}
        FILES
            ${Vulkan_INCLUDE_DIR}/vulkan/vulkan.cppm
            ${Vulkan_INCLUDE_DIR}/vulkan/vulkan_video.cppm)
else()
    # when you do not have the right version installed, just fetch the headers directly
    include(FetchContent)
    FetchContent_Declare(vulkan-headers
        GIT_REPOSITORY "https://github.com/KhronosGroup/Vulkan-Headers.git"
        GIT_TAG "v1.4.344"
        GIT_SHALLOW ON
        OVERRIDE_FIND_PACKAGE
        SYSTEM)
    FetchContent_MakeAvailable(vulkan-headers)
    # make sure to set up the include path for CMake
    set(Vulkan_INCLUDE_DIR "${vulkan-headers_SOURCE_DIR}/include")
endif()

# configure the module target (import std currently requires C++23 with CMake)
target_compile_features( Vulkan-HppModule PUBLIC cxx_std_23 )
set_target_properties( Vulkan-HppModule PROPERTIES CXX_MODULE_STD ON )

if( TRUE ) # if you want to use the dynamic dispatcher
    target_link_libraries( Vulkan-HppModule PUBLIC Vulkan::Headers )
    target_compile_definitions( Vulkan-HppModule PUBLIC VULKAN_HPP_DISPATCH_LOADER_DYNAMIC )
else() # otherwise, use Vulkan::Vulkan to link to vulkan-1
    target_link_libraries( Vulkan-HppModule PUBLIC Vulkan::Vulkan )
endif()

# link Vulkan C++ module into user project
add_executable( YourProject main.cpp )
target_link_libraries( YourProject PRIVATE Vulkan-HppModule )
```

this assumes the use of Vulkan-Headers v1.4.344, which will probably release around next week and will hopefully include the ABI-breaking PR

docs/Usage.md
```diff
@@ -0,0 +1,982 @@
...
+  // initialize minimal set of function pointers
+  VULKAN_HPP_DEFAULT_DISPATCHER.init();
+
+  auto appInfo = vk::ApplicationInfo("My App", 1, "My Engine", 1, vk::makeApiVersion(1, 0, 0, 0));
+  // ...
+  // load function pointers with created instance
+  VULKAN_HPP_DEFAULT_DISPATCHER.init(instance);
+  // OPTIONAL: load function pointers specifically for a vk::device
+  VULKAN_HPP_DEFAULT_DISPATCHER.init(device);
+}
```
M2-TE:
```suggestion
// optional include for VULKAN_HPP_DEFAULT_DISPATCHER
// otherwise, use vk::detail::defaultDispatchLoaderDynamic
#include <vulkan/vulkan_hpp_macros.hpp>
import vulkan;

auto main(int argc, char* const argv[]) -> int
{
    // initialize minimal set of function pointers
    VULKAN_HPP_DEFAULT_DISPATCHER.init();

    auto appInfo = vk::ApplicationInfo("My App", 1, "My Engine", 1, vk::makeApiVersion(1, 0, 0, 0));

    vk::Instance instance;
    // create vk::Instance instance here

    // load function pointers with created instance
    VULKAN_HPP_DEFAULT_DISPATCHER.init(instance);

    vk::Device device;
    // create vk::Device here

    // OPTIONAL: load function pointers for a specific vk::Device
    VULKAN_HPP_DEFAULT_DISPATCHER.init(device);
}
```

This better points out that insance/device need to be created first. It also assumes the ABI-breaking PR (same as the CMake change above).

docs/Usage.md
M2-TE: Should we just refer to the Build.md section here?

README.md
sharadhr: This was autogenerated. I can remove it, but I think if I ever edit the readme again, it'll come back...

M2-TE:
Leaving a tracking comment here to make sure to wait until v1.4.344 release to test the module example with the ABI breaking changes in #2443.

docs/Usage.md
sharadhr: Added an example below.

docs/Usage.md
sharadhr: Maybe, but I think at least 1-2 sentences on how it works would be good.

asuessenbach:
Wow!

I tried hard, but could not spot anything to improve!

After you've resolved the conflict, I'll merge this.

sharadhr:
@asuessenbach there are a few TODOs though; hope you can think of some solutions for them. (You may need to revert to the old GitHub suggestions view; the new one hides all the comments).

docs/Usage.md
```diff
@@ -0,0 +1,1006 @@
...
+Instead, Vulkan-Hpp provides two class templates, `vk::ArrayProxy<T>` and `vk::ArrayProxyNotTemporaries<T>`, which generalise the following into a single parameter:
+
+- empty arrays;
+- a single value;
+- STL containers including `std::initializer_list`, `std::array`, `std::span` and `std::vector`;
+- C-style arrays;
+- pointers with counts.
+
+<!-- TODO: uncomment with more detail. Fill in when `ArrayProxy` vs `ArrayProxyNoTemporaries` is typically expected; function out-parameters? Ask @asuessenbach -->
+<!-- Note that `vk::ArrayProxyNoTemporaries<T>` **disallows** temporary values, which may be important for out-parameters. -->
```
asuessenbach: Well, `ArrayProxyNoTemporaries` is designed to be used with struct constructors.
While functions that accept a data array do not need to worry about the lifetime of that array (it is guaranteed to exist at least until the end of that function), structures do need to do so.
For a structure that holds a count and a pointer to some data, it is assumed that that pointer is valid even after construction time of that structure. That is, you should not pass a temporary array into such a constructor. That's why struct constructors take an ArrayProxyNoTemporaries, while functions take an ArrayProxy.

docs/Usage.md
```diff
@@ -0,0 +1,1006 @@
...
+Vulkan-Hpp provides a constructor similar to the `CreateInfo`, which accepts a list of all structures part of the chain.
+The `pNext` field of each structure is populated automatically:
+
+```cpp
+vk::StructureChain<vk::MemoryAllocateInfo, vk::MemoryDedicatedAllocateInfo> c = {
+  vk::MemoryAllocateInfo(size, type),
+  vk::MemoryDedicatedAllocateInfo(image)
+};
+```
+<!-- TODO: need an example here. Also more info and examples: what about the same type, repeated? I don't use structure chains. @asuessenbach? -->
```
asuessenbach: Ok, here's a simple example:
```
vk::StructureChain<vk::InstanceCreateInfo, vk::DebugUtilsMessengerCreateInfoEXT, vk::ValidationFlagsEXT> instanceCreateInfoChain = {
  { .pApplicationInfo = &applicationInfo },
  { .messageSeverity = vk::ebugUtilsMessageSeverityFlagBitsEXT::eError },
  { .disabledValidationCheckCount = 1, .pDisabledValidationChecks = &check }
};

if ( dontUseValidationFlags )
{
  instanceCreateInfoChain.unlink<vk::ValidationFlagsEXT>();
}

vk::Instance = vk::createInstance( instanceCreateInfoChain.get<vk::InstanceCreateInfo>() );
```

asuessenbach: On the same type, repeated:
Some structs are allowed to occur multiple times in a structure chain, like `vk::LayerSettingsCreateInfoEXT`. They are marked `with allowduplicate="true"` in vk.xml.
Such structures are allowed to be multiple times in a `vk::StructureChain` as well, while others are not. That's also checked at compile time.
Functions like `vk::StructureChain::get` have a template value parameter called `Which`, that allows to address a specific element of the specified type in the chain.

docs/Usage.md
```diff
@@ -0,0 +1,1006 @@
...
+
+### Error handling
+
+Most Vulkan operations return a `VkResult` type, indicating whether the operation was successful, and if not, what kind of error occurred.
+This is mapped to `vk::Result` in Vulkan-Hpp.
+Vulkan-Hpp provides several ways to handle these result codes.
+There are two class templates, `vk::ResultValue<T>` and `vk::ResultValueType<T>`, which wrap return values along with their corresponding `vk::Result` codes.
+These class templates may be defined differently depending on the error-handling method chosen.
+
+<!-- TODO: Explain the class templates in greater detail. @asuessenbach? -->
```
asuessenbach: A `vk::ResultValue` unconditionally consists of a `result` and a `value`.
It is returned by functions that gets you something and have more than one success code, like `vk::Device::createGraphicsPipeline`.

A `vk::ResultValueType` is returned by functions that gets you something and have just one success code, like `vk::createInstance`. 
If `VULKAN_HPP_NO_EXCEPTIONS` is defined, it the type actually is a `vk::ResultValue`, as it might return as well an error code.

docs/Usage.md
asuessenbach: @M2-TE is right, it's a duplicate of the next section.
And as using the VULKAN_HPP_DEFAULT_DISPATCHER macro is probably the "better" way to encode this, we should skip the code section here and merge the two text paragraphs into one.

docs/Usage.md
asuessenbach: Well, yes. I think, having your own dynamic loader is a pretty advanced movement. But showing a stripped down version of the `vk::detail::DynamicLoader` might be a good fit here.

docs/Usage.md
asuessenbach: If I remember correctly, there was some company having their own (private) graphics library as a dll, based on Vulkan-Hpp. And the actual user of that dll doesn't need to know about that.

asuessenbach (commented):
Indeed, a couple of Todos I didn't see before.
I tried to answer some of them, others (like the one in the Readme.md) just needs to be done.

asuessenbach:
@sharadhr and @M2-TE, while we're working on the documentation of Vulkan-Hpp: I'm thinking about completely removing those `vk::UniqueHandles`. I think, they could be easily replaced by the more versatile `vk::raii::Handles`.

I imagine to have some words about removing them in the Readme, telling that they're gonna be marked as deprecated in about a year, and finally be removed in about another year.

What would you think about that?

sharadhr:
I think this is something that needs to be discussed with the wider community. I think maybe this should be in its own issue and PR?

There is overlap, but I'm not sure I can speak for everyone about removing `vk::UniqueHandle`.

M2-TE:
I agree with @sharadhr as in that should probably be discussed in a separate PR.

Off the top of my head, one thing to discuss might be the dispatchers; `vk::UniqueHandles` use the static/dynamic dispatchers, whereas `vk::raii::Handles` have their own from what I've seen. So, unique handles could easily be mixed with normal handles, whereas raii handles likely have duplicate/separate dispatch storage?

I've used neither myself, so this is a question that probably needs a little more community reach like on the Vulkan Discord or Reddit.

docs/Usage.md
sharadhr: Thanks for this; having this explicitly written down in the readme makes it much clearer.

sharadhr:
OK—I think this is ready for final review and merge. 

I don't want any more scope creep and hope to address the rest of the readme in phases and separate PRs.

docs/Usage.md
```diff
@@ -601,22 +626,7 @@ To use Vulkan-Hpp with extensions, a mechanism must be provided such that _all_
...
-
-//  Previously-created `VkDevice` and optionally a function pointer resolving `vkGetDeviceProcAddr`, all functions are fetched as well, but now device-specific functions are fetched via `vkDeviceGetProcAddr`.
-vk::detail::DispatchLoaderDynamic dldid(instance, getInstanceProcAddr, device);
-
-// Pass dispatch class to function call as last parameter
-device.getQueue(graphics_queue_family_index, 0, &graphics_queue, dldid);
-```
-
-Vulkan-Hpp provides a default dispatch loader, `vk::detail::DispatchLoaderDynamic`, which fetches all function pointers known to the library.
+Vulkan-Hpp provides a default dispatch loader, `vk::detail::DispatchLoaderDynamic`, which fetches _all_ function pointers known to the library.
```
M2-TE: The default dispatch loader is the static one, the dynamic dispatcher has to be enabled via flags. Still, mentioning that static dispatcher only dispatches a few main functions and dynamic dispatches all of them is good here.

M2-TE:
I don't want to be overly picky, but you have this line in your latest commit about the static dispatcher:

> Vulkan-Hpp provides a static dispatch loader, `vk::detail::DispatchLoaderStatic`, which contains stubs for all functions known to the library. This is the default, and a loader library such as [volk](https://github.com/zeux/volk) can be used to populate the stubs with function pointers

The last part `populate the stubs with function pointers` is not entirely accurate. The functions, when using the static dispatcher, are simply declared (`vulkan_core.h`) and just never defined. The static dispatcher simply short-circuits to these.

So, for the static dispatcher, the function declarations need to have function definitions somewhere (some select few provided by vulkan-1, all provided by volk for example). This is why you can get linking errors with vulkan-1.

The dynamic dispatcher however does indeed use proper function pointers, so that is correct. When using it, all the function declarations are never declared in `vulkan_core.h` in the first place (`VK_NO_PROTOTYPES`).

M2-TE:
Perfect, I see nothing else to comment on!

asuessenbach:
What a great work!
The readme file is now no longer just a “readme file,” but documentation.

Thanks a lot for this great contribution.

PR #2362, opened 2025-11-14
Fix typo leading to module compilation error

YaaZ:
This typo breaks module compilation.
Unfortunately, it already got into Vulkan-Headers 1.4.333.
If you are interested, take a look at the [CI I set up](https://github.com/YaaZ/VulkanMemoryAllocator-Hpp/actions/runs/19361567816/job/55394558090?pr=66) to test the modules - it works with MSVC on Windows and LLVM/Clang+libc on Linux & macOS.


sharadhr:
The CI setup is really nice; maybe we can use it almost-wholesale.

More importantly, we should find a way to parametrise even the hardcoded stuff; having typos like this making it into point releases is a bit of an issue.

PR #2373, opened 2025-11-26
Rename `vulkan_hpp` module to `vulkan`

### LINKED ISSUE #2311: C++20 module name

sharadhr:
Resolves #2311.


asuessenbach:
Thanks a lot for your contribution, again!

sharadhr:
Oh no, could you revert this? I hadn't renamed the module in Vulkan-Video yet! I was planning to edit that in after #2372 was merged.

asuessenbach:
No panic, I've changed that.

sharadhr:
Sorry, has this been done? I still see `vulkan_hpp:video` in the snippet:

https://github.com/KhronosGroup/Vulkan-Hpp/blob/cbf8b5815429d4a7a32e5d511d134964ffaddce0/snippets/VideoCppmTemplate.hpp#L20

sharadhr:
@asuessenbach bump; this has broken the 1.4.335 release as well.

asuessenbach:
> Sorry, has this been done? I still see vulkan_hpp:video in the snippet:

@sharadhr 
Bummer!
I've missed that as well!
Would you mind to craft a little PR to fix that?

sharadhr:
@asuessenbach resolved in #2377.

PR #2377, opened 2025-12-01
Fix `vulkan:video` module by correctly removing suffix

sharadhr:
Corrects a bug in #2373.

### DIFF
--- a/snippets/VideoCppmTemplate.hpp
+++ b/snippets/VideoCppmTemplate.hpp
@@ -17,7 +17,7 @@ module;
 
 #include <vulkan/vulkan_video.hpp>
 
-export module vulkan_hpp:video;
+export module vulkan:video;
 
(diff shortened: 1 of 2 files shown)


asuessenbach:

stripe2933:
This change is not applied to Vulkan SDK 1.4.335 and can't compile `vulkan.cppm` again... 🤯

sharadhr:
> This change is not applied to Vulkan SDK 1.4.335 and can't compile `vulkan.cppm` again... 🤯

Yeah... Is there a way we can ask KhronosGroup to do a point release, @asuessenbach?

asuessenbach:
> This change is not applied to Vulkan SDK 1.4.335

That's right, it was merged a few days after #2374. And thus did not make it into the 1.4.335 release.
I'll ask at Khronos for a point relase, but please don't expect it to happen.

... that is... it's not Khronos, but LunarG, that releases the SDKs. And I don't have any contact to that company.

PR #2383, opened 2025-12-04
Add support for Homebrew installations on Apple Silicon

chizkiyahu:


asuessenbach:
Wow, that's the seventh try on apple to load the vulkan dll!

Is there an apple-guru out there who can confirm the validity of this path, and its position between "/usr/local/lib/libvulkan.dylib" and "libMoltenVK.dylib"?

chizkiyahu:
> Wow, that's the seventh try on apple to load the vulkan dll!
> 
> Is there an apple-guru out there who can confirm the validity of this path, and its position between "/usr/local/lib/libvulkan.dylib" and "libMoltenVK.dylib"?

please check this 
https://github.com/arm/ai-ml-sdk-for-vulkan/blob/672db3db8ad256cc3038f414a8cfa042099dce21/.github/workflows/build-with-manifest.yml#L78-L81

without the last line its fail to load 

the link list is 
 /opt/homebrew/lib/libvulkan.dylib -> ../Cellar/vulkan-loader/1.4.328.1/lib/libvulkan.dylib
 /opt/homebrew/Cellar/vulkan-loader/1.4.328.1/lib/libvulkan.dylib -> libvulkan.1.dylib
 /opt/homebrew/Cellar/vulkan-loader/1.4.328.1/lib/libvulkan.1.dylib -> libvulkan.1.4.328.dylib
 /opt/homebrew/Cellar/vulkan-loader/1.4.328.1/lib/libvulkan.1.4.328.dylib - is a file 
 
 /opt/homebrew/Cellar/molten-vk/1.4.0/lib/libMoltenVK.dylib - is a file
 
 the file 
 /opt/homebrew/Cellar/molten-vk/1.4.0/etc/vulkan/icd.d/MoltenVK_icd.json
 ```json
 {
    "file_format_version" : "1.0.0",
    "ICD": {
        "library_path": "../../../lib/libMoltenVK.dylib",
        "api_version" : "1.4.0",
        "is_portability_driver" : true
    }
}

 ```

sharadhr:
It feels like we could refactor this into a loop, or better yet, a `std::find_if`.

asuessenbach:
@chizkiyahu would you like to adjust this PR to what @sharadhr suggested above?

chizkiyahu:
@asuessenbach @sharadhr 
I push a change that use for loop 
1. I needs help with `clang-format` any version that I try make a big change in the files (more than I change)
2. why we needs the check of env `DYLD_FALLBACK_LIBRARY_PATH` - can I remove this code ? and make it simpler

asuessenbach:
> why we needs the check of env DYLD_FALLBACK_LIBRARY_PATH - can I remove this code ? and make it simpler

I actually don't know.
The apple-specific loading sequence is inspired by that in volk. Which actually doesn't hold anything with "homebrew" in their sequence. Which raises the question of whether this change is necessary at all.

chizkiyahu:
> > why we needs the check of env DYLD_FALLBACK_LIBRARY_PATH - can I remove this code ? and make it simpler
> 
> I actually don't know. The apple-specific loading sequence is inspired by that in volk. Which actually doesn't hold anything with "homebrew" in their sequence. Which raises the question of whether this change is necessary at all.


Given Homebrew's shift to the sandboxed `/opt/homebrew/lib` environment, implementing this change would significantly simplify the experience for the user. I see no potential risk or harm in doing so; it would simply resolve issues for users who are currently facing linking failures.

asuessenbach:
I'm so sorry to say that, but it seems the sequence of if-clauses is easier to read and follow than the loop! (I never thought I would write something like this!!) It's probably due to that `DYLD_FALLBACK_LIBRARY_PATH`, that's seems to be needed, but just for one of those cases.

That is, would be great if you could revert to that if-clause sequence.
I assume, all those possible vulkan libraries are mutually exclusive? Maybe there is another way to decide which DLL should be loaded? Some define or whatnot?
And, finally, would passing in the actual `vulkanLibraryName` into the `DynamicLoader` be an option?

chizkiyahu:
> That is, it would be great if you could revert to that if-clause sequence.

Done.

> I assume all those possible Vulkan libraries are mutually exclusive? Maybe there is another way to decide which DLL should be loaded? Some define or similar? And finally, would passing the actual `vulkanLibraryName` into the `DynamicLoader` be an option?

I assume this is possible, but any change like this could break existing dependencies, and I do not have a test suite to verify it safely.
I am also not sure why this code exists in its current form.

@davidegrohmann

asuessenbach:
And for your patience with my back and forth on how to encode it :)

PR #2386, opened 2025-12-08
Make allocator arguments const-ref

### LINKED ISSUE #2315: const-ref for optional allocator arguments?
In this issue: https://github.com/KhronosGroup/Vulkan-Hpp/issues/237, it was requested to have an optional `Allocator` argument for various functions that allocate `std::vector`s.

This works, but the argument for the allocator is taken by reference, not by const-reference. For instance, the method for allocating descriptor sets is:

```cpp
  // wrapper function for command vkAllocateDescriptorSets, see https://registry.khronos.org/vulkan/specs/latest/man/html/vkAllocateDescriptorSets.html
  template <typename DescriptorSetAllocator,
            typename Dispatch,
            typename std::enable_if<std::is_same<typename DescriptorSetAllocator::value_type, VULKAN_HPP_NAMESPACE::DescriptorSet>::value, int>::type>
  VULKAN_HPP_NODISCARD VULKAN_HPP_INLINE typename ResultValueType<std::vector<VULKAN_HPP_NAMESPACE::DescriptorSet, DescriptorSetAllocator>>::type
                       Device::allocateDescriptorSets( const VULKAN_HPP_NAMESPACE::DescriptorSetAllocateInfo & allocateInfo,
                                    DescriptorSetAllocator &                                descriptorSetAllocator,
                                    Dispatch const &                                        d ) const
  {
    VULKAN_HPP_ASSERT( d.getVkHeaderVersion() == VK_HEADER_VERSION );
#  if ( VULKAN_HPP_DISPATCH_LOADER_DYNAMIC == 1 )
    VULKAN_HPP_ASSERT( d.vkAllocateDescriptorSets && "Function <vkAllocateDescriptorSets> requires <VK_VERSION_1_0>" );
#  endif

    std::vector<VULKAN_HPP_NAMESPACE::DescriptorSet, DescriptorSetAllocator> descriptorSets( allocateInfo.descriptorSetCount, descriptorSetAllocator );
    VULKAN_HPP_NAMESPACE::Result                                             result = static_cast<VULKAN_HPP_NAMESPACE::Result>( d.vkAllocateDescriptorSets(
      m_device, reinterpret_cast<const VkDescriptorSetAllocateInfo *>( &allocateInfo ), reinterpret_cast<VkDescriptorSet *>( descriptorSets.data() ) ) );
    VULKAN_HPP_NAMESPACE::detail::resultCheck( result, VULKAN_HPP_NAMESPACE_STRING "::Device::allocateDescriptorSets" );

    return VULKAN_HPP_NAMESPACE::detail::createResultValueType( result, std::move( descriptorSets ) );
  }
```

You can see here that the template type DescriptorSetAllocator must be passed by mutable reference as DescriptorSetAllocator &descriptorSetAllocator.

Now, the various constructors for the various datastructures in the std lib take their allocator argument by _const-ref_: https://en.cppreference.com/w/cpp/container/vector/vector.html

Would it make sense to make all the allocator arguments `const&` instead of `&`? I've taken to writing my functions taking a `const Alloc&` allocator type, and when such a `const Alloc&` allocator type is passed to `allocateDescriptorSets`, the template type is deduced to be `const Alloc`, which makes the return type be `std::vector<vk::DescriptorSet, const Alloc>`, which is not allowed.

If you pass a mutable `Alloc&` type to `allocateDescriptorSets`, an additional `const` would not make the function fail to compile because the type of `DescriptorSetAllocator` would still be `Alloc`. Let me make a table summary...

| input arg from user | function input arg type  | template | deduced type | vector return type
| ------------------------- | ------------------------------ | ------------- | ----------------- | ---------------
| `Alloc&`                   | `Alloc&`                        | `Alloc`      | `Alloc`           | `std::vector<vk::DescriptorSet, Alloc>` (1) ✅
| `const Alloc&`          | `Alloc&`                       | `Alloc`      | `const Alloc`  | `std::vector<vk::DescriptorSet, const Alloc>` (1) ❌
| `Alloc&`                   | `const Alloc&`              | `Alloc`      | `Alloc`           | `std::vector<vk::DescriptorSet, Alloc>` (2) ✅
| `const Alloc&`         | `const Alloc&`              | `Alloc`       | `Alloc`          | `std::vector<vk::DescriptorSet, Alloc>` (2) ✅

So I'm suggesting to take all allocator args by const-ref to avoid the situation in the second row. Alternatively, you could use `std::decay_t<Alloc>` or `std::remove_const_t<Alloc>` to remove the constness when defining the vector type.

asuessenbach:
Resolves #2315.

### DIFF
--- a/tests/DeviceFunctions/DeviceFunctions.cpp
+++ b/tests/DeviceFunctions/DeviceFunctions.cpp
@@ -24,24 +24,32 @@
 // unknow compiler... just ignore the warnings for yourselves ;)
 #endif
 
-
 #ifdef VULKAN_HPP_USE_CXX_MODULE
-#  include <vulkan/vulkan_hpp_macros.hpp> // VULKAN_HPP_DISPATCH_LOADER_DYNAMIC_TYPE
 #  include <cassert>
 #  include <cstdint>
+#  include <vulkan/vulkan_hpp_macros.hpp>  // VULKAN_HPP_DISPATCH_LOADER_DYNAMIC_TYPE
 import vulkan;
 #else
-#  include <vector>
-#  include <cstdint>
-#  include <iostream>
 #  include <algorithm>
 #  include <cassert>
+#  include <cstdint>
+#  include <iostream>
+#  include <vector>
 #  include <vulkan/vulkan.hpp>
 #endif
 
-
 static char const * AppName    = "DeviceFunctions";
 static char const * EngineName = "Vulkan.hpp";
+
+template <class Dispatch = VULKAN_HPP_DEFAULT_DISPATCHER_TYPE, class Alloc = std::allocator<vk::UniqueHandle<vk::CommandBuffer, Dispatch>>>
+
+std::vector<vk::UniqueHandle<vk::CommandBuffer, Dispatch>, Alloc> createCommandBuffers( const vk::Device &                    device,
+                                                                                        const vk::CommandBufferAllocateInfo & allocateInfo,
+                                                                                        const Alloc &                         alloc = Alloc(),
+                                                                                        const Dispatch & d = VULKAN_HPP_DEFAULT_DISPATCHER )
+{
+  return device.allocateCommandBuffersUnique( allocateInfo, alloc, d );
+}
 
 int main( int /*argc*/, char ** /*argv*/ )
 {
@@ -76,11 +84,20 @@ int main( int /*argc*/, char ** /*argv*/ )
     vk::UniqueCommandBuffer                              commandBuffer =
       std::move( device->allocateCommandBuffersUnique( {}, vectorAllocator, VULKAN_HPP_DISPATCH_LOADER_STATIC_TYPE() ).front() );
 
+    std::vector<vk::UniqueCommandBuffer> uniqueCommandBuffers;
+    uniqueCommandBuffers = createCommandBuffers( device.get(), {}, uniqueCommandBuffers.get_allocator(), VULKAN_HPP_DISPATCH_LOADER_STATIC_TYPE() );
+    commandBuffer        = std::move( uniqueCommandBuffers.front() );
+
     commandBuffer->begin( vk::CommandBufferBeginInfo() );
 
     std::vector<vk::UniqueHandle<vk::CommandBuffer, VULKAN_HPP_DISPATCH_LOADER_DYNAMIC_TYPE>>::allocator_type dynamicVectorAllocator;
     vk::UniqueHandle<vk::CommandBuffer, VULKAN_HPP_DISPATCH_LOADER_DYNAMIC_TYPE>                              dynamicCommandBuffer =
       std::move( device->allocateCommandBuffersUnique( {}, dynamicVectorAllocator, VULKAN_HPP_DISPATCH_LOADER_DYNAMIC_TYPE() ).front() );
+
+    std::vector<vk::UniqueHandle<vk::CommandBuffer, VULKAN_HPP_DISPATCH_LOADER_DYNAMIC_TYPE>> dynamicUniqueCommandBuffers;
+    dynamicUniqueCommandBuffers =
+      createCommandBuffers( device.get(), {}, dynamicUniqueCommandBuffers.get_allocator(), VULKAN_HPP_DISPATCH_LOADER_DYNAMIC_TYPE() );
+    dynamicCommandBuffer = std::move( dynamicUniqueCommandBuffers.front() );
 
     vk::Buffer       buffer       = device->createBuffer( {} );
     vk::UniqueBuffer uniqueBuffer = vk::UniqueBuffer( buffer, *device );


rwols:
I checked out this branch and I can confirm that this branch has the desired behavior (would close https://github.com/KhronosGroup/Vulkan-Hpp/issues/2315).

I verified this by changing the test in tests/DeviceFunctions/DeviceFunctions.cpp in the below diff. The function `createCommandBuffers` in this diff does _not_ compile on the main branch while it does compile on the const_ref_allocators branch :+1: perhaps you can add these changes to DeviceFunctions.cpp?

```diff
diff --git a/tests/DeviceFunctions/CMakeLists.txt b/tests/DeviceFunctions/CMakeLists.txt
index 718e7ab..ba6d114 100644
--- a/tests/DeviceFunctions/CMakeLists.txt
+++ b/tests/DeviceFunctions/CMakeLists.txt
@@ -1,4 +1,4 @@
-# Copyright(c) 2018, NVIDIA CORPORATION. All rights reserved.
+# Copyright(c) 2026, NVIDIA CORPORATION. All rights reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
diff --git a/tests/DeviceFunctions/DeviceFunctions.cpp b/tests/DeviceFunctions/DeviceFunctions.cpp
index 1c11e8f..2cefc82 100644
--- a/tests/DeviceFunctions/DeviceFunctions.cpp
+++ b/tests/DeviceFunctions/DeviceFunctions.cpp
@@ -1,4 +1,4 @@
-// Copyright(c) 2018, NVIDIA CORPORATION. All rights reserved.
+// Copyright(c) 2026, NVIDIA CORPORATION. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -41,6 +41,17 @@
 static char const * AppName    = "DeviceFunctions";
 static char const * EngineName = "Vulkan.hpp";
 
+template <
+    class Dispatch = VULKAN_HPP_DEFAULT_DISPATCHER_TYPE,
+    class Alloc = std::allocator<vk::UniqueHandle<vk::CommandBuffer, Dispatch>>>
+std::vector<vk::UniqueHandle<vk::CommandBuffer, Dispatch>, Alloc>
+createCommandBuffers(const vk::Device &device,
+                     const vk::CommandBufferAllocateInfo &allocateInfo,
+                     const Alloc &alloc = Alloc(),
+                     const Dispatch &d = VULKAN_HPP_DEFAULT_DISPATCHER) {
+  return device.allocateCommandBuffersUnique(allocateInfo, alloc, d);
+}
+
 int main( int /*argc*/, char ** /*argv*/ )
 {
   try
@@ -70,15 +81,15 @@ int main( int /*argc*/, char ** /*argv*/ )
 
     uint64_t handle = device->getAccelerationStructureHandleNV<uint8_t>( {}, VULKAN_HPP_DISPATCH_LOADER_DYNAMIC_TYPE() );
 
-    std::vector<vk::UniqueCommandBuffer>::allocator_type vectorAllocator;
-    vk::UniqueCommandBuffer                              commandBuffer =
-      std::move( device->allocateCommandBuffersUnique( {}, vectorAllocator, VULKAN_HPP_DISPATCH_LOADER_STATIC_TYPE() ).front() );
-
+    std::vector<vk::UniqueCommandBuffer> vector;
+    vector = createCommandBuffers(device.get(), {}, vector.get_allocator(), VULKAN_HPP_DISPATCH_LOADER_STATIC_TYPE());
+    vk::UniqueCommandBuffer commandBuffer = std::move(vector.front());
     commandBuffer->begin( vk::CommandBufferBeginInfo() );
 
-    std::vector<vk::UniqueHandle<vk::CommandBuffer, VULKAN_HPP_DISPATCH_LOADER_DYNAMIC_TYPE>>::allocator_type dynamicVectorAllocator;
-    vk::UniqueHandle<vk::CommandBuffer, VULKAN_HPP_DISPATCH_LOADER_DYNAMIC_TYPE>                              dynamicCommandBuffer =
-      std::move( device->allocateCommandBuffersUnique( {}, dynamicVectorAllocator, VULKAN_HPP_DISPATCH_LOADER_DYNAMIC_TYPE() ).front() );
+    std::vector<vk::UniqueHandle<vk::CommandBuffer, VULKAN_HPP_DISPATCH_LOADER_DYNAMIC_TYPE>> dynamicVector;
+    dynamicVector = createCommandBuffers(device.get(), {}, dynamicVector.get_allocator(), VULKAN_HPP_DISPATCH_LOADER_DYNAMIC_TYPE());
+
+    vk::UniqueHandle<vk::CommandBuffer, VULKAN_HPP_DISPATCH_LOADER_DYNAMIC_TYPE> dynamicCommandBuffer = std::move(dynamicVector.front());
 
     vk::Buffer       buffer       = device->createBuffer( {} );
     vk::UniqueBuffer uniqueBuffer = vk::UniqueBuffer( buffer, *device );
```

rwols:
Well, when trying to run this test it segfaults at the vector::front() call because we get an empty vector. I guess this is just a compile-time test?

asuessenbach:
Thanks for providing some test code. Added it to the DeviceFunctions test.

And yes, it's a pure compile test, not meant for actually running it.

PR #2396, opened 2025-12-15
Fix VULKAN_HPP_NO_SMART_HANDLE compilation

YaaZ:
Forward declaration of `AllocationCallbacks` is guarded by `VULKAN_HPP_NO_SMART_HANDLE`, while being used uncoditionally in `resultCheck`. This breaks compilation when `VULKAN_HPP_NO_SMART_HANDLE` is defined.

Note that it's not enough to add `VULKAN_HPP_NO_SMART_HANDLE` to a sample to reproduce this, because the PCH used does not define this. You need to either disable the PCH, or compile PCH with that definition.

### DIFF
--- a/snippets/HppTemplate.hpp
+++ b/snippets/HppTemplate.hpp
@@ -28,8 +28,9 @@ namespace VULKAN_HPP_NAMESPACE
   }
   ${Exchange}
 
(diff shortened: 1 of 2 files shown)


asuessenbach:
Thanks for spotting this issue and providing a fix for it!

PR #2409, opened 2025-12-23
Include <cstdlib> on macos for getenv

agamache:
Building vulkan.cppm on macOS fails after <cstdlib> was removed from vulkan.hpp in #2344 since the macOS version calls getenv to read the DYLD_LIBRARY_PATH environment variable.

### DIFF
--- a/snippets/DynamicLoader.hpp
+++ b/snippets/DynamicLoader.hpp
@@ -30,7 +30,7 @@
         }
 		// modern versions of macOS don't search /usr/local/lib automatically contrary to what man dlopen says
 		// Vulkan SDK uses this as the system-wide installation location, so we're going to fallback to this if all else fails
-		if ( !m_library && ( getenv("DYLD_FALLBACK_LIBRARY_PATH") == NULL ) )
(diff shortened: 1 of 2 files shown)


M2-TE:
Thanks for catching that! Could you adjust `snippets/includes.hpp` instead of the generated file and then run the generator?

Also, could you test whether substituting `getenv` with `std::getenv` allows it to compile without including the additional header? The `import std` is a bit inconsistent about what it exports and e.g. `std::exit` is exported, whereas `exit` is not.

dsvensson:
@M2-TE Yep, changing to std::getenv also works.

asuessenbach:
@dsvensson That is, this PR can be closed without merging?

M2-TE:
The issue is still valid, though the solution should be different than what is presented. In short:
* instead of including `<cstdlib>`, `getenv` should be prefixed by `std::`
* generator code should be changed, not just `vulkan.hpp`

I could create a separate PR to do this, unless @agamache wants to implement the changes here

agamache:
I have updated this pr to use std::getenv in snippets/DynamicLoader.hpp and reran the generator

PR #2422, opened 2026-01-07
Introduce flag VULKAN_HPP_NO_FORMAT to prevent formatting of generated files.

asuessenbach:
This flag is used with the CI workflows to reduce time spent there.


asuessenbach:
I'll keep this PR for a while, but it doesn't substantially improve CI performance. Maybe a minute less per workflow.
The major part of the time is spent with building the samples and tests.

M2-TE:
Was thinking about this again and it might be bad for parsing errors in CI when they inevitably occur.
Unformatted headers will display different line/character values of code for those errors.

asuessenbach:
Great, I was just looking for a reason to cancel this PR!
After all, it's not as needed as expected, after all the CI work of the past weeks.

PR #2427, opened 2026-01-08
CI: Testing C++Modules on MacOS

### LINKED ISSUE #2420: Automated testing of C++Modules on MacOS
MacOS is the last OS missing for full coverage of modules (excluding gcc on linux).
Is there a bleeding edge distribution of Clang 20 or higher over there that we could use? Same with CMake?

There's [Clang 21](https://formulae.brew.sh/formula/llvm) and [CMake 4.2.1](https://formulae.brew.sh/formula/cmake) in homebrew, so perhaps it can be done through that. I'm wholly unfamiliar with MacOS myself though, so I would love if someone with more experience could chime in.

M2-TE:
Will resolve #2420.
Runner is based on the new `MacOS 26` image and uses `Clang 21` with `CMake 4.2.1`.
It **only** runs module tests, which I've denoted in the runner name. Same for the MSVC module runner.

Note that the tests will fail until #2409 is merged. I do at the very least _want_ to see them fail here at least once; just to know that they do their job.

### DIFF
--- a/snippets/DynamicLoader.hpp
+++ b/snippets/DynamicLoader.hpp
@@ -30,7 +30,7 @@
         }
 		// modern versions of macOS don't search /usr/local/lib automatically contrary to what man dlopen says
 		// Vulkan SDK uses this as the system-wide installation location, so we're going to fallback to this if all else fails
-		if ( !m_library && ( getenv("DYLD_FALLBACK_LIBRARY_PATH") == NULL ) )
+		if ( !m_library && ( std::getenv("DYLD_FALLBACK_LIBRARY_PATH") == NULL ) )
 		{
 		  m_library = dlopen( "/usr/local/lib/libvulkan.dylib", RTLD_NOW | RTLD_LOCAL );
 		}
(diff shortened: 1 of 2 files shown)


M2-TE:
Yup, looks good!
```
In file included from /Users/runner/work/Vulkan-Hpp/Vulkan-Hpp/vulkan/vulkan_video.cppm:22:
In file included from /Users/runner/work/Vulkan-Hpp/Vulkan-Hpp/vulkan/vulkan_video.hpp:17:
/Users/runner/work/Vulkan-Hpp/Vulkan-Hpp/vulkan/vulkan.hpp:20969:32: error: missing '#include <_stdlib.h>'; 'getenv' must be declared before it is used
 20969 |           if ( !m_library && ( getenv( "DYLD_FALLBACK_LIBRARY_PATH" ) == NULL ) )
       |                                ^
/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include/_stdlib.h:167:18: note: declaration here is not visible
  167 | char *_LIBC_CSTR        getenv(const char *);
      |                         ^
2 warnings and 1 error generated.
```

PR #2428, opened 2026-01-08
Run the FunctionCalls test unconditionally.

asuessenbach:


M2-TE:
You can remove the CTest guard altogether, it's not necessary anymore. The test section in CMake should look like this:

```cmake
# can only run these tests with VULKAN_HPP_DISPATCH_LOADER_DYNAMIC disabled
if( NOT VULKAN_HPP_DISPATCH_LOADER_DYNAMIC )
	vulkan_hpp__setup_test( NAME DeviceFunctions )
	vulkan_hpp__setup_test( NAME DispatchLoaderStatic )
endif()
# can only run these tests with VULKAN_HPP_DISPATCH_LOADER_DYNAMIC enabled
if( VULKAN_HPP_DISPATCH_LOADER_DYNAMIC )
	vulkan_hpp__setup_test( NAME DispatchLoaderDynamic )
	vulkan_hpp__setup_test( NAME DispatchLoaderDynamicSharedLibraryClient )
	vulkan_hpp__setup_test( NAME NoExceptions )
  	vulkan_hpp__setup_test( NAME FunctionCalls )
endif()
```

Which is assuming `FunctionCalls` still uses the dynamic dispatcher in its test.

PR #2430, opened 2026-01-08
CI: Merge all CI workflows into single workflow file

### LINKED ISSUE #2421: Expand CI for MacOS
We currently test with Clang on `MacOS 14` /`15`. Apparently, a GitHub hosted [`MacOS 26`](https://github.com/actions/runner-images/blob/main/images/macos/macos-26-arm64-Readme.md)runner is now in public beta, which we could consider using. It has `gcc 15`, `Clang 21` and `CMake 4.2.1,` which is great for module testing in particular (finally some GCC module coverage). Relevant for #2420.

There are two considerations:
1. Some of the MacOS images run on ARM processors. Might be worth throwing an ARM runner into the mix, no matter the version.
2. The `MacOS 26` runner is both ARM and also in public _beta_, though I'm not sure what the latter part entails exactly (software version changes in the future?). Having a runner with more up-to-date tools would be great either way.

M2-TE:
Current workflow setup is spread across too many files and makes changes to e.g. flags rather cumbersome. They are now all merged into one, providing a good overview over all the OSs, compilers, etc that are being used. Based on [VulkanMemoryAllocator-Hpp](https://github.com/YaaZ/VulkanMemoryAllocator-Hpp) workflow by @YaaZ. Also resolves #2421.

There are a few small changes that will make managing these workflows a bit easier:
* adding/removing/changing workflows should be much easier now
* `cxx_max` allows to easily lower or raise the maximum C++ version for either older or newer compilers respectively.
* `gen` can be specified to be a certain generator. This merges the previously fractured MSBuild/Ninja workflows.
* separated setup tasks for each OS, particularly for modules that require a bit of tinkering for each

Changes that I would like to ask for feedback on:
* ~Headers are generated with `cxx_max` instead of always with C++20~ Generators have C++20 hardcoded in CMake
* Samples are only compiled once with `cxx_max`, similar to #2424
* Header gen and sample compilation merged into single step for better file readability

So let me know @asuessenbach and @sharadhr if this is something you like or not.

...ignore the commit names, I was going insane


.github/workflows/build.yml
```diff
@@ -0,0 +1,222 @@
...
+
+          { os: windows-2022,                  arch: x86, gen: Visual Studio 17 2022, cxx_max: 23 },
+          { os: windows-2022,                  arch: x64, gen: Visual Studio 17 2022, cxx_max: 23 },
+          { os: windows-2022, cxx: c++,        arch: x64, gen: Visual Studio 17 2022, cxx_max: 23 },
+          { os: windows-2022, cxx: clang-cl,   arch: x64, gen: Ninja,                 cxx_max: 23 },
+
+          { os: windows-2025,                  arch: x86, gen: Visual Studio 17 2022, cxx_max: 23 },
+          { os: windows-2025,                  arch: x64, gen: Visual Studio 17 2022, cxx_max: 23 },
+          { os: windows-2025, cxx: c++,        arch: x64, gen: Visual Studio 17 2022, cxx_max: 23 },
+          { os: windows-2025, cxx: clang-cl,   arch: x64, gen: Ninja,                 cxx_max: 23 },
```
sharadhr: The Windows invocations will need to be tweaked a little.

The independent variables for Windows:

1. Compilers:
   1. MSVC `cl.exe`; 
   2. `clang-cl.exe`. This is equivalent to `clang++.exe --driver-mode=cl`; accepts MSVC-style arguments; 
   3. `clang++.exe`; accepts GNU-style arguments, links against MSVC STL
2. Generators: 
   1. `Visual Studio 17 2022`;
   2. `Visual Studio 18 2026`;
   3. `Ninja`
3. Windows version

CMake only supports `cl.exe` + `Ninja` out-of-the-box for modules; it can be convinced to use `clang++.exe` with [this patch](https://gitlab.kitware.com/cmake/cmake/-/issues/26761#note_1734813); note my comment below to suppress some spurious warnings.

.github/workflows/build.yml
```diff
@@ -0,0 +1,222 @@
...
+          -D VULKAN_HPP_RUN_GENERATOR=OFF                 \
+          -D VULKAN_HPP_SAMPLES_BUILD=ON                  \
+          -D VULKAN_HPP_TESTS_BUILD=OFF                   \
+          -D VULKAN_HPP_DISPATCH_LOADER_DYNAMIC=ON        \
+          -D VULKAN_HPP_BUILD_WITH_LOCAL_VULKAN_HPP=ON    \
+          -D VULKAN_HPP_PRECOMPILE=OFF                    \
+          -D CMAKE_C_COMPILER=${{matrix.env.cxx}}         \
+          ${{matrix.env.cxx && format('-D CMAKE_CXX_COMPILER={0}', matrix.env.cxx)}} \
+          -D CMAKE_CXX_STANDARD=${{matrix.env.cxx_max}}   \
+          -D CMAKE_BUILD_TYPE=Debug
```
sharadhr: These CMake variables are repeated throughout the script; is there a way we can generate, compile, and test in one go? Perhaps by moving from Ninja to [Ninja Multi-Config](https://cmake.org/cmake/help/latest/generator/Ninja%20Multi-Config.html)?

.github/workflows/build.yml
M2-TE: Not sure what exactly you want to see tweaked here;

* A missing `cxx` uses the default compiler (`cl.exe`) and `clang-cl` is explicitly used (don't think we need to use clang++ in addition to it, right?). `c++` uses gcc from what I gather. Should I explicitly use `cxx: cl` for the first?
* As for generators, I chose `Visual Studio 17 2022` (MSBuild), as that is what's installed on the runner image (apart from `Ninja`).
* Windows versions are just 2022 and 2025, same as before

The runners for module tests are separate, a little below these. On Windows, it uses `cl` + `Ninja` as you already mentioned.

.github/workflows/build.yml
M2-TE: We could probably shorten all of these CMake args a lot, yea. I wasn't sure where to draw the line between "concise" and "readable", since I didn't want to end up writing numpy style python code that summons demons when you aren't looking..

Using Multi-Config might be something worth trying out! Generate+Samples can be combined quite easily already. Tests should likely be separate as those clock through the different C++ versions.

And, I completely forgot, with my changes so far, we don't need to install the `glm`, `glfw` and `glsl` libs anymore (they are only needed for samples now). I'll remove those!

.github/workflows/build.yml
sharadhr: Hmm, that makes sense. If the compiler is not explicitly specified then it defaults to `cl.exe`, yes. 

> don't think we need to use clang++ in addition to it, right?

If we want to test modules then possibly, but it's not critical.

> c++ uses gcc from what I gather

Is this MinGW GCC, and if so, is it installed on the runner? I'm not sure it'll work with the VS generator; best to check. Also, do we need GCC here if we already have it on Linux?

M2-TE: I already ran these tests on my fork and they all pass, so it does seem to work at least!

> If we want to test modules then possibly, but it's not critical.

Fair point. I was happy with having modules on every major OS for now. MacOS 26 has g++-15, so we could even have module coverage for all major compilers. I would prioritize that next.

As for the MinGW GCC, I'm really not sure. Is it a similar instance as with Apple-Clang, where it's slightly different than linux Clang? If not, we can probably leave it out. I was just interested to see if it would compile.

sharadhr: > Is it a similar instance as with Apple-Clang

It's GCC proper, and the standard library is provided by MinGW. Naturally yes, there are slight differences (e.g. `syscall.h` is naturally not available), but by and large GCC is used much more often on Linux, and I would prioritise that over GCC on any other platform.

.github/workflows/build.yml
M2-TE: Looks pretty concise now, don't think we even need any multi-config things.

.github/workflows/build.yml
M2-TE: agreed, I've disabled the windows+gcc runner for now!

.github/workflows/build.yml
```diff
@@ -0,0 +1,148 @@
...
+
+          { os: windows-2022, cxx: cl,         c: cl,       arch: x86, gen: Visual Studio 17 2022, cxx_max: 23 },
+          { os: windows-2022, cxx: cl,         c: cl,       arch: x64, gen: Visual Studio 17 2022, cxx_max: 23 },
+          { os: windows-2022, cxx: clang-cl,   c: clang-cl, arch: x64, gen: Ninja,                 cxx_max: 23 },
+          # { os: windows-2022, cxx: c++,        c: cc,       arch: x64, gen: Visual Studio 17 2022, cxx_max: 23 },
+
+          { os: windows-2025, cxx: cl,         c: cl,       arch: x86, gen: Visual Studio 17 2022, cxx_max: 23 },
+          { os: windows-2025, cxx: cl,         c: cl,       arch: x64, gen: Visual Studio 17 2022, cxx_max: 23 },
+          { os: windows-2025, cxx: clang-cl,   c: clang-cl, arch: x64, gen: Ninja,                 cxx_max: 23 },
+          # { os: windows-2025, cxx: c++,        c: cc,       arch: x64, gen: Visual Studio 17 2022, cxx_max: 23 },
```
asuessenbach, line 43: What's this out-commented line for?

CMakeLists.txt
```diff
@@ -565,6 +565,7 @@ function( vulkan_hpp__setup_test )
 					-D VULKAN_HPP_TYPESAFE_CONVERSION=${VULKAN_HPP_TYPESAFE_CONVERSION}
 					-D VULKAN_HPP_USE_REFLECT=${VULKAN_HPP_USE_REFLECT}
 					# forward compiler setup
+					-D CMAKE_EXE_LINKER_FLAGS=${CMAKE_EXE_LINKER_FLAGS}
```
asuessenbach, line 568: What is this `CMAKE_EXE_LINKER_FLAGS` for?

Vulkan-Headers
asuessenbach, line 1: Is this needed for this PR?

asuessenbach (commented):
This is a great PR, I really like it!

Just some minor questions.

.github/workflows/build.yml
M2-TE: That tests gcc on windows (MinGW). It would be a sort of duplicate testing of gcc, so I took it out of the matrix temporarily. Still left it in as a comment in case it is ever desired to test that specific configuration.

CMakeLists.txt
M2-TE: Some compiler configurations need special linker flags to get modules to work.
That line specifically simply forwards the linker flags to the otherwise isolated ctest (I tried to be very careful about what exactly to forward to it).

Vulkan-Headers
M2-TE: Nope, that snuck in due to a mistake on my end. I do wonder how though, since I specifically gitignored it locally..

M2-TE:
Are there any changes you would like to see before merging this @asuessenbach?
I'm working on fixing issues regarding `VULKAN_HPP_PRECOMPILE` as a separate PR and would like test it using this runner matrix

asuessenbach:
Great work, @M2-TE, thanks a lot!

PR #2431, opened 2026-01-10
CI: Added workflow to build against latest spec from Vulkan-Docs

### LINKED ISSUE #1761: Add CI stage to build against current Vulkan-Docs main branch
Per #1760, could you add a CI stage that pulls the current Vulkan-Docs main branch, generates whatever artifacts you need from that, and builds / compiles Vulkan-Hpp based on that? Basically the counterpart of the vulkan hpp-generate / hpp-compile stages. That would hopefully catch new Vulkan-Hpp issues before they affect all new runs of the CI stages in the vulkan repos.

M2-TE:
This will resolve #1761 and performs the following:
1. Pulls and runs on khronos docker `sha256:f1ca671f3bdb10ad49e238b9bf28853088a21af49504498fc9084c9b4fea4762`
2. Pulls latest Vulkan-Docs and generate headers via `make -C Vulkan-Docs/xml clean install codec_headers`
3. Generates Vulkan-Hpp headers from latest Vulkan-Docs xml registry
4. Builds samples/tests with latest Vulkan-Docs generated headers in C++23 using the default compiler (gcc 14.2.0)

Please have a look @oddhack if this is sufficient; in particular the method of generating the header artefacts in step 2.
I could also update the CI steps over at `Vulkan-Docs` once everyone is happy with this runner, so that they both run identical setups (at the very least, it should also use the docker image over there as tracked in KhronosGroup/Vulkan-Docs#2654).

### DIFF
--- a/.github/workflows/build-latest-spec.yml
+++ b/.github/workflows/build-latest-spec.yml
@@ -1,0 +1,46 @@
+name: CI Build with latest Vulkan-Docs
+
+on:
+  workflow_dispatch:
+  pull_request:
+    branches: [ main ]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    container: khronosgroup/docker-images@sha256:f1ca671f3bdb10ad49e238b9bf28853088a21af49504498fc9084c9b4fea4762
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          submodules: recursive
+
+      - name: Generate Vulkan Headers
+        run: |
+          # Generate the headers
+          git clone https://github.com/KhronosGroup/Vulkan-Docs.git --depth 1
+          make -C Vulkan-Docs/xml clean install codec_headers
+
+          # Copy headers and xml registry to Vulkan-Headers (makes it easier for Vulkan-Hpp)
+          rm -rf Vulkan-Headers/include/ Vulkan-Headers/registry
+          cp -r Vulkan-Docs/gen/include/ Vulkan-Headers/include/
+          cp -r Vulkan-Docs/xml Vulkan-Headers/registry/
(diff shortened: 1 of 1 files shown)


oddhack:
It seems fine to me. I take it the 'ctest' bit at the end is what actually runs compiler tests against the generated header?

I'm good with updating Vulkan-Docs to use the container as well. Not sure about updating the hpp-generate / hpp-compile rules there to match here - we could try and see if it makes a big performance difference. If I looked at the right page, the action in your personal repo this PR is from added something like 10 minutes on the runner to CI.

M2-TE:
> I take it the 'ctest' bit at the end is what actually runs compiler tests against the generated header?

Yup, those are the unit tests as controlled by `VULKAN_HPP_TESTS_BUILD` and `VULKAN_HPP_TESTS_CTEST`.

> something like 10 minutes on the runner to CI

Most of that time is spent on the samples (`VULKAN_HPP_SAMPLES_BUILD`). I've included them for completeness, but they can be left out; unit tests are likely more than sufficient for this. Those take like 2-3 minutes at worst. Even the tests probably aren't needed over at Vulkan-Docs.

> Not sure about updating the hpp-generate / hpp-compile rules there to match here

Potentially no need to update much (if anything), the big CMake arg sausage you see here can easily be split up if needed (and parts of it removed). I just put it all into the same step for brevity.

oddhack:
BTW, I think you had commented somewhere about the 'No matching API versions' warning from the codec header generation, and that will go away after the next public spec update - fixed internally in vulkan MR 7972.

asuessenbach:
Thanks a lot @M2-TE, for this great addition. Should make issues with new spec features less likely.

PR #2439, opened 2026-01-14
Samples: Reintroduced support for VULKAN_HPP_PRECOMPILE

M2-TE:
Using `VULKAN_HPP_PRECOMPILE` did not work before, as the PCH and samples sometimes had conflicting `VULKAN_HPP_DISPATCH_LOADER_DYNAMIC` values. Changes made to solve this:
1. Samples simply use include directories instead of linking to `Vulkan::Hpp`. It would otherwise need some kind of "completely clean" Vulkan target with no compile definitions.
2. Added new `utils` library variants `utils_static` and `utils_dynamic` for the respective samples.
3. Extended the PCH of samples with more headers. Reused by all variants (static, dynamic, RAII).
4. Enabled `VULKAN_HPP_PRECOMPILE` on CI

These changes quite noticably reduced the compile times of the sample build stage.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -402,8 +402,6 @@ function( vulkan_hpp__setup_library )
 		vulkan_hpp__setup_platform( NAME ${TARGET_NAME} )
 		vulkan_hpp__setup_warning_level( NAME ${TARGET_NAME} )
 		set_target_properties( ${TARGET_NAME} PROPERTIES CXX_STANDARD_REQUIRED ON )
-		vulkan_hpp__setup_platform( NAME ${TARGET_NAME} )
-		vulkan_hpp__setup_warning_level( NAME ${TARGET_NAME} )
 		if( VULKAN_HPP_BUILD_WITH_LOCAL_VULKAN_HPP )
 			# Vulkan C headers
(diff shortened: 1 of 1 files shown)


sharadhr:
To further work on this, my goal after #2358 is to build the samples with the Vulkan module. It would be a further stress/compile test to see how robust the module is.

CMakeLists.txt
```diff
@@ -401,8 +401,18 @@ function( vulkan_hpp__setup_library )
 		endif()
 		vulkan_hpp__setup_platform( NAME ${TARGET_NAME} )
 		vulkan_hpp__setup_warning_level( NAME ${TARGET_NAME} )
-		target_link_libraries( ${TARGET_NAME} PUBLIC Vulkan::Hpp )
 		set_target_properties( ${TARGET_NAME} PROPERTIES CXX_STANDARD_REQUIRED ON )
+		vulkan_hpp__setup_platform( NAME ${TARGET_NAME} )
+		vulkan_hpp__setup_warning_level( NAME ${TARGET_NAME} )
```
asuessenbach: `setup_platform` and `setup_warning_level` again?

M2-TE: Those were indeed spurious, good catch!

asuessenbach:

PR #2441, opened 2026-01-16
CI: Separate header gen and sample stages

M2-TE:
I had previously merged them to be concise, but separating these has two distinct time advantages:
1. We can skip running the generator in Debug mode, which is suprisingly slow to execute
2. Module runners can skip the sample stage altogether, shaving at least 10 minutes off their total runtime

This pretty much covers all the ideas I had for reducing these runner times and will (hopefully) be the last PR about them for a while..

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -120,7 +120,7 @@ jobs:
             -D VULKAN_HPP_RUN_GENERATOR=OFF                                                                      \
             -D VULKAN_HPP_SAMPLES_BUILD=ON                                                                       \
             -D VULKAN_HPP_TESTS_BUILD=OFF                                                                        \
-            -D VULKAN_HPP_TESTS_CTEST=ON                                                                         \
+            -D VULKAN_HPP_TESTS_CTEST=OFF                                                                        \
             -D VULKAN_HPP_DISPATCH_LOADER_DYNAMIC=ON                                                             \
             -D VULKAN_HPP_BUILD_WITH_LOCAL_VULKAN_HPP=ON                                                         \
             -D VULKAN_HPP_PRECOMPILE=ON                                                                          \


.github/workflows/build.yml
```diff
@@ -90,24 +90,40 @@ jobs:
...
-            -D VULKAN_HPP_SAMPLES_BUILD=ON                \
-            -D VULKAN_HPP_TESTS_BUILD=OFF                 \
-            -D VULKAN_HPP_TESTS_CTEST=ON                  \
-            -D VULKAN_HPP_DISPATCH_LOADER_DYNAMIC=ON      \
-            -D VULKAN_HPP_BUILD_WITH_LOCAL_VULKAN_HPP=ON  \
-            -D VULKAN_HPP_PRECOMPILE=OFF                  \
-            -D CMAKE_CXX_STANDARD=${{matrix.env.cxx_max}} \
+            -D VULKAN_HPP_GENERATOR_BUILD=ON                                                                     \
+            -D VULKAN_HPP_RUN_GENERATOR=$RUN_GENERATOR                                                           \
+            -D VULKAN_HPP_DISPATCH_LOADER_DYNAMIC=ON                                                             \
```
asuessenbach, line 104: When you just build the generator and generate the headers, `VULKAN_HPP_DISPATCH_LOADER_DYNAMIC` should not be needed.

.github/workflows/build.yml
```diff
@@ -90,24 +90,40 @@ jobs:
...
+          cmake -B build -G '${{matrix.env.gen}}' --fresh                                                        \
+            ${{matrix.env.c && format('-D CMAKE_C_COMPILER={0}', matrix.env.c)}}                                 \
+            ${{matrix.env.cxx && format('-D CMAKE_CXX_COMPILER={0}', matrix.env.cxx)}}                           \
+            ${{matrix.env.cxx_flags && format('-D CMAKE_CXX_FLAGS''={0}''', matrix.env.cxx_flags)}}              \
+            ${{matrix.env.linker_flags && format('-D CMAKE_EXE_LINKER_FLAGS=''{0}''', matrix.env.linker_flags)}} \
+            -D VULKAN_HPP_GENERATOR_BUILD=OFF                                                                    \
+            -D VULKAN_HPP_RUN_GENERATOR=OFF                                                                      \
+            -D VULKAN_HPP_SAMPLES_BUILD=ON                                                                       \
+            -D VULKAN_HPP_TESTS_BUILD=OFF                                                                        \
+            -D VULKAN_HPP_TESTS_CTEST=ON                                                                         \
```
asuessenbach: Why is `VULKAN_HPP_TESTS_CTEST` on here?

.github/workflows/build.yml
M2-TE: It may not be needed, but the `Vulkan::Hpp` that will always be created uses `find_package(Vulkan)` if we do not enable the dynamic dispatcher. This suppresses warnings (and potential issues) when it cannot be found (CI), so I left it there.

.github/workflows/build.yml
M2-TE: It has no effect, but I can explicitly disable it for the sake of consistency

PR #2443, opened 2026-01-17
Use the ABI-breaking style for the C++ module

### LINKED ISSUE #2366: Find a better way to hardcode `using` statements in module interface file
We had an unfortunate bug slip through that was only resolved with #2362. I have two suggestions:

1. Push all the required symbols into some container in the generator, and then expand them inline. This still requires hard-coding of the symbols themselves, which might still lead to typos and failed compilation.
2. Find some language-level commonality in the hardcoded symbols (for example, `operator<=` etc) during generation so that they can be pushed into the container mentioned in (1). This might need some analysis that I don't yet have the knowledge for. 

Of course, this would have been caught with module testing, but I feel those are orthogonal issues. We need testing anyway, and the generator can also be made more robust.

### LINKED ISSUE #2379: Consider the Clang 'ABI breaking' style for the Vulkan-Hpp module
I've been reading the Clang documentation for migrating header libraries to modules, and I took note of the so-called ['ABI-breaking' style](https://clang.llvm.org/docs/StandardCPlusPlusModules.html#abi-breaking-style). 

We already use a similar 'is in module' macro (`VULKAN_HPP_CXX_MODULE`) to enable `import std` instead of standard library headers. I would like to extend that to have `#define EXPORT export` (naturally guarded behind `#if defined(...)` etc) as detailed in the above documentation. I then hope to simplify the module interface file, as well as completely simplify the generation of the module interface file itself, which currently takes quite a different path to the header generation. This will also probably remove spurious `using` statements for `operator<=>` etc.

Naturally this means the generation of the module interface file, as well as its actual contents would change by a lot, and hence I would like some input on this. I don't expect the consumer-facing experience to change.

### LINKED ISSUE #2447: Move fwd declarations for dynamic dispatch out of the macro header and into vulkan.hpp
The dynamic dispatcher (class + storage) is currently being forward-declared/extern in the macro header:
https://github.com/KhronosGroup/Vulkan-Hpp/blob/55b25f400bcf1eee98f7824e6af104d4441d5943/vulkan/vulkan_hpp_macros.hpp#L271-L283

But it would make much more sense to place this into `vulkan.hpp`, seeing how it is not a macro.

### LINKED ISSUE #2448: Automatically create storage for dynamic dispatcher when using modules
Since the vulkan module has its own TU, it would make sense to automate the storage of the dynamic dispatcher.

sharadhr:
Resolves #2366 and resolves #2379. 

Also resolves #2447 and resolves #2448.

Starts with an initial experiment with `vulkan:video`, which seems to work pretty well with the `MODULE_Video` test. 

The only issue is the duplicated `#include`s in both the `.hpp` and the `.module`, but I would rather duplicated headers than all the issues with the `using` statements. The module interface file is also so much shorter now.

### DIFF
--- a/vulkan/vulkan.hpp
+++ b/vulkan/vulkan.hpp
@@ -37,7 +37,7 @@
 #  endif
 #endif
 
-VULKAN_HPP_STATIC_ASSERT( VK_HEADER_VERSION == 342, "Wrong VK_HEADER_VERSION!" );
+VULKAN_HPP_STATIC_ASSERT( VK_HEADER_VERSION == 343, "Wrong VK_HEADER_VERSION!" );
 
 VULKAN_HPP_EXPORT namespace VULKAN_HPP_NAMESPACE
 {
@@ -50,8 +50,6 @@ VULKAN_HPP_EXPORT namespace VULKAN_HPP_NAMESPACE
 #endif
   }  // namespace detail
 }  // namespace VULKAN_HPP_NAMESPACE
-
-VULKAN_HPP_STATIC_ASSERT( VK_HEADER_VERSION == 343, "Wrong VK_HEADER_VERSION!" );
 
 // <tuple> includes <sys/sysmacros.h> through some other header
 // this results in major(x) being resolved to gnu_dev_major(x)
@@ -7753,24 +7751,25 @@ VULKAN_HPP_EXPORT namespace VULKAN_HPP_NAMESPACE
         return ::vkCmdSetComputeOccupancyPriorityNV( commandBuffer, pParameters );
       }
 
-#  if defined( VK_USE_PLATFORM_UBM_SEC )
+#    if defined( VK_USE_PLATFORM_UBM_SEC )
       //=== VK_SEC_ubm_surface ===
 
-      VkResult vkCreateUbmSurfaceSEC( VkInstance                        instance,
-                                      const VkUbmSurfaceCreateInfoSEC * pCreateInfo,
-                                      const VkAllocationCallbacks *     pAllocator,
-                                      VkSurfaceKHR *                    pSurface ) const VULKAN_HPP_NOEXCEPT
+      VULKAN_HPP_INLINE VkResult vkCreateUbmSurfaceSEC( VkInstance                        instance,
+                                                        const VkUbmSurfaceCreateInfoSEC * pCreateInfo,
+                                                        const VkAllocationCallbacks *     pAllocator,
+                                                        VkSurfaceKHR *                    pSurface ) const VULKAN_HPP_NOEXCEPT
       {
         return ::vkCreateUbmSurfaceSEC( instance, pCreateInfo, pAllocator, pSurface );
       }
 
-      VkBool32 vkGetPhysicalDeviceUbmPresentationSupportSEC( VkPhysicalDevice    physicalDevice,
-                                                             uint32_t            queueFamilyIndex,
-                                                             struct ubm_device * ubm_device ) const VULKAN_HPP_NOEXCEPT
+      VULKAN_HPP_INLINE VkBool32 vkGetPhysicalDeviceUbmPresentationSupportSEC( VkPhysicalDevice    physicalDevice,
+                                                                               uint32_t            queueFamilyIndex,
+                                                                               struct ubm_device * ubm_device ) const VULKAN_HPP_NOEXCEPT
       {
         return ::vkGetPhysicalDeviceUbmPresentationSupportSEC( physicalDevice, queueFamilyIndex, ubm_device );
       }
-#  endif /*VK_USE_PLATFORM_UBM_SEC*/
+#    endif /*VK_USE_PLATFORM_UBM_SEC*/
+#  endif   /*VK_ONLY_EXPORTED_PROTOTYPES*/
     };
 
     inline DispatchLoaderStatic & getDispatchLoaderStatic()
@@ -10833,6 +10832,7 @@ VULKAN_HPP_EXPORT namespace VULKAN_HPP_NAMESPACE
   VULKAN_HPP_CONSTEXPR_INLINE auto SECUbmSurfaceSpecVersion   = VK_SEC_UBM_SURFACE_SPEC_VERSION;
   VULKAN_HPP_CONSTEXPR_INLINE auto SECUbmSurfaceExtensionName = VK_SEC_UBM_SURFACE_EXTENSION_NAME;
 #endif /*VK_USE_PLATFORM_UBM_SEC*/
+
 }  // namespace VULKAN_HPP_NAMESPACE
 
 // clang-format off
--- a/vulkan/vulkan_funcs.hpp
+++ b/vulkan/vulkan_funcs.hpp
@@ -31853,10 +31853,8 @@ VULKAN_HPP_EXPORT namespace VULKAN_HPP_NAMESPACE
 
   // wrapper function for command vkCreateUbmSurfaceSEC, see https://registry.khronos.org/vulkan/specs/latest/man/html/vkCreateUbmSurfaceSEC.html
   template <typename Dispatch, typename std::enable_if<IS_DISPATCHED( vkCreateUbmSurfaceSEC ), bool>::type>
-  VULKAN_HPP_NODISCARD VULKAN_HPP_INLINE Result Instance::createUbmSurfaceSEC( const UbmSurfaceCreateInfoSEC * pCreateInfo,
-                                                                               const AllocationCallbacks *     pAllocator,
-                                                                               SurfaceKHR *                    pSurface,
-                                                                               Dispatch const &                d ) const VULKAN_HPP_NOEXCEPT
+  VULKAN_HPP_NODISCARD VULKAN_HPP_INLINE Result Instance::createUbmSurfaceSEC(
+    const UbmSurfaceCreateInfoSEC * pCreateInfo, const AllocationCallbacks * pAllocator, SurfaceKHR * pSurface, Dispatch const & d ) const VULKAN_HPP_NOEXCEPT
   {
     VULKAN_HPP_ASSERT( d.getVkHeaderVersion() == VK_HEADER_VERSION );
     return static_cast<Result>( d.vkCreateUbmSurfaceSEC( static_cast<VkInstance>( m_instance ),
@@ -31868,8 +31866,8 @@ VULKAN_HPP_EXPORT namespace VULKAN_HPP_NAMESPACE
 #  ifndef VULKAN_HPP_DISABLE_ENHANCED_MODE
   // wrapper function for command vkCreateUbmSurfaceSEC, see https://registry.khronos.org/vulkan/specs/latest/man/html/vkCreateUbmSurfaceSEC.html
   template <typename Dispatch, typename std::enable_if<IS_DISPATCHED( vkCreateUbmSurfaceSEC ), bool>::type>
-  VULKAN_HPP_NODISCARD VULKAN_HPP_INLINE typename ResultValueType<SurfaceKHR>::type
-    Instance::createUbmSurfaceSEC( const UbmSurfaceCreateInfoSEC & createInfo, Optional<const AllocationCallbacks> allocator, Dispatch const & d ) const
+  VULKAN_HPP_NODISCARD VULKAN_HPP_INLINE typename ResultValueType<SurfaceKHR>::type Instance::createUbmSurfaceSEC(
+    const UbmSurfaceCreateInfoSEC & createInfo, Optional<const AllocationCallbacks> allocator, Dispatch const & d ) const
   {
     VULKAN_HPP_ASSERT( d.getVkHeaderVersion() == VK_HEADER_VERSION );
 #    if ( VULKAN_HPP_DISPATCH_LOADER_DYNAMIC == 1 )
@@ -31889,8 +31887,8 @@ VULKAN_HPP_EXPORT namespace VULKAN_HPP_NAMESPACE
 #    ifndef VULKAN_HPP_NO_SMART_HANDLE
   // wrapper function for command vkCreateUbmSurfaceSEC, see https://registry.khronos.org/vulkan/specs/latest/man/html/vkCreateUbmSurfaceSEC.html
   template <typename Dispatch, typename std::enable_if<IS_DISPATCHED( vkCreateUbmSurfaceSEC ), bool>::type>
-  VULKAN_HPP_NODISCARD VULKAN_HPP_INLINE typename ResultValueType<UniqueHandle<SurfaceKHR, Dispatch>>::type
-    Instance::createUbmSurfaceSECUnique( const UbmSurfaceCreateInfoSEC & createInfo, Optional<const AllocationCallbacks> allocator, Dispatch const & d ) const
+  VULKAN_HPP_NODISCARD VULKAN_HPP_INLINE typename ResultValueType<UniqueHandle<SurfaceKHR, Dispatch>>::type Instance::createUbmSurfaceSECUnique(
+    const UbmSurfaceCreateInfoSEC & createInfo, Optional<const AllocationCallbacks> allocator, Dispatch const & d ) const
   {
     VULKAN_HPP_ASSERT( d.getVkHeaderVersion() == VK_HEADER_VERSION );
 #      if ( VULKAN_HPP_DISPATCH_LOADER_DYNAMIC == 1 )
@@ -31913,9 +31911,8 @@ VULKAN_HPP_EXPORT namespace VULKAN_HPP_NAMESPACE
   // wrapper function for command vkGetPhysicalDeviceUbmPresentationSupportSEC, see
   // https://registry.khronos.org/vulkan/specs/latest/man/html/vkGetPhysicalDeviceUbmPresentationSupportSEC.html
   template <typename Dispatch, typename std::enable_if<IS_DISPATCHED( vkGetPhysicalDeviceUbmPresentationSupportSEC ), bool>::type>
-  VULKAN_HPP_INLINE Bool32 PhysicalDevice::getUbmPresentationSupportSEC( uint32_t            queueFamilyIndex,
-                                                                         struct ubm_device * ubm_device,
-                                                                         Dispatch const &    d ) const VULKAN_HPP_NOEXCEPT
+  VULKAN_HPP_INLINE Bool32 PhysicalDevice::getUbmPresentationSupportSEC( uint32_t queueFamilyIndex, struct ubm_device * ubm_device, Dispatch const & d )
+    const VULKAN_HPP_NOEXCEPT
   {
     VULKAN_HPP_ASSERT( d.getVkHeaderVersion() == VK_HEADER_VERSION );
     return static_cast<Bool32>(
@@ -31926,8 +31923,8 @@ VULKAN_HPP_EXPORT namespace VULKAN_HPP_NAMESPACE
   // wrapper function for command vkGetPhysicalDeviceUbmPresentationSupportSEC, see
   // https://registry.khronos.org/vulkan/specs/latest/man/html/vkGetPhysicalDeviceUbmPresentationSupportSEC.html
   template <typename Dispatch, typename std::enable_if<IS_DISPATCHED( vkGetPhysicalDeviceUbmPresentationSupportSEC ), bool>::type>
-  VULKAN_HPP_NODISCARD VULKAN_HPP_INLINE std::pair<Bool32, struct ubm_device>
-                                         PhysicalDevice::getUbmPresentationSupportSEC( uint32_t queueFamilyIndex, Dispatch const & d ) const VULKAN_HPP_NOEXCEPT
+  VULKAN_HPP_NODISCARD VULKAN_HPP_INLINE std::pair<Bool32, struct ubm_device> PhysicalDevice::getUbmPresentationSupportSEC(
+    uint32_t queueFamilyIndex, Dispatch const & d ) const VULKAN_HPP_NOEXCEPT
   {
     VULKAN_HPP_ASSERT( d.getVkHeaderVersion() == VK_HEADER_VERSION );
 #    if ( VULKAN_HPP_DISPATCH_LOADER_DYNAMIC == 1 )
@@ -31942,5 +31939,6 @@ VULKAN_HPP_EXPORT namespace VULKAN_HPP_NAMESPACE
   }
 #  endif /* VULKAN_HPP_DISABLE_ENHANCED_MODE */
 #endif   /*VK_USE_PLATFORM_UBM_SEC*/
+
 }  // namespace VULKAN_HPP_NAMESPACE
 #endif
--- a/vulkan/vulkan.cppm
+++ b/vulkan/vulkan.cppm
@@ -9,6 +9,8 @@ module;
 
 #define VULKAN_HPP_CXX_MODULE 1
 
+#include <cassert>
+#include <cstring>
 #include <vulkan/vulkan.h>
 #include <vulkan/vulkan_hpp_macros.hpp>
 
@@ -24,6 +26,8 @@ VULKAN_HPP_COMPILE_WARNING( VULKAN_HPP_CXX_MODULE_EXPERIMENTAL_WARNING )
 export module vulkan;
 
 export import std;
+
+VULKAN_HPP_STATIC_ASSERT( VK_HEADER_VERSION == 343, "Wrong VK_HEADER_VERSION!" );
 
 #if defined( _MSC_VER )
 #  pragma warning( push )
--- a/vulkan/vulkan_to_string.hpp
+++ b/vulkan/vulkan_to_string.hpp
@@ -11106,6 +11106,7 @@ VULKAN_HPP_EXPORT namespace VULKAN_HPP_NAMESPACE
     return "(void)";
   }
 #endif /*VK_USE_PLATFORM_UBM_SEC*/
+
 }  // namespace VULKAN_HPP_NAMESPACE
 
 #if defined( __clang__ ) || defined( __GNUC__ )


vulkan/vulkan_video.cppm
```diff
@@ -8,317 +8,64 @@ module;
...
-#include <vulkan/vulkan_hpp_macros.hpp>
-
-#if !defined( VULKAN_HPP_CXX_MODULE_EXPERIMENTAL_WARNING )
-#  define VULKAN_HPP_CXX_MODULE_EXPERIMENTAL_WARNING \
-    "\n\tThe Vulkan-Hpp C++ named module is experimental. It is subject to change without prior notice.\n" \
-  "\tTo silence this warning, define the VULKAN_HPP_CXX_MODULE_EXPERIMENTAL_WARNING macro.\n" \
-  "\tFor feedback, go to: https://github.com/KhronosGroup/Vulkan-Hpp/issues"
-
-VULKAN_HPP_COMPILE_WARNING( VULKAN_HPP_CXX_MODULE_EXPERIMENTAL_WARNING )
+#if __has_include( <vk_video/vulkan_video_codecs_common.h> )
```
M2-TE, line 11: Why do you use `__has_include` at all the include places?
I don't think there would ever be only partial availability for these on any system.
Anyone who has the Vulkan C headers, should also have these video codec headers, no?

Also, Clangd seems to be unable to handle `__has_include` for some reason.

M2-TE (commented):
Simply `VULKAN_HPP_EXPORT`ing the entire namespace is a good solution!

The real test would be to run the `Handles` and `ArrayWrapper` module tests (once you have the vulkan module changes), as those currently don't work due to lack of exported operators.

vulkan/vulkan_video.cppm
sharadhr: My initial attempt was to really just copy-paste what was originally done in the video header. I'm not entirely sure why it was done, need input from the maintainers here.

sharadhr:
I've actually got to the next step and passed all module tests locally, but it's a bit of a mess because `vulkan.hpp` is a common header between `module vulkan` and the video partition. 

We would have to drop the partition altogether, directly include `vulkan_video.hpp` in the main cppm, and guard the include of `vulkan.hpp`. 

Now, this is OK when manually editing the files (I should commit and push what I have, coming shortly...), but generating the headers is a problem, especially when it comes to the `has_include` stuff.

vulkan/vulkan_video.cppm
M2-TE: Oh I had no idea the video header already did this.. that is odd

M2-TE:
> I've actually got to the next step and passed all module tests locally

The tests I mentioned before are currently disabled, you have to manually uncomment them in `tests/CMakeLists.txt`. Unless you already did that, of course.

> We would have to drop the partition altogether, directly include vulkan_video.hpp in the main cppm, and guard the include of vulkan.hpp

I don't think that would be too much of a problem. The video module is kind of unnecessary when all the export statements are in the hpp headers anyways.

sharadhr:
> I don't think that would be too much of a problem. 

That's not the biggest issue, but it's generating video-related stuff in the `VulkanHppGenerator.cpp` that is more problematic. I suppose I can instantiate the object of `VideoHppGenerator.cpp` in the former, but I'd like to adhere to single-responsibility, and having the normal generator depend on the video stuff is... not that.

In fact it might be better for `vulkan_video.cppm` to return to being a separate module and `export import vulkan` rather than the other way round, because _that_ is the dependency order.

sharadhr:
OK, this latest set of patches implements that; we are back to two modules: `vulkan`, and `vulkan_video`. I am open to changing the name of the latter to `vulkan.video`, or `vulkan-video`, or any other delimiter. 

IMO this is the most logical separation, as it is clear the video module depends on Vulkan-Hpp core functionality, but the normal Vulkan module doesn't need any video-related functionality.

As a stretch goal I would also like to get rid of the `using PFN...` stuff, but that will require interacting with the C headers, which will probably be much more difficult than this.

sharadhr:
This is a _massive_ change to how the Vulkan-Hpp module works, so I want to tag everyone who has possibly used it to try this new version. We have tests (which we need to expand), but real-world testing is the best indicator.

@stripe2933, @M2-TE, @YaaZ, @dsvensson, @qbojj, @HildarTheDorf, @Zingam, and even @saschawillems (if you have time...) you are some of the more prolific module users I've interacted with since releasing this; please try this branch and see how well it works for you. If you've been using `import vulkan_hpp;`, you would probably have had to already change that to `import vulkan;`. You will additionally have to set up CMake with `import std` if you have not updated your headers. 

Thanks very much!

M2-TE:
Currently, the module breaks for me (Clang 21) when using the dynamic dispatcher. The `DispatchLoaderDynamic` is declared in `vulkan/vulkan_hpp_macros.hpp` within the global module fragment and defined in `vulkan/vulkan.hpp` in the module body. This leads to the error seen below: 
```
[build] In file included from vulkan/vulkan.cppm:37:
[build] vulkan/vulkan.hpp:21063:11: error: declaration of 'DispatchLoaderDynamic' in module vulkan follows declaration in the global module
[build]  21063 |     class DispatchLoaderDynamic : public DispatchLoaderBase
[build]        |           ^
[build] vulkan/vulkan_hpp_macros.hpp:288:11: note: previous declaration is here
[build]   288 |     class DispatchLoaderDynamic;
[build]       |           ^
```

Simply moving `vulkan/vulkan_hpp_macros.hpp` from the GMF to the module body does not work either, as it is _declared_ with internal linkage and _defined_ with export in the module body.
```
[build] In file included from vulkan/vulkan.cppm:38:
[build] vulkan/vulkan.hpp:21063:11: error: cannot export redeclaration 'DispatchLoaderDynamic' here since the previous declaration has module linkage
[build]  21063 |     class DispatchLoaderDynamic : public DispatchLoaderBase
[build]        |           ^
[build] vulkan/vulkan_hpp_macros.hpp:288:11: note: previous declaration is here
[build]   288 |     class DispatchLoaderDynamic;
[build]       |           ^
```

Could we use this opportunity to think about removing the need to have external storage for the dynamic dispatcher when using the module? It makes sense for the normal headers, but the module is its own TU.

sharadhr:
> Could we use this opportunity to think about removing the need to have external storage for the dynamic dispatcher when using the module? It makes sense for the normal headers, but the module is its own TU.

Yes; if the module is being used we can probably opt into having the storage directly in the module.

I've been thinking about the `vk::detail` namespace; this should be obviated with modules. And more importantly I'm not sure why the dispatcher is in a nested namespace at all; it's an advertised feature of Vulkan that should be in the top-level.

M2-TE:
You forgot to export the namespace with `vulkan_to_string.hpp`. I'll create a test for that one, since it should not have passed those.

> I've been thinking about the vk::detail namespace; this should be obviated with modules

Agreed, users often interact with it directly (initialization and storage in particular, sometimes even passing it to functions).

Easiest way I can get the dynamic dispatcher compiled within the module is by:
1. Remove the forward declarations from `vulkan_hpp_macros.hpp`. Only the macros should be there.
2. Add the forward declarations for `DispatchLoaderDynamic` and `defaultDispatchLoaderDynamic` to start of `vulkan.hpp`.
3. Create a `DispatchLoaderDynamic defaultDispatchLoaderDynamic` right after the definition of `DispatchLoaderDynamic`.

Of course, these should be hidden by similar `#if` gates as in the current macro header. Step 3) should also only apply to the module, so that's another gate to check for.

sharadhr:
I don't really use the dynamic dispatcher (I predominantly use `vk::raii`, which has its own dispatcher), so forgive me for the dumb question. Exactly what _are_ the forward declarations for? Can we get by with simply _not_ defining them for the module case, and just `export`ing the definitions, like we have done for everything else?

M2-TE:
The order declarations/definitions is a bit odd. As far as I can tell, it is a sort of circular dependency?
Certain functions and handles (predominantly `vulkan_handles.hpp`) need the _defined_ dispatcher, whereas the dispatcher definition needs some handles like `vk::Instance` and `vk::Device` for its respective initialization phases.

As an example if I do not forward declare `defaultDispatchLoaderDynamic` with `extern`:
```
[build] vulkan/vulkan_handles.hpp:5119:71: error: no member named 'defaultDispatchLoaderDynamic' in namespace 'vk::detail'; did you mean 'DispatchLoaderDynamic'?
[build]  5119 |                                        Dispatch const & d             VULKAN_HPP_DEFAULT_DISPATCHER_ASSIGNMENT ) const VULKAN_HPP_NOEXCEPT;
[build]       |                                                                       ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
[build] vulkan/vulkan_hpp_macros.hpp:335:81: note: expanded from macro 'VULKAN_HPP_DEFAULT_DISPATCHER_ASSIGNMENT'
[build]   335 | #define VULKAN_HPP_DEFAULT_DISPATCHER_ASSIGNMENT VULKAN_HPP_DEFAULT_ASSIGNMENT( VULKAN_HPP_DEFAULT_DISPATCHER )
[build]       |                                                                                 ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
[build] vulkan/vulkan_hpp_macros.hpp:315:75: note: expanded from macro 'VULKAN_HPP_DEFAULT_DISPATCHER'
[build]   315 | #    define VULKAN_HPP_DEFAULT_DISPATCHER ::VULKAN_HPP_NAMESPACE::detail::defaultDispatchLoaderDynamic
[build]       |                                                                           ^~~~~~~~~~~~~~~~~~~~~~~~~~~~
[build] vulkan/vulkan.hpp:91:11: note: 'DispatchLoaderDynamic' declared here
[build]    91 |     class DispatchLoaderDynamic;
[build]       |           ^
```

I can create a patch off of your branch with a working implementation of what I mentioned, if you want?

sharadhr:
> I can create a patch off of your branch with a working implementation of what I mentioned, if you want?

By all means; I can accept that in when you think it's ready. Cheers!

M2-TE:
The patch: [dyndis.patch](https://github.com/user-attachments/files/24716714/dyndis.patch)

What this does is three things:
1. Moves forward declarations out of `vulkan_hpp_macros.hpp` and into `vulkan.hpp`.
2. Places storage for the default dynamic dispatcher into the module (so long as users have not defined a custom one).
3. Updates tests to use the storage macro and also remove storage entirely when the module version is being tested.

These changes honestly belong into their own PR, as they touch on the normal headers as well. I don't want such a change to drown in an already big PR like this. Especially since it should also work with the current main branch modules (just gotta add the export keyword).

What do you think?

sharadhr:
I reckon if we want to separate it out, we should merge that in first, and then I can rebase my work on top of that. Although I suspect the rebase will be quite painful, and it might be easier to just add this patch in, it's not the biggest, really. 

I absolutely want to avoid a 1.4.335 situation where we had a _broken_ master branch, and it seems this might be such a case, because we haven't enabled those tests.

M2-TE:
Oh yea, the rebase might be painful.. in that case, just add it in now.
I've created issues #2447 and #2448 that would be closed by this PR (if you could add them to your "resolves" list).
Just so we have something to track that specific change in case it causes issues!

sharadhr:
Side question...

Could we make the module tests reuse the compiled module? As they are now, the test timings don't look very flattering because the module is recompiled for every test.

M2-TE:
I had separated them via ctest, because some tests need special `VULKAN_HPP_*` flags.

Previously I had made a sort of "hash map" in the root Vulkan-Hpp CMake, where tests with the same flags would share the same Vulkan module dependency.. but it was incredibly messy and error prone (leaking flags, etc..).

Perhaps this is worth revisiting..

sharadhr:
Fair enough. Although I'd really like to prepare some header-vs-module timings. Hopefully I can migrate the modern [Khronos samples](https://github.com/KhronosGroup/Vulkan-Samples) to this repository... some day.

M2-TE:
The tests would be a bad fit for timings either way. You would need an atleast somewhat larger project (like those samples) where multiple TUs otherwise include `vulkan.hpp` and std headers.

By the way, the current state of this PR is now fully working for me (Clang 21).

sharadhr:
> By the way, the current state of this PR is now fully working for me (Clang 21).

Woohoo! Works for me on MSVC too. 

@asuessenbach this is ready to run CI and for your review...

SaschaWillems:
I sadly don't have much experience with modules and rarely use them.

Tried to get this work on our version of the Vulkan Tutorial with MSVC 2022, but I'm stuck with the following error:

"vulkan.cppm(8,1): error C7577: a global module fragment can only appear at the start of a translation unit"

I have enabled both the  ISO C++23 langauge preview and Build ISO C++23 Standard Library modules

M2-TE:
@SaschaWillems that sounds like you have precompiled headers enabled. I think I had that issue for that reason before, since it places the pch at the beginning where the global module fragment is supposed to be.

SaschaWillems:
Just checked, precompiled headers are explicitly disable (option is set to "Not Using Precompiled Headers").

M2-TE:
Hm.. something is either being placed at the start of the module automatically, or you perhaps included `vulkan.cppm` somewhere on accident instead of using `import`? If it's none of those, @sharadhr compiles modules with MSVC so he might have an idea.

sharadhr:
> "vulkan.cppm(8,1): error C7577: a global module fragment can only appear at the start of a translation unit"

Hmm, I haven't seen this. If you have a branch I can clone and test, that would be great! Thanks for providing feedback.

sharadhr:
The CI isn't very encouraging either. :(

M2-TE:
Could you share a link to failed runs, if any? If you mean the stale CI on this PR, you probably need to either push an empty commit or open/close it.

sharadhr:
Oh, hmm. It _is_ [stale CI](https://github.com/KhronosGroup/Vulkan-Hpp/actions/runs/21132316644) from almost yesterday; I was very confused for a moment because it was a bug we fixed.

HildarTheDorf:
I've attempted to compile just an empty file with `import vulkan;` and had no success with the latest cmake (4.2.1) in combination with clang-21 and gcc-15 (and their respective libc++/libstdc++ versions).

I think this is a tooling or libstdc++ issue with `export import std;`, as the error messages all point to conflicting definitions of everything in std (the two versions being via #include vs import).

Cmake file used for reference:

```cmake
cmake_minimum_required(VERSION 4.2)
set(CMAKE_EXPERIMENTAL_CXX_IMPORT_STD "d0edc3af-4c50-42ea-a356-e2862fe7a444")
set(CMAKE_CXX_MODULE_STD ON)

project(vkmod)
include(FetchContent)

set(CMAKE_CXX_STANDARD 23)
set(CMAKE_CXX_STANDARD_REQUIRED YES)
#set(CMAKE_CXX_FLAGS -stdlib=libc++)

add_executable(vkmod main.cpp)
target_compile_definitions(vkmod PRIVATE VULKAN_HPP_CXX_MODULE_EXPERIMENTAL_WARNING)
target_sources(vkmod PRIVATE FILE_SET vulkan_hpp TYPE CXX_MODULES BASE_DIRS ${CMAKE_CURRENT_SOURCE_DIR} FILES ${CMAKE_CURRENT_SOURCE_DIR}/vulkan.cppm)
```

EDIT: Further to the above, compiling directly without CMake results in the same issue. "Conflicting definitions" of a large chunk of the standard library, with every definition pointing to the exact same position/include path. I can compile "Hello World" with `import std;` fine as long as I don't also #include any standard headers.

EDIT2: Looks like a possible bug or unexpected behaviour in clang. `#include <>` inside a module is ignoring `-isystem` and using the installed version of `vulkan.hpp` in `/usr`. This fails so hard it never even gets to the vulkan header version check before bailing from >200 errors!

HildarTheDorf:
After much issue with tooling, I have things compiling on clang-21/libc++21.

I agree the dynamic dispatcher storage is a pain point. It should just vanish with modules. But if it needs changes to the non-module headers, I agree with making a separate PR (and leaving the module as experimental at least until that's done). That's my 2 cents anyhow.

stripe2933:
Failed on all platforms for building [vk-gltf-viewer](https://github.com/stripe2933/vk-gltf-viewer). [CI](https://github.com/stripe2933/vk-gltf-viewer/actions/runs/21165807103)

Notable observations:

1. Dynamic dispatch loader is not supported, as discussed.
2. When `VK_USE_PLATFORM_WIN32_KHR` is defined, including `<windows.h>` in the module purview seems problematic.
3. Both Clang 19 and Clang 21 confuse for multiple symbols declaration of the system header.

```
In file included from vk-gltf-viewer/build/debug/vcpkg_installed/arm64-macos-homebrew-clang/include/vulkan/vulkan.cppm:37:
In file included from vk-gltf-viewer/build/debug/vcpkg_installed/arm64-macos-homebrew-clang/include/vulkan/vulkan.hpp:39:
In file included from /opt/homebrew/Cellar/llvm/21.1.5/bin/../include/c++/v1/cassert:26:
In file included from /Library/Developer/CommandLineTools/SDKs/MacOSX15.sdk/usr/include/assert.h:44:
In file included from /opt/homebrew/Cellar/llvm/21.1.5/bin/../include/c++/v1/stdlib.h:89:
In file included from /Library/Developer/CommandLineTools/SDKs/MacOSX15.sdk/usr/include/stdlib.h:58:
In file included from /Library/Developer/CommandLineTools/SDKs/MacOSX15.sdk/usr/include/_stdlib.h:70:
In file included from /Library/Developer/CommandLineTools/SDKs/MacOSX15.sdk/usr/include/sys/wait.h:110:
/Library/Developer/CommandLineTools/SDKs/MacOSX15.sdk/usr/include/sys/resource.h:575:31: error: declaration of 'rlimit' in the global module follows declaration in module vulkan
  575 | int     getrlimit(int, struct rlimit *) __DARWIN_ALIAS(getrlimit);
      |                               ^
/Library/Developer/CommandLineTools/SDKs/MacOSX15.sdk/usr/include/sys/resource.h:464:8: note: previous declaration is here
  464 | struct rlimit {
      |        ^
/Library/Developer/CommandLineTools/SDKs/MacOSX15.sdk/usr/include/sys/resource.h:575:9: error: conflicting types for 'getrlimit'
  575 | int     getrlimit(int, struct rlimit *) __DARWIN_ALIAS(getrlimit);
      |         ^
/Library/Developer/CommandLineTools/SDKs/MacOSX15.sdk/usr/include/sys/resource.h:575:9: note: previous declaration is here
  575 | int     getrlimit(int, struct rlimit *) __DARWIN_ALIAS(getrlimit);
      |         ^
In file included from vk-gltf-viewer/build/debug/vcpkg_installed/arm64-macos-homebrew-clang/include/vulkan/vulkan.cppm:37:
In file included from vk-gltf-viewer/build/debug/vcpkg_installed/arm64-macos-homebrew-clang/include/vulkan/vulkan.hpp:39:
In file included from /opt/homebrew/Cellar/llvm/21.1.5/bin/../include/c++/v1/cassert:26:
In file included from /Library/Developer/CommandLineTools/SDKs/MacOSX15.sdk/usr/include/assert.h:44:
In file included from /opt/homebrew/Cellar/llvm/21.1.5/bin/../include/c++/v1/stdlib.h:89:
In file included from /Library/Developer/CommandLineTools/SDKs/MacOSX15.sdk/usr/include/stdlib.h:58:
In file included from /Library/Developer/CommandLineTools/SDKs/MacOSX15.sdk/usr/include/_stdlib.h:70:
In file included from /Library/Developer/CommandLineTools/SDKs/MacOSX15.sdk/usr/include/sys/wait.h:110:
/Library/Developer/CommandLineTools/SDKs/MacOSX15.sdk/usr/include/sys/resource.h:576:31: error: declaration of 'rusage' in the global module follows declaration in module vulkan
  576 | int     getrusage(int, struct rusage *);
      |                               ^
/Library/Developer/CommandLineTools/SDKs/MacOSX15.sdk/usr/include/sys/resource.h:152:9: note: previous declaration is here
  152 | struct  rusage {
      |         ^
/Library/Developer/CommandLineTools/SDKs/MacOSX15.sdk/usr/include/sys/resource.h:576:9: error: conflicting types for 'getrusage'
  576 | int     getrusage(int, struct rusage *);
      |         ^
/Library/Developer/CommandLineTools/SDKs/MacOSX15.sdk/usr/include/sys/resource.h:576:9: note: previous declaration is here
  576 | int     getrusage(int, struct rusage *);
      |         ^
In file included from vk-gltf-viewer/build/debug/vcpkg_installed/arm64-macos-homebrew-clang/include/vulkan/vulkan.cppm:37:
In file included from vk-gltf-viewer/build/debug/vcpkg_installed/arm64-macos-homebrew-clang/include/vulkan/vulkan.hpp:39:
In file included from /opt/homebrew/Cellar/llvm/21.1.5/bin/../include/c++/v1/cassert:26:
In file included from /Library/Developer/CommandLineTools/SDKs/MacOSX15.sdk/usr/include/assert.h:44:
In file included from /opt/homebrew/Cellar/llvm/21.1.5/bin/../include/c++/v1/stdlib.h:89:
In file included from /Library/Developer/CommandLineTools/SDKs/MacOSX15.sdk/usr/include/stdlib.h:58:
In file included from /Library/Developer/CommandLineTools/SDKs/MacOSX15.sdk/usr/include/_stdlib.h:70:
In file included from /Library/Developer/CommandLineTools/SDKs/MacOSX15.sdk/usr/include/sys/wait.h:110:
/Library/Developer/CommandLineTools/SDKs/MacOSX15.sdk/usr/include/sys/resource.h:581:37: error: declaration of 'rlimit' in the global module follows declaration in module vulkan
  581 | int     setrlimit(int, const struct rlimit *) __DARWIN_ALIAS(setrlimit);
      |                                     ^
/Library/Developer/CommandLineTools/SDKs/MacOSX15.sdk/usr/include/sys/resource.h:464:8: note: previous declaration is here
  464 | struct rlimit {
      |        ^
/Library/Developer/CommandLineTools/SDKs/MacOSX15.sdk/usr/include/sys/resource.h:581:9: error: conflicting types for 'setrlimit'
  581 | int     setrlimit(int, const struct rlimit *) __DARWIN_ALIAS(setrlimit);
      |         ^
/Library/Developer/CommandLineTools/SDKs/MacOSX15.sdk/usr/include/sys/resource.h:581:9: note: previous declaration is here
  581 | int     setrlimit(int, const struct rlimit *) __DARWIN_ALIAS(setrlimit);
      |         ^
In file included from vk-gltf-viewer/build/debug/vcpkg_installed/arm64-macos-homebrew-clang/include/vulkan/vulkan.cppm:37:
In file included from vk-gltf-viewer/build/debug/vcpkg_installed/arm64-macos-homebrew-clang/include/vulkan/vulkan.hpp:39:
In file included from /opt/homebrew/Cellar/llvm/21.1.5/bin/../include/c++/v1/cassert:26:
In file included from /Library/Developer/CommandLineTools/SDKs/MacOSX15.sdk/usr/include/assert.h:44:
In file included from /opt/homebrew/Cellar/llvm/21.1.5/bin/../include/c++/v1/stdlib.h:89:
In file included from /Library/Developer/CommandLineTools/SDKs/MacOSX15.sdk/usr/include/stdlib.h:58:
In file included from /Library/Developer/CommandLineTools/SDKs/MacOSX15.sdk/usr/include/_stdlib.h:70:
/Library/Developer/CommandLineTools/SDKs/MacOSX15.sdk/usr/include/sys/wait.h:249:9: error: conflicting types for 'waitid'
  249 | int     waitid(idtype_t, id_t, siginfo_t *, int) __DARWIN_ALIAS_C(waitid);
      |         ^
/Library/Developer/CommandLineTools/SDKs/MacOSX15.sdk/usr/include/sys/wait.h:249:9: note: previous declaration is here
  249 | int     waitid(idtype_t, id_t, siginfo_t *, int) __DARWIN_ALIAS_C(waitid);
      |         ^
In file included from vk-gltf-viewer/build/debug/vcpkg_installed/arm64-macos-homebrew-clang/include/vulkan/vulkan.cppm:37:
In file included from vk-gltf-viewer/build/debug/vcpkg_installed/arm64-macos-homebrew-clang/include/vulkan/vulkan.hpp:39:
In file included from /opt/homebrew/Cellar/llvm/21.1.5/bin/../include/c++/v1/cassert:26:
In file included from /Library/Developer/CommandLineTools/SDKs/MacOSX15.sdk/usr/include/assert.h:44:
In file included from /opt/homebrew/Cellar/llvm/21.1.5/bin/../include/c++/v1/stdlib.h:89:
In file included from /Library/Developer/CommandLineTools/SDKs/MacOSX15.sdk/usr/include/stdlib.h:58:
In file included from /Library/Developer/CommandLineTools/SDKs/MacOSX15.sdk/usr/include/_stdlib.h:70:
/Library/Developer/CommandLineTools/SDKs/MacOSX15.sdk/usr/include/sys/wait.h:252:34: error: declaration of 'rusage' in the global module follows declaration in module vulkan
  252 | pid_t   wait3(int *, int, struct rusage *);
      |                                  ^
/Library/Developer/CommandLineTools/SDKs/MacOSX15.sdk/usr/include/sys/resource.h:152:9: note: previous declaration is here
  152 | struct  rusage {
      |         ^
In file included from vk-gltf-viewer/build/debug/vcpkg_installed/arm64-macos-homebrew-clang/include/vulkan/vulkan.cppm:37:
In file included from vk-gltf-viewer/build/debug/vcpkg_installed/arm64-macos-homebrew-clang/include/vulkan/vulkan.hpp:39:
In file included from /opt/homebrew/Cellar/llvm/21.1.5/bin/../include/c++/v1/cassert:26:
In file included from /Library/Developer/CommandLineTools/SDKs/MacOSX15.sdk/usr/include/assert.h:44:
In file included from /opt/homebrew/Cellar/llvm/21.1.5/bin/../include/c++/v1/stdlib.h:89:
In file included from /Library/Developer/CommandLineTools/SDKs/MacOSX15.sdk/usr/include/stdlib.h:58:
In file included from /Library/Developer/CommandLineTools/SDKs/MacOSX15.sdk/usr/include/_stdlib.h:70:
/Library/Developer/CommandLineTools/SDKs/MacOSX15.sdk/usr/include/sys/wait.h:252:9: error: conflicting types for 'wait3'
  252 | pid_t   wait3(int *, int, struct rusage *);
      |         ^
/Library/Developer/CommandLineTools/SDKs/MacOSX15.sdk/usr/include/sys/wait.h:252:9: note: previous declaration is here
  252 | pid_t   wait3(int *, int, struct rusage *);
      |         ^
In file included from vk-gltf-viewer/build/debug/vcpkg_installed/arm64-macos-homebrew-clang/include/vulkan/vulkan.cppm:37:
In file included from vk-gltf-viewer/build/debug/vcpkg_installed/arm64-macos-homebrew-clang/include/vulkan/vulkan.hpp:39:
In file included from /opt/homebrew/Cellar/llvm/21.1.5/bin/../include/c++/v1/cassert:26:
In file included from /Library/Developer/CommandLineTools/SDKs/MacOSX15.sdk/usr/include/assert.h:44:
In file included from /opt/homebrew/Cellar/llvm/21.1.5/bin/../include/c++/v1/stdlib.h:89:
In file included from /Library/Developer/CommandLineTools/SDKs/MacOSX15.sdk/usr/include/stdlib.h:58:
In file included from /Library/Developer/CommandLineTools/SDKs/MacOSX15.sdk/usr/include/_stdlib.h:70:
/Library/Developer/CommandLineTools/SDKs/MacOSX15.sdk/usr/include/sys/wait.h:253:41: error: declaration of 'rusage' in the global module follows declaration in module vulkan
  253 | pid_t   wait4(pid_t, int *, int, struct rusage *);
      |                                         ^
/Library/Developer/CommandLineTools/SDKs/MacOSX15.sdk/usr/include/sys/resource.h:152:9: note: previous declaration is here
  152 | struct  rusage {
      |         ^
In file included from vk-gltf-viewer/build/debug/vcpkg_installed/arm64-macos-homebrew-clang/include/vulkan/vulkan.cppm:37:
In file included from vk-gltf-viewer/build/debug/vcpkg_installed/arm64-macos-homebrew-clang/include/vulkan/vulkan.hpp:39:
In file included from /opt/homebrew/Cellar/llvm/21.1.5/bin/../include/c++/v1/cassert:26:
In file included from /Library/Developer/CommandLineTools/SDKs/MacOSX15.sdk/usr/include/assert.h:44:
In file included from /opt/homebrew/Cellar/llvm/21.1.5/bin/../include/c++/v1/stdlib.h:89:
In file included from /Library/Developer/CommandLineTools/SDKs/MacOSX15.sdk/usr/include/stdlib.h:58:
In file included from /Library/Developer/CommandLineTools/SDKs/MacOSX15.sdk/usr/include/_stdlib.h:70:
/Library/Developer/CommandLineTools/SDKs/MacOSX15.sdk/usr/include/sys/wait.h:253:9: error: conflicting types for 'wait4'
  253 | pid_t   wait4(pid_t, int *, int, struct rusage *);
      |         ^
/Library/Developer/CommandLineTools/SDKs/MacOSX15.sdk/usr/include/sys/wait.h:253:9: note: previous declaration is here
  253 | pid_t   wait4(pid_t, int *, int, struct rusage *);
      |         ^
In file included from vk-gltf-viewer/build/debug/vcpkg_installed/arm64-macos-homebrew-clang/include/vulkan/vulkan.cppm:37:
In file included from vk-gltf-viewer/build/debug/vcpkg_installed/arm64-macos-homebrew-clang/include/vulkan/vulkan.hpp:39:
In file included from /opt/homebrew/Cellar/llvm/21.1.5/bin/../include/c++/v1/cassert:26:
In file included from /Library/Developer/CommandLineTools/SDKs/MacOSX15.sdk/usr/include/assert.h:44:
In file included from /opt/homebrew/Cellar/llvm/21.1.5/bin/../include/c++/v1/stdlib.h:89:
In file included from /Library/Developer/CommandLineTools/SDKs/MacOSX15.sdk/usr/include/stdlib.h:58:
/Library/Developer/CommandLineTools/SDKs/MacOSX15.sdk/usr/include/_stdlib.h:159:8: error: conflicting types for 'div'
  159 | div_t    div(int, int) __pure2;
      |          ^
/Library/Developer/CommandLineTools/SDKs/MacOSX15.sdk/usr/include/_stdlib.h:159:8: note: previous declaration is here
  159 | div_t    div(int, int) __pure2;
      | ~~~~~    ^
In file included from vk-gltf-viewer/build/debug/vcpkg_installed/arm64-macos-homebrew-clang/include/vulkan/vulkan.cppm:37:
In file included from vk-gltf-viewer/build/debug/vcpkg_installed/arm64-macos-homebrew-clang/include/vulkan/vulkan.hpp:39:
In file included from /opt/homebrew/Cellar/llvm/21.1.5/bin/../include/c++/v1/cassert:26:
In file included from /Library/Developer/CommandLineTools/SDKs/MacOSX15.sdk/usr/include/assert.h:44:
In file included from /opt/homebrew/Cellar/llvm/21.1.5/bin/../include/c++/v1/stdlib.h:89:
In file included from /Library/Developer/CommandLineTools/SDKs/MacOSX15.sdk/usr/include/stdlib.h:58:
/Library/Developer/CommandLineTools/SDKs/MacOSX15.sdk/usr/include/_stdlib.h:164:9: error: conflicting types for 'ldiv'
  164 | ldiv_t   ldiv(long, long) __pure2;
      |          ^
/Library/Developer/CommandLineTools/SDKs/MacOSX15.sdk/usr/include/_stdlib.h:164:9: note: previous declaration is here
  164 | ldiv_t   ldiv(long, long) __pure2;
      | ~~~~~~   ^
In file included from vk-gltf-viewer/build/debug/vcpkg_installed/arm64-macos-homebrew-clang/include/vulkan/vulkan.cppm:37:
In file included from vk-gltf-viewer/build/debug/vcpkg_installed/arm64-macos-homebrew-clang/include/vulkan/vulkan.hpp:39:
In file included from /opt/homebrew/Cellar/llvm/21.1.5/bin/../include/c++/v1/cassert:26:
In file included from /Library/Developer/CommandLineTools/SDKs/MacOSX15.sdk/usr/include/assert.h:44:
In file included from /opt/homebrew/Cellar/llvm/21.1.5/bin/../include/c++/v1/stdlib.h:89:
In file included from /Library/Developer/CommandLineTools/SDKs/MacOSX15.sdk/usr/include/stdlib.h:58:
/Library/Developer/CommandLineTools/SDKs/MacOSX15.sdk/usr/include/_stdlib.h:168:10: error: conflicting types for 'lldiv'
  168 | lldiv_t  lldiv(long long, long long);
      |          ^
/Library/Developer/CommandLineTools/SDKs/MacOSX15.sdk/usr/include/_stdlib.h:168:10: note: previous declaration is here
  168 | lldiv_t  lldiv(long long, long long);
      | ~~~~~~~  ^
In file included from vk-gltf-viewer/build/debug/vcpkg_installed/arm64-macos-homebrew-clang/include/vulkan/vulkan.cppm:37:
In file included from vk-gltf-viewer/build/debug/vcpkg_installed/arm64-macos-homebrew-clang/include/vulkan/vulkan.hpp:39:
In file included from /opt/homebrew/Cellar/llvm/21.1.5/bin/../include/c++/v1/cassert:26:
In file included from /Library/Developer/CommandLineTools/SDKs/MacOSX15.sdk/usr/include/assert.h:44:
/opt/homebrew/Cellar/llvm/21.1.5/bin/../include/c++/v1/stdlib.h:126:37: error: functions that differ only in their return type cannot be overloaded
  126 | inline _LIBCPP_HIDE_FROM_ABI ldiv_t div(long __x, long __y) _NOEXCEPT { return ::ldiv(__x, __y); }
      |                              ~~~~~~ ^
/opt/homebrew/Cellar/llvm/21.1.5/bin/../include/c++/v1/stdlib.h:126:37: note: previous definition is here
  126 | inline _LIBCPP_HIDE_FROM_ABI ldiv_t div(long __x, long __y) _NOEXCEPT { return ::ldiv(__x, __y); }
      |                              ~~~~~~ ^
In file included from vk-gltf-viewer/build/debug/vcpkg_installed/arm64-macos-homebrew-clang/include/vulkan/vulkan.cppm:37:
In file included from vk-gltf-viewer/build/debug/vcpkg_installed/arm64-macos-homebrew-clang/include/vulkan/vulkan.hpp:39:
In file included from /opt/homebrew/Cellar/llvm/21.1.5/bin/../include/c++/v1/cassert:26:
In file included from /Library/Developer/CommandLineTools/SDKs/MacOSX15.sdk/usr/include/assert.h:44:
/opt/homebrew/Cellar/llvm/21.1.5/bin/../include/c++/v1/stdlib.h:128:38: error: functions that differ only in their return type cannot be overloaded
  128 | inline _LIBCPP_HIDE_FROM_ABI lldiv_t div(long long __x, long long __y) _NOEXCEPT { return ::lldiv(__x, __y); }
      |                              ~~~~~~~ ^
/opt/homebrew/Cellar/llvm/21.1.5/bin/../include/c++/v1/stdlib.h:128:38: note: previous definition is here
  128 | inline _LIBCPP_HIDE_FROM_ABI lldiv_t div(long long __x, long long __y) _NOEXCEPT { return ::lldiv(__x, __y); }
      |                              ~~~~~~~ ^
1 warning and 16 errors generated.
```

https://github.com/sharadhr/Vulkan-Hpp/blob/86ddf36d1b57a2562f386ede413afaf35b51563b/vulkan/vulkan.cppm#L37

https://github.com/sharadhr/Vulkan-Hpp/blob/86ddf36d1b57a2562f386ede413afaf35b51563b/vulkan/vulkan.hpp#L39-L40

Based on my experience, C-compatibility header should not be included inside the module purview. They may contains the system header inclusion and will cause the ODR violation.

4. It fixes the structured binding decomposability of `vk::StructureChain` for MSVC.

M2-TE:
> Dynamic dispatch loader is not supported, as discussed.

It should be supported though; we added a patch where dynamic dispatcher works as before when not using modules. When using modules, you do not even have to create storage for it anymore, as it is now automated. If you do still declare storage for it, things will break of course, as it can only exist once.

> When VK_USE_PLATFORM_WIN32_KHR is defined, including <windows.h> in the module purview seems problematic.

~That is indeed interesting.. while I assume that people building x86 windows with module toolchains are in the minority, it is still something we should probably add a test for.~

> Both Clang 19 and Clang 21 confuse for multiple symbols declaration of the system header.

There seems to be a discrepency between Vulkan-Hpp and VulkanMemoryAllocator-Hpp.. My guess would be either a C++ version difference or accidental mixing of import std and traditional headers. In my project with both of those mixed as well, things compile without much issue (granted, you need the newest VMA headers containing fixes from GPUOpen-LibrariesAndSDKs/VulkanMemoryAllocator#514).

> Based on my experience, C-compatibility header should not be included inside the module purview. They may contains the system header inclusion and will cause the ODR violation.

You are right, those should be moved out of that specific `#if` branch in vulkan.hpp and into the vulkan.cppm GMF.

I also spot a MacOS build failure in your CI, which I recognize from dabbling with that CI runner. Check out [.github/workflows/build.yaml](.github/workflows/build.yaml) to see what steps we take to set up modules for each platform. All platforms seem to succeed with modules there (see [https://github.com/M2-TE/Vulkan-Hpp/actions/runs/21165200678](https://github.com/M2-TE/Vulkan-Hpp/actions/runs/21165200678)).

I'm gonna take a closer look at your repo to try and trace those issues.

sharadhr:
Thanks everyone for your testing inputs. Clearly there is a documentation gap; I am addressing this so others who use the module can get started much quicker. 

Tasks ahead: 

1. Move all of the `#include`s into the module GMF; this includes (heh) the Windows API header, and the C headers;
2. VMA may need to rethink its own manner of defining its module; they probably need a refactor as well.

sharadhr:
> building x86 windows with module toolchains are in the minority

In this case, `WIN32` refers to the [Win32 API](https://learn.microsoft.com/en-gb/windows/win32/api/), which 64-bit Windows also provides (the '32' is just to differentiate from Win16, which was a pre-NT API). This is used to provide a native Direct2D/GDI+ surface for Vulkan to render into.

HildarTheDorf:
@stripe2933 Those errors look a lot like the errors I had when vulkan.cppm was incorrectly including the system vulkan.hpp (321 on my machine) not the correct version (338 as of writing)

Nothing should be doing `#include <>` in module purview unless the code is designed to be part of a module or contains no decls. That is exclusively `vulkan_hpp_macros.hpp` as far as I can tell.

M2-TE:
@sharadhr perhaps you could add a version comparison into `vulkan.cppm`, similar to what is in `vulkan.hpp`? It should be easy to update the generator and have it spit out a `static_assert` for us. 

At the very least, it would better clarify cases where header/module versions do not match. Some of the errors resulting from mismatches can be rather.. strange.

HildarTheDorf:
Also, speaking of `vulkan_hpp_macros.hpp` you can't include it both inside and outside module purview in different TUs when using the dynamic dispatcher, as it then contains the decls for said dispatcher.

sharadhr:
> perhaps you could add a version comparison into vulkan.cppm, similar to what is in vulkan.hpp?

Good idea. Though... what should we compare it to? 

> as it then contains the decls for said dispatcher.

Perhaps this means another `if !defined ( VULKAN_HPP_CXX_MODULE )` gate.

M2-TE:
@HildarTheDorf how would that occur? From what I can tell, it is only included in the global module fragment in `vulkan.cppm` and user code can still include it in their own GMF. Besides, it no longer contains dispatcher forward declarations, as those are now inside `vulkan.hpp`.

@sharadhr compare it to the version in vulkan.hpp; use the same source in the generator for that version number. If successful, module, C++, and C headers should all be at the same version.

HildarTheDorf:
> @HildarTheDorf how would that occur? From what I can tell, it is only included in the global module fragment in `vulkan.cppm` and user code can still include it in their own GMF. Besides, it no longer contains dispatcher forward declarations, as those are now inside `vulkan.hpp`.

By me being dumb and putting it under `export module` instead of in the GMF. Clang does warn.
Combing modular code and 'classic' code is just a pain in general, it's not a problem with vulkan.cppm. Sorry for confusion.

sharadhr:
> When VK_USE_PLATFORM_WIN32_KHR is defined, including <windows.h> in the module purview seems problematic.

I'm looking for this, and I can't find an explicit include in any of the generated headers (or I'm not looking hard enough). Is this a transitive include from the two C headers?

M2-TE:
Yea I think this is only a C header thing:
https://github.com/KhronosGroup/Vulkan-Headers/blob/0007545ec811852c8ea6154f201624583c88563f/include/vulkan/vulkan.h#L45-L48

Since you placed `vulkan.h` in the GMF, you can probably hide the `vulkan.h` include in `vulkan.hpp` behind the `!defined( VULKAN_HPP_CXX_MODULE )` like the others.

sharadhr:
> you can probably hide the vulkan.h include

Which I have [done here](https://github.com/sharadhr/Vulkan-Hpp/blob/86ddf36d1b57a2562f386ede413afaf35b51563b/snippets/includes.hpp#L2).

M2-TE:
Oh my bad, was looking at the wrong branch. I'm not sure where that windows header could otherwise be getting included then, other than those two C headers. Though either way, those should likely be moved into the GMF as well.

dsvensson:
@sharadhr Will have a look. I don't suppose you have some discord or so for more interactive communication?

M2-TE:
The `static_assert` version check needs to be moved into module purview for MSVC I reckon.

sharadhr:
> The static_assert version check needs to be moved into module purview for MSVC I reckon.

It does, but I also suppose we need to suppress the actual error, (which is a real version mismatch) in CI. The build works locally, for the record.

M2-TE:
Not sure how it could end up using the wrong header though. It correctly uses the `Vulkan-Hpp/vulkan/` for its primary include: `-ID:\a\Vulkan-Hpp\Vulkan-Hpp -ID:\a\Vulkan-Hpp\Vulkan-Hpp\Vulkan-Headers\include`.
The path for the compiled module is also correct: `CMakeFiles/VulkanHppModule.dir/vulkan/vulkan.cppm.obj`.

Regardless, even the headers in `Vulkan-Headers/include` use the same version. Perhaps it is just a spurious failure due to being in the GMF?

Full output for reference:
```
 FAILED: [code=2] CMakeFiles/VulkanHppModule.dir/vulkan/vulkan.cppm.obj CMakeFiles/VulkanHppModule.dir/vulkan.ifc 
C:\PROGRA~1\MICROS~2\2022\ENTERP~1\VC\Tools\MSVC\1444~1.352\bin\Hostx64\x64\cl.exe  /nologo /TP -DVK_NO_PROTOTYPES -DVULKAN_HPP_DISPATCH_LOADER_DYNAMIC -ID:\a\Vulkan-Hpp\Vulkan-Hpp -ID:\a\Vulkan-Hpp\Vulkan-Hpp\Vulkan-Headers\include /EHsc /Zi /Ob0 /Od /RTC1 -std:c++latest -MDd /showIncludes @CMakeFiles\VulkanHppModule.dir\vulkan\vulkan.cppm.obj.modmap /FoCMakeFiles\VulkanHppModule.dir\vulkan\vulkan.cppm.obj /FdCMakeFiles\VulkanHppModule.dir\VulkanHppModule.pdb /FS -c D:\a\Vulkan-Hpp\Vulkan-Hpp\vulkan\vulkan.cppm
vulkan.cppm
D:\a\Vulkan-Hpp\Vulkan-Hpp\vulkan\vulkan.cppm(23): warning: 
	The Vulkan-Hpp C++ named module is experimental. It is subject to change without prior notice.
	To silence this warning, define the VULKAN_HPP_CXX_MODULE_EXPERIMENTAL_WARNING macro.
	For feedback, go to: https://github.com/KhronosGroup/Vulkan-Hpp/issues
D:\a\Vulkan-Hpp\Vulkan-Hpp\vulkan\vulkan.cppm(26): error C2338: static_assert failed: 'Wrong VK_HEADER_VERSION!'
D:\a\Vulkan-Hpp\Vulkan-Hpp\vulkan\vulkan.cppm(26): warning C5202: a global module fragment can only contain preprocessor directives
```

sharadhr:
I reckon there are two failures; notice how the error came before the warning (probably suggesting that the error was parsed and emitted first, then the warning, which refers to the macro that surrounds it).

I moved the assert to the module purview anyway, but I'm not certain it'll fix it.

M2-TE:
The reason for the failure lies with CI atm. There is a typo in the bash line for selecting `ON` and `OFF` for header generation that I introduced in #2441. This will be fixed by #2455.

Although I _do_ wonder why only the windows runner was affected and why it didn't cause problems before..

sharadhr:
This is blocked by #2457.

sharadhr:
This PR is now ready for review and merge.

asuessenbach:
@sharadhr Would be great, if you could resolve the conflicts.

VideoHppGenerator.cpp
```diff
@@ -36,7 +36,7 @@ VideoHppGenerator::VideoHppGenerator( tinyxml2::XMLDocument const & document )
 void VideoHppGenerator::generateCppmFile() const
 {
   generateFileFromTemplate(
-    "vulkan_video.cppm", "VideoCppmTemplate.hpp", { { "copyrightMessage", m_copyrightMessage }, { "usings", generateCppModuleUsings() } } );
```
asuessenbach, line 39: That is, `generateCppModuleUsings` (and the functions called in there) aren't used anymore?
Could be removed then.

VulkanHppGenerator.cpp
```diff
@@ -139,10 +139,9 @@ void VulkanHppGenerator::generateCppmFile() const
   generateFileFromTemplate( m_api + ".cppm",
                             "CppmTemplate.hpp",
                             { { "api", m_api },
-                              { "hashSpecializations", generateCppModuleHashSpecializations() },
+                              { "vulkan_h", ( m_api == "vulkansc" ) ? "vulkan_sc_core.h" : ( m_api + ".h" ) },
                               { "licenseHeader", m_vulkanLicenseHeader },
-                              { "raiiUsings", generateCppModuleRaiiUsings() },
-                              { "usings", generateCppModuleUsings() },
```
asuessenbach, line 147: `generateCppModuleHashSpecializations` and `generateCppModuleRaiiUsings` aren't used any more?
Should be removed then.

snippets/FormatTraits.hpp
```diff
@@ -4,93 +4,99 @@
 
   //=== Function Declarations ===
 
+  #if defined( VULKAN_HPP_CXX_MODULE )
+  #define VULKAN_HPP_EXPORT export
+  #else
+  #define VULKAN_HPP_EXPORT
+  #endif
```
asuessenbach: Is that needed here, at all?
Isn't it already set in FormatTraitsHppTemplate.hpp?

snippets/FormatTraits.hpp
```diff
@@ -4,93 +4,99 @@
...
+
   // The three-dimensional extent of a texel block.
-  VULKAN_HPP_CONSTEXPR_14 std::array<uint8_t, 3> blockExtent( Format format );
+  VULKAN_HPP_EXPORT VULKAN_HPP_CONSTEXPR_14 std::array<uint8_t, 3> blockExtent( Format format );
```
asuessenbach, line 8: Why do you use `VULKAN_HPP_EXPORT` on each function here, instead of on namespace level as in FormatTraitsHppTemplate.hpp?

snippets/FormatTraitsHppTemplate.hpp
```diff
@@ -3,10 +3,15 @@
 #ifndef VULKAN_FORMAT_TRAITS_HPP
 #  define VULKAN_FORMAT_TRAITS_HPP
 
-#include <vulkan/${api}.hpp>
+#if defined( VULKAN_HPP_CXX_MODULE )
+#  define VULKAN_HPP_EXPORT export
+#else
+#  define VULKAN_HPP_EXPORT
+#  include <vulkan/${api}.hpp>
```
asuessenbach, line 9: Isn't `VULKAN_HPP_EXPORT` re-defined by including vulkan.hpp?

snippets/HashHppTemplate.hpp
```diff
@@ -3,9 +3,11 @@
 #ifndef VULKAN_HASH_HPP
 #  define VULKAN_HASH_HPP
 
-#include <vulkan/${api}.hpp>
+#if !defined( VULKAN_HPP_CXX_MODULE )
+#  include <vulkan/${api}.hpp>
+#endif
 
-namespace std
+VULKAN_HPP_EXPORT namespace std
```
asuessenbach, line 10: Don't you need to define `VULKAN_HPP_EXPORT`, when `VULKAN_HPP_CXX_MODULE` is defined?

snippets/RAIIHppTemplate.hpp
```diff
@@ -2,14 +2,14 @@
...
-#include <vulkan/${api}.hpp>
 #if !defined( VULKAN_HPP_CXX_MODULE )
+#  include <vulkan/${api}.hpp>
 #  include <memory>   // std::unique_ptr
 #  include <utility>  // std::forward
 #endif
 
 #if !defined( VULKAN_HPP_DISABLE_ENHANCED_MODE )
-namespace VULKAN_HPP_NAMESPACE
+VULKAN_HPP_EXPORT namespace VULKAN_HPP_NAMESPACE
```
asuessenbach, line 12: Again: where's `VULKAN_HPP_EXPORT` defined when `VULKAN_HPP_CXX_MODULE` is defined?

snippets/SharedHppTemplate.hpp
```diff
@@ -3,13 +3,12 @@
...
-#include <vulkan/${api}.hpp>
-
 #if !defined( VULKAN_HPP_CXX_MODULE )
-#include <atomic>  // std::atomic_size_t
+#  include <vulkan/${api}.hpp>
+#  include <atomic>  // std::atomic_size_t
 #endif
 
-namespace VULKAN_HPP_NAMESPACE
+VULKAN_HPP_EXPORT namespace VULKAN_HPP_NAMESPACE
```
asuessenbach, line 11: As above: where's VULKAN_HPP_EXPORT defined when VULKAN_HPP_CXX_MODULE is defined?

snippets/VideoHppTemplate.hpp
```diff
@@ -9,17 +9,16 @@
...
+#  include <vulkan/vulkan.hpp>
 ${includes}
+#endif
 
 #if !defined( VULKAN_HPP_VIDEO_NAMESPACE )
 #  define VULKAN_HPP_VIDEO_NAMESPACE video
 #endif
 
-namespace VULKAN_HPP_NAMESPACE
+VULKAN_HPP_EXPORT namespace VULKAN_HPP_NAMESPACE
```
asuessenbach, line 21: As above: where's `VULKAN_HPP_EXPORT` defined when `VULKAN_HPP_CXX_MODULE` is defined?

vulkan/vulkan.hpp
```diff
@@ -1887,437 +1879,456 @@ namespace VULKAN_HPP_NAMESPACE
     public:
       //=== VK_VERSION_1_0 ===
 
-      VkResult
-        vkCreateInstance( const VkInstanceCreateInfo * pCreateInfo, const VkAllocationCallbacks * pAllocator, VkInstance * pInstance ) const VULKAN_HPP_NOEXCEPT
+      VULKAN_HPP_INLINE VkResult vkCreateInstance( const VkInstanceCreateInfo *  pCreateInfo,
```
asuessenbach, line 1887: Are all those `VULKAN_HPP_INLINE` here required for the modules?
The functions are defined within the class anyway.

vulkan/vulkan_funcs.hpp
asuessenbach, line 1: This file is changed mostly by different formatting?

vulkan/vulkan_handles.hpp
asuessenbach, line 1: This file also has some pure formatting changes.

vulkan/vulkan.hpp
M2-TE: Yes, this is to make sure that the linker does not try to resolve these functions within the module (#2457).
In short, it forces those functions to be TU-external and should have no impact on normal `vulkan.hpp` usage.

asuessenbach: That's weird.

Maybe add some comment somewhere to tell us that since [P1779](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2020/p1779r3.html) function definitions inside a class definition aren't implicitly marked as inline anymore. Instead you have to explicitly mark them as inline.

snippets/HashHppTemplate.hpp
M2-TE: Commenting for all the instances of this; it is gonna be defined since `vulkan.cppm` includes these in a given order.

I do think the include guards around `#include <vulkan/vulkan.hpp>` that is found is some satelite headers can be omitted (macro header will only ever be included once per TU).

vulkan/vulkan.hpp
M2-TE: A comment might be good. But yea, P1779 is why these must be marked as inline since they now sit in the module purview and are explicitly exported by it.

VideoHppGenerator.cpp
sharadhr: Removed!

VulkanHppGenerator.cpp
sharadhr: Removed. The only `Usings` function left is for the PFN commands.

snippets/FormatTraitsHppTemplate.hpp
sharadhr: Indeed. I have collapsed it to just have the `#include`.

snippets/FormatTraits.hpp
sharadhr: If I did that, the compiler would complain about ODR violations (I think that's what caused some local tests to fail). 

What I could do instead, is guard the entire list of _declarations_ with `#if !defined( VULKAN_HPP_CXX_MODULE )`, and then put all the _definitions_ in a similar `VULKAN_HPP_EXPORT` block, also appropriately guarded with `#if defined ( VULKAN_HPP_CXX_MODULE )`. 

With modules, the module interface unit should only contain either the entire definition, or just the declaration.

snippets/HashHppTemplate.hpp
sharadhr: Correct. The module interface file, either `vulkan.cppm` or `vulkan_video.cppm` will define the macro, and by virtue of being `#include`d into the module file, the macro will be defined.

There is separate behaviour for headers that cannot be separately included, versus headers which themselves `#include <vulkan/vulkan.hpp>`. 

@M2-TE I have included behind a macro guard, because I want to suppress the repeated inclusion of `vulkan.hpp` altogether.

vulkan/vulkan.hpp
sharadhr: Added a comment.

vulkan/vulkan_funcs.hpp
sharadhr: Which `clang-format` version are we using? I am running 21.1.3, almost the same as the repository.

snippets/FormatTraits.hpp
asuessenbach: > What I could do instead, is guard the entire list of declarations with `#if !defined( VULKAN_HPP_CXX_MODULE )`, and then put all the _definitions_ in a similar `VULKAN_HPP_EXPORT` block, also appropriately guarded with `#if defined ( VULKAN_HPP_CXX_MODULE )`.

Well, I'm not sure what would be less unexpected.
That is, I think, we can keep it as is.

vulkan/vulkan_funcs.hpp
asuessenbach: Thats indeed surprising.
Maybe you could give it a try with clang-format version 21.1.8? That is used to generate the headers.

Besides that, you still should resolve the conflicts.

M2-TE: I think you forgot to update your forked main branch before merging it into this pr @sharadhr

vulkan/vulkan_handles.hpp
sharadhr: I think these changes are a result of having the `VULKAN_HPP_EXPORT` keyword in front of the namespace. I can't change them even after using the correct formatter; the only way I can revert them is by removing the keyword altogether. So they have to come with each other...

vulkan/vulkan_funcs.hpp
sharadhr: As mentioned below, I think some of these formatting changes come hand-in-hand with `VULKAN_HPP_EXPORT`. They only go back when I remove the macro.

M2-TE:
Could you try re-enabling the extension inspection module test? It should pass now.
Tested locally on linux with both clang++ and g++ and ran just fine.

tests/CMakeLists.txt
```diff
@@ -72,21 +72,19 @@ if( VULKAN_HPP_TESTS_CTEST AND VULKAN_HPP_BUILD_CXX_MODULE )
 	vulkan_hpp__setup_test( NAME EnableBetaExtensions CXX_MODULE )
 	# vulkan_hpp__setup_test( NAME ExtensionInspection CXX_MODULE )
```
sharadhr:
```suggestion
    vulkan_hpp__setup_test( NAME ExtensionInspection CXX_MODULE )
```

sharadhr:
Yay, good to see everything green. Thanks for the help everyone, and especially @M2-TE for help with the dynamic and static dispatchers.

asuessenbach:
You guys rock!

When there's no obejction from your side, I'll merge this on monday.

sharadhr:
No further edits! Thanks, I'll let this stew until Monday then.

asuessenbach:
@sharadhr Would you mind to rebase this one last time?

sharadhr:
Regenerated after a merge.

YaaZ:
Sorry, I'm way too late to the party. I tried this patch with on my project and it seems to work fine. I will definitely consider adopting this approach for VMA-Hpp.

sharadhr:
@asuessenbach I think the CI run got caught up in the GitHub outage; could we try again, please?

PR #2444, opened 2026-01-19
Tests: Added test for vk::to_string

M2-TE:
Should catch issues with `vk::to_string`, also with modules as discussed in #2443.


M2-TE:
Actually, `FormatTraits` uses a single `vk::to_string`, so this should theoretically already be covered.
@sharadhr are you sure the module version of `FormatTraits` passed for you in 2443?

PR #2449, opened 2026-01-19
Tests: Added Debug/Release assert function for tests

### LINKED ISSUE #2389: Replace test asserts with variant that asserts in release mode
Tracking issue for changes discussed in #2292

M2-TE:
This will resolve #2389.

It introduces a new `release_assert` that will also work when compiling in Release mode, which the standard `assert` did not. Pretty much exactly what was discussed in #2292.

### DIFF
--- a/tests/Hash/Hash.cpp
+++ b/tests/Hash/Hash.cpp
@@ -27,6 +27,7 @@
 #endif
 
 
+#include "../test_macros.hpp"
 #ifdef VULKAN_HPP_USE_CXX_MODULE
   #include <vulkan/vulkan_hpp_macros.hpp>
   #include <vulkan/vulkan.h>
@@ -40,7 +41,6 @@ import vulkan;
 #  include "vulkan/vulkan_hash.hpp"
 #endif
 
-#include "../test_macros.hpp"
 #if VULKAN_HPP_DISPATCH_LOADER_DYNAMIC == 1
 namespace vk {
   namespace detail {


tests/test_macros.hpp
```diff
@@ -0,0 +1,12 @@
...
+#include <iostream>
+#define release_assert( x ) \
+  do \
+  { \
+    if ( !( x ) ) \
+    { \
+      std::cout << __FILE__ << ':' << __LINE__ << ": Assertion \'" << #x << "\' failed."<< std::endl; \
+      std::abort(); \
+    } \
+  } while ( false )
```
asuessenbach: Please clang-format this file, would then look like
```
#define release_assert(x)                                                      \
  do {                                                                         \
    if (!(x)) {                                                                \
      std::cout << __FILE__ << ':' << __LINE__ << ": Assertion \'" << #x       \
                << "\' failed." << std::endl;                                  \
      std::abort();                                                            \
    }                                                                          \
  } while (false)
```

M2-TE: Using `clang-format --style='file:.clang-format_21' tests/test_macros.hpp` gives me the following:
```cpp
#define release_assert( x )                                                                            \
  do                                                                                                   \
  {                                                                                                    \
    if ( !( x ) )                                                                                      \
    {                                                                                                  \
      std::cout << __FILE__ << ':' << __LINE__ << ": Assertion \'" << #x << "\' failed." << std::endl; \
      std::abort();                                                                                    \
    }                                                                                                  \
  } while ( false )
```

Did I forget an option here?

asuessenbach: > Did I forget an option here?

No, that's fine. Don't know where that additional line break in my code snippet above comes from.

PR #2455, opened 2026-01-21
CI: Fix header generation in CI

M2-TE:
A bash typo has snuck in, causing headers to not be generated as they should have.


M2-TE:
Perfect, just double checked and it all works as it should!

Issue #2454, opened 2026-01-21
Consider using CMake presets


sharadhr:
This will considerably simplify our CI listings, which have lots of CMake arguments repeated over and over; having it all explicitly listed out in CMake presets that inherit and override one another would make things much more straightforward.

We could have presets for:

- Platforms (Windows, macOS, Ubuntu)
- Toolchains (compilers, build systems (Make, Ninja, MSBuild)
- Environments 
  - generator
  - testing
  - samples
  - consumer install: no generation, no tests, no samples

And many more. This will also simplify our version control. I have something set up in [my own repository](https://github.com/sharadhr/caldera/tree/master/tools/cmake/presets); this could be useful here.

M2-TE:
With #2458 merged, what else should be added?
As already mentioned there, adding presets for all the environments would result in way too many permutations, since Vulkan-Hpp can be used in multiple ways on even just one toolchain setup.

sharadhr:
I think this is done and can be closed. I might create another issue to set up [_toolchain_ files](https://cmake.org/cmake/help/latest/manual/cmake-toolchains.7.html) so we can completely move off defining the toolchain in CI, and just use those toolchain files directly, locally or in CI.

PR #2458, opened 2026-01-29
CMake: Added presets

M2-TE:
This partially addresses #2454 by adding a `CMakePresets.json` to the project root, which contains the most common `VULKAN_HPP_*` flag combinations that people may be interested in.

CI workflow files are now much more concise as you do not get bombarded by all those CMake flags. It also prints out all the applied CMake vars within the run logs, which can be quite nice for tracing down issues.

In the original issue, there was a mention of creating separate presets for different operating systems and toolchain setups. While this could be achieved using something like
```
"condition": {
        "type": "equals",
        "lhs": "${hostSystemName}",
        "rhs": "Windows"
      }
```
for additional presets, it would balloon the preset file due to the large number of possible permutations. Unlike other projects, Vulkan-Hpp has multiple usage scenarios. Perhaps I should add a single preset for each OS/toolchain that does everything (build+run generator, build samples, run tests)?

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -100,7 +100,7 @@ jobs:
     - name: Generate Headers with C++20
       run: |
         for BUILD_TYPE in Debug Release; do
-          [ $BUILD_TYPE = Release ] && PRESET="generator-build" || PRESET="generator-run"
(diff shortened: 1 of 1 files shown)


asuessenbach:
This actually makes the makefiles much clearer.

PR #2460, opened 2026-01-30
Add suffix 'KHR' to eRayTracingSkipBuiltInPrimitives in vulkan_enums.hpp

EasyVulkan:
So I guess this specific enumerator name is not auto-generated?

----------------------------------------

Seems I've created a PR I don't know how to participate in the discussion again.

I'll just leave the follow-up work to you guys.

### DIFF
--- a/vulkan/vulkan_enums.hpp
+++ b/vulkan/vulkan_enums.hpp
@@ -5756,7 +5756,7 @@ namespace VULKAN_HPP_NAMESPACE
     eRetainLinkTimeOptimizationInfoEXT             = VK_PIPELINE_CREATE_2_RETAIN_LINK_TIME_OPTIMIZATION_INFO_BIT_EXT,
     eLibraryKHR                                    = VK_PIPELINE_CREATE_2_LIBRARY_BIT_KHR,
     eRayTracingSkipTrianglesKHR                    = VK_PIPELINE_CREATE_2_RAY_TRACING_SKIP_TRIANGLES_BIT_KHR,
-    eRayTracingSkipBuiltInPrimitives               = VK_PIPELINE_CREATE_2_RAY_TRACING_SKIP_BUILT_IN_PRIMITIVES_BIT_KHR,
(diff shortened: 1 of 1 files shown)


M2-TE:
It does get auto-generated, so it is odd that it doesn't get the KHR suffix.
Registry entry:
```xml
<enum extends="VkPipelineCreateFlagBits2" name="VK_PIPELINE_CREATE_2_RAY_TRACING_SKIP_BUILT_IN_PRIMITIVES_BIT_KHR" alias="VK_PIPELINE_CREATE_2_RAY_TRACING_SKIP_TRIANGLES_BIT_KHR"/>
```

Its alias `VK_PIPELINE_CREATE_2_RAY_TRACING_SKIP_TRIANGLES_BIT_KHR` also has the correct suffix with `eRayTracingSkipTrianglesKHR`..

asuessenbach:
Well, that's an interesting corner case I've never thought about...

Let's try to decode it:
Back in the days, when `VK_PIPELINE_CREATE_2_RAY_TRACING_SKIP_BUILT_IN_PRIMITIVES_BIT_KHR` was introduced, it was an enum value of the enum `VkPipelineCreateFlagBits2KHR`. And as they both had the same tag "KHR", that tag was considered to be redundant and was skipped for that value, resulting in `vk::PipelineCreateFlagBits2KHR::eRayTracingSkipBuiltInPrimitives`.
That is, something like `eRayTracingSkipBuiltInPrimitivesKHR` did not exist.
Then, at some later point in time, `VK_PIPELINE_CREATE_2_RAY_TRACING_SKIP_BUILT_IN_PRIMITIVES_BIT_KHR` became an alias of `VK_PIPELINE_CREATE_2_RAY_TRACING_SKIP_TRIANGLES_BIT_KHR`, and `VkPipelineCreateFlagBits2KHR` became an alias of `VkPipelineCreateFlagBits2`.
The generator now successfully identifies that enum value to not have the tag attached before, and thus doesn't attach it now.
And that now looks like that tag is missing.

Maybe I should extend the logic here a bit to identify such a corner case, and have both, a tagged and an untagged value listed here?

PR #2469, opened 2026-02-07
Move header include above import

mikomikotaishi:
It is a common bug, such as on GCC, where `#include`ing a header after an `import` seems to cause esoteric or otherwise strange compilation errors (see [here](https://github.com/yhirose/cpp-httplib/pull/2291) for an example), even on non-modular translation units. For this reason I recommend moving the `#include`s to be before the `import`.

Furthermore, the `main()` function signature probably need not use trailing return types or `char* const[]`, the traditional `int main(int argc, char* argv[])` is perhaps better for a simple example.

### DIFF
--- a/README.md
+++ b/README.md
@@ -887,22 +887,24 @@ target_link_libraries( VulkanHppModule PUBLIC Vulkan::Headers )
 Finally, supply the macro `VULKAN_HPP_DEFAULT_DISPATCH_LOADER_DYNAMIC_STORAGE` exactly once in your source code, just as in the non-module case. In order to have that macro available, include [`vulkan_hpp_macros.hpp`](vulkan/vulkan_hpp_macros.hpp), a lightweight header providing all Vulkan-Hpp related macros and defines. And as explained above, you need to initialize that dispatcher in two or three steps:
 
 ```cpp
-import vulkan;
-
 #include <vulkan/vulkan_hpp_macros.hpp>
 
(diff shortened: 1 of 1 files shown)


M2-TE:
Indeed, this is an issue on GCC.
The module example in particular is very outdated and is being addressed by #2358.

sharadhr:
As mentioned, this PR will be superseded by #2358; recommend closing.

mikomikotaishi:
This is an independent proposal and the only thing in common with your PR is moving the `#include` above the `import` (which granted, was the main purpose of my PR), but this PR further replaces the trailing return type in the `main()` function (which frankly doesn't have any benefits over the traditional prototype) as well as just not using `auto` (which none of the other code snippets do, anyway).

In any case I don't think it's reasonable to request every PR related to the README be closed because there is currently ongoing work overhauling it. There is no reason why this can't be merged independently, and then the changes be made in the overhauled PR afterwards.

mikomikotaishi:
@asuessenbach Are these changes reasonable to be made to the example? I think it would be appropriate to make this small change in the meantime until #2358 is eventually merged.

asuessenbach:
@mikomikotaishi This PR really is not that big any more. But I prefer to explicitly use the complete namspace hierarchy over a couple of using directives.

That is, I'll close this PR.

mikomikotaishi:
But neither of my two PRs are about using directives (anymore)?

asuessenbach:
With #2358 approaching, you could re-start your PR based on that.

Closing this PR for now.

PR #2470, opened 2026-02-07
Rename shared_handle to MySharedHandle, and use [[maybe_unused] in README example

mikomikotaishi:
~~The README often has verbose names prepended with `vk::`, which adds noise to already long class names. By using `using` statements we can trim some of these long lines and potentially reduce visual load. Notably, the examples already seemed to have some symbols imported into scope, but without a matching `using` statement - this PR adds `using` consistently.~~

For whatever reason there doesn't seem to be a way for me to change the title of the PR, to accurately reflect the changes of it, so the title is currently outdated. This PR intends to fix some inconsistencies in the code examples in the README, as well as use modern features where possible. These include:
- Rename `shared_handle` to `MySharedHandle`, to follow PascalCase for types which the rest of the codebase uses, as well as emphasise that it is a user-created shared handle type
- Use `[[maybe_unused]]` to indicate an unused parameter, rather than commenting out the parameter name

Changes to the module examples are made in #2469, and so are not touched in this PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -351,13 +351,13 @@ Vulkan allows chaining of structures through the `pNext` pointer. Vulkan-Hpp has
 ```c++
 // This will compile successfully.
 vk::StructureChain<vk::MemoryAllocateInfo, vk::ImportMemoryFdInfoKHR> c;
-vk::MemoryAllocateInfo &allocInfo = c.get<vk::MemoryAllocateInfo>();
-vk::ImportMemoryFdInfoKHR &fdInfo = c.get<vk::ImportMemoryFdInfoKHR>();
+vk::MemoryAllocateInfo& allocInfo = c.get<vk::MemoryAllocateInfo>();
+vk::ImportMemoryFdInfoKHR& fdInfo = c.get<vk::ImportMemoryFdInfoKHR>();
 
 // This will fail compilation since it's not valid according to the spec.
 vk::StructureChain<vk::MemoryAllocateInfo, vk::MemoryDedicatedRequirementsKHR> c;
-vk::MemoryAllocateInfo &allocInfo = c.get<vk::MemoryAllocateInfo>();
-vk::ImportMemoryFdInfoKHR &fdInfo = c.get<vk::ImportMemoryFdInfoKHR>();
+vk::MemoryAllocateInfo& allocInfo = c.get<vk::MemoryAllocateInfo>();
+vk::ImportMemoryFdInfoKHR& fdInfo = c.get<vk::ImportMemoryFdInfoKHR>();
 ```
 
 Vulkan-Hpp provides a constructor for these chains similar to the `CreateInfo` objects which accepts a list of all structures part of the chain. The `pNext` field is automatically set to the correct value:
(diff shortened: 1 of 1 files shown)


mikomikotaishi:
~~Notably, some of the snippets like~~
```cpp
using vk::Format;
using vk::Image;
using vk::ImageCreateInfo;
using vk::ImageLayout;
using vk::ImageTiling;
using vk::ImageType;
using vk::ImageUsageFlagBits;
using vk::SampleCountFlagBits;
using vk::SharingMode;

ImageCreateInfo ci({}, ImageType::e2D, Format::eR8G8B8A8Unorm,
                       { width, height, 1 },
                       1, 1, SampleCountFlagBits::e1,
                       ImageTiling::eOptimal, ImageUsageFlagBits::eColorAttachment,
                       SharingMode::eExclusive, 0, nullptr, ImageLayout::eUndefined);
Image image = device.createImage(ci);
```
~~may look worse off with this change, but I applied it anyway for consistency, but feel free to give thoughts. (Of course, this could be done with the dreaded `using namespace vk;` but I believe spelling out where each class comes from is still important - but doing it inside of the code is verbose.)~~

Scratch this, no longer part of the scope of this PR or even an attempted change.

M2-TE:
Doesn't this just add more noise by adding quite a few extra lines of code?

That aside, it is good practice to explicitly use namespaces when using symbols that stem from them. This goes for both `std` and `vk`.

Even though these are just examples, I do not think we should encourage the use of `using vk:XXX` through them.

mikomikotaishi:
> That aside, it is good practice to explicitly use namespaces when using symbols that stem from them. This goes for both `std` and `vk`.

This notion is rooted in the assumption that `using` statements leak across translation units, especially headers.

~~If the concern on those lines is noise, then we should either use using namespace or just assume the symbol is already imported into scope. Either way in these README examples the context is obvious that these classes belong to `vk::`, which is why I think we could just drop `vk::` qualifiers as they don't add anything that isn't already obvious.~~

sharadhr:
>  Either way in these README examples the context is obvious that these classes belong to vk::

We can't say that for certain, as the C symbols may still be available. Additionally, at least within the readme, I don't want to rely on what may seem 'obvious'; what is obvious to someone may be completely novel to someone else. 

We have readers of quite a variety of programming skill working through the readme, and I think it is important to have something that everyone can approach. I want to err on the side of copy-pastability and immediate use and therefore fully qualifying every symbol used, whereas notions of 'verbose' and 'visual load' are not exactly well-defined. This is similar to `using namespace std`, which is strongly discouraged across the board. 

If advanced users desire to re-qualify these names within their code, sure, let's not stop them, but for now I do not agree with the content in this PR. Furthermore, this will clash with the very immense (and in my view, very necessary) changes in #2358, which I recommend prioritising, and if you have any comments on _that_, we can address it there, I think.

mikomikotaishi:
> We can't say that for certain, as the C symbols may still be available.

The C symbols are prefixed with `Vk`.

> Additionally, at least within the readme, I don't want to rely on what may seem 'obvious'; what is obvious to someone may be completely novel to someone else.

This isn't a particularly convincing argument. It could be extrapolated into asking questions like, "how are we sure our users know what header to include to get these symbols"?

> This is similar to `using namespace std`, which is strongly discouraged across the board.

This is only discouraged in headers, not source files.

There was something this PR hoped to address, specifically the example where a custom shared device is created - I find it strange that `class shared_device<VkDevice>` suddenly uses snake_case (as is `friend base;`) when the rest of the code base does not. I think a name like `MySharedDevice<VkDevice>` is far more appropriate, as not only does it match the rest of the code, but the prefix "My" emphasises that it is a user-created class, rather than awkwardly using `shared_device` just to get around there being a `vk::SharedDevice` class already (which, in the example, was not even imported into the global namespace).

I further wanted to use `[[maybe_unused]]` rather than a commented-out parameter to indicate an unused parameter as I think the intent is far clearer in doing so.

~~Also, `const T&` and `T const&` are the same. `const T&` is more prevalent and reads more naturally. This PR aims to also change those as well.~~ Scratch this, it seems the code base uses `T const&`.

Edit: after further thought, I don't think the changes of adding `using` statements in (most of) the snippets is necessary, but I do think it is a desirable change in snippets demonstrating a full program where there is freedom for additional lines (such as the example module program).

mikomikotaishi:
@asuessenbach Do you have any thoughts on this PR? Once again I want to remind you that the title is outdated and the current scope of the PR is just a few changes to names and adding a `[[maybe_unused]]` attribute (and possibly some other minor refactors).

asuessenbach:
With #2358 approaching, you could re-start your PR based on that.

Closing this PR for now.

PR #2476, opened 2026-02-11
Add missing VULKAN_HPP_DISPATCH_LOADER_DYNAMIC define in StructureChain test.

asuessenbach:


M2-TE:
The `VULKAN_HPP_DISPATCH_LOADER_DYNAMIC` is supposed to already be defined in `tests/StructureChain/CMakeLists.txt`. Is it missing when compiling the test without `ctest`?

asuessenbach:
> Is it missing when compiling the test without ctest?

Yep.

M2-TE:
I think you can safely merge this; I'll check out why it is missing!

M2-TE:
I reckon you ran the tests with `VULKAN_HPP_DISPATCH_LOADER_DYNAMIC` left at the default value (disabled).
Without CTest, there is no separate `Vulkan::Hpp` target that has the dynamic dispatcher enabled, which is why some tests are hidden behind `if( VULKAN_HPP_DISPATCH_LOADER_DYNAMIC )`. This one was moved out of it apparently.

A better solution might be to remove that gate altogether and _always_ run the tests requiring dynamic dispatchers and just add the `VULKAN_HPP_DISPATCH_LOADER_DYNAMIC` in the test source, just like you have done here.

I'll create a PR for it once you have merged the changes here!

PR #2480, opened 2026-02-12
CMake: Generate .gitignore template when top-level

M2-TE:
This will ignore common build artefacts that I know of. I've always kept this one (alongside an entry for `.gitignore` itself) when working with this repository. Having one by default makes it easier to work with this repo out of the gate.

Are there any entries I missed? I have not used any MacOS IDEs for example.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -85,6 +85,27 @@ function( vulkan_hpp__setup_warning_level )
 	endif()
 endfunction()
 
+# Create a .gitignore file that ignores itself to avoid tracking
+if( NOT EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/.gitignore" AND PROJECT_IS_TOP_LEVEL )
+	string(CONCAT VULKAN_HPP_GITIGNORE_ENTRIES
+		"builds/\n"
+		"build/\n"
+		"out/\n"
+		"\n"
+		".cache/\n"
+		".vscode/\n"
+		".vs/\n"
+		"\n"
+		".clang-format\n"
+		".gitignore\n"
+		"\n"
+		"*.pdb\n"
+		"*.proj.DotSettings\n"
+		"\n"
+		"Folder.DotSettings.user\n" )
+	file( WRITE ${CMAKE_CURRENT_SOURCE_DIR}/.gitignore ${VULKAN_HPP_GITIGNORE_ENTRIES} )
+endif()
+
 # Build Vulkan-Hpp and Video-Hpp generators
 if( VULKAN_HPP_GENERATOR_BUILD )
 	set_property( GLOBAL PROPERTY USE_FOLDERS ON )


asuessenbach:
I've thought about adding `.gitignore` to the project long time ago.
But you can, for example, name the "build" directory to your liking (mine is called "builds"). That is, if you want to have it named differently, you would have to modify a tracked file. And then you have to explicitly exclude it from your PRs, as your local changes should not be part of the repository.
Maybe we could add something like a gitignore template for everyone to start with. But I think, that file is pretty easy to generate and maintain, so I don't see a compelling reason to track it.

M2-TE:
While it is true there are a few build folder names out there, the ones listed here are _very_ common. Using `builds` should also be common, so it's worth adding.

> Maybe we could add something like a gitignore template for everyone to start with

Would this not be a good starting template too? If they need to ignore something in addition to what we listed here, they can easily add it.

> And then you have to explicitly exclude it from your PRs, as your local changes should not be part of the repository.

I could add `.gitignore` itself to the ignored files at the end of this PR, so people can add or remove entries without it being tracked by git.

sharadhr:
I use `out/` :)

asuessenbach:
Ok, when .gitignore ignores itself, you don't need to exclude it from your PRs. Check.
But I don't know what happens if you locally changed your .gitignore, and pull a newer version of the repository?

M2-TE:
When you change branches or pull a newer version from the repo, it will work just fine and should update the local file.
To update the .gitignore in the future, you gotta make a PR with 3 commits:

1. Remove the `.gitignore` entry
2. Add/Remove/Change the entries you want
3. Add the `.gitignore` entry

I'll experiment a bit with this in my own repo, to see if it works as expected.

M2-TE:
Unfortunately, `.gitignore` can only ignore untracked files, thus it cannot ignore itself once it is already tracked. So if you make changes to it because you have special output folder names, it will not be excluded from PRs.

That being said, I reckon there aren't that many non-standard output folder names out there. Are those <1% users worth not having a standardized .gitgnore file for? If there are common entries we missed, they can simply be updated later.

M2-TE:
Another idea:

To pick up on the "template" idea, we could generate a `.gitignore` in the `CMakeLists.txt`, if it detects that it is top-level. The generated `.gitignore` would contain all the common folders we have listed, as well as an entry to ignore itself (which will never be tracked).

This way, the generated `.gitignore` will _locally_ persist across pulls and branch changes without issues, even if you change entries in it _locally_; wdyt?

asuessenbach:
Yep, nice.
However, if we determine at some point in the future that the generated .gitignore file needs an additional entry, you will not receive it because you already have one.

M2-TE:
We could write some version number into the first line of the .gitignore as a comment. This can be parsed in CMake, forcefully updating the local .gitignore if we have made any changes.

Perhaps a bit overkill? It is very easy to remove the local .gitignore after all.

asuessenbach:
> Perhaps a bit overkill?

Exactly.
Are you ok with closing this PR?

M2-TE:
Yup!

PR #2481, opened 2026-02-12
Generator: Moved related files into separate `generator/` folder

M2-TE:
This groups all files related to the generator into a dedicated `generator` folder. It greatly helps visual clarity in the project root, while also improving clarity about separation from the other components in this repo.

Should not be a breaking change, as relevant paths have been adjusted. The generator itself uses the `BASE_PATH` variable, which is set in CMake and does not need any adjustments.


M2-TE:
Is this something you would consider merging @asuessenbach? This PR is rather sensitive to any and all changes to the generator, so upkeep isn't the easiest. Just let me know whether you would even want to see this change!

asuessenbach:
> Just let me know whether you would even want to see this change!

Well, I would not say, this change is very important, but you're right, is would clean up the repository a bit, make it less cluttered.
That is: if you're willing to resolve the conflicts, and everything continues to work as expected, I'll merge it.

M2-TE:
> I would not say, this change is very important

That is very true! It's a subjective change that others hopefully appreciate, rather than need.

I've merged it to be up-to-date with main, meaning all merge conflicts should be gone now.

asuessenbach:
Thanks again for your ongoing contribution!

PR #2485, opened 2026-02-14
Accept const char* instead of std::string for setDebugUtilsObjectNameEXT().

stripe2933:
Is there any reason for using `std::string` over `const char*` for `setDebugUtilsObjectNameEXT()` function parameter?

https://github.com/KhronosGroup/Vulkan-Hpp/blob/cd3d0892e501a091c2cbda31b75f9905096b2f5f/samples/DebugUtilsObjectName/DebugUtilsObjectName.cpp#L58

Seems the example also calling `std::string::string(const char*)`, which seems redundant.


asuessenbach:
It's because `std::string const &` is the preferred data type for string parameters in the "enhanced" function versions.

PR #2491, opened 2026-02-17
Update docs to be consistent with prevalent style

mikomikotaishi:
This PR addresses some of the inconsistencies seen in the docs:
- The `MySharedHandle` class names its base class as `base`, though the rest of the library uses PascalCase, so this should be updated to `Base`.
- `MySharedHandle::getParent` should be clearer on what it returns, so I updated the return type from `const auto&` to `const vk::SharedInstance&`. I also marked it `[[nodiscard]]` because it is a getter function and its value shouldn't be discarded if called.
- Instead of commenting out the parameter name on `MySharedHandle::internalDestroy`, use the `[[maybe_unused]]` attribute to make it clear the parameter is not considered.
- The C++ modules example used `auto` declarations (`auto appInfo = ApplicationInfo(/* ... */);`), but the rest of the library doesn't. This could instead more simply (and more briefly) be written with direct-initialisation (`ApplicationInfo appInfo(/* ... */);`).
- The traditional `main()` signature is most often depicted as `int main()`, rather than with trailing return type (`auto main() -> int`), and there's no reason to make this arbitrary difference. Also, `argv` is most often depicted as `char* argv[]`, not `char* const argv[]`, so we might as well just stick with `char* argv[]` (without the `const` specifier).

### DIFF
--- a/docs/Handles.md
+++ b/docs/Handles.md
@@ -83,21 +83,22 @@ Add a `friend` declaration for the base class.
 // Example of a custom shared device, that accepts an instance as a parent
 class MySharedHandle<VkDevice> : public vk::SharedHandleBase<VkDevice, vk::SharedInstance, MySharedHandle<VkDevice>>
 {
-  using base = vk::SharedHandleBase<VkDevice, vk::SharedInstance, MySharedHandle<VkDevice>>;
-  friend base;
+  using Base = vk::SharedHandleBase<VkDevice, vk::SharedInstance, MySharedHandle<VkDevice>>;
+  friend Base;
 
 public:
   MySharedHandle() = default;
-  explicit MySharedHandle(VkDevice handle, vk::SharedInstance parent) noexcept
-    : base(handle, std::move(parent)) {}
+  explicit MySharedHandle(VkDevice handle, vk::SharedInstance parent) noexcept:
+    Base(handle, std::move(parent)) {}
 
-  const auto& getParent() const noexcept
+  [[nodiscard]]
+  const vk::SharedInstance& getParent() const noexcept
   {
     return getHeader();
   }
 
 protected:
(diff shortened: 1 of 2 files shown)


sharadhr:
Thanks for trying again. I meant to fold your corrections from earlier into the big commit, but clearly I missed many. Any additional wording suggestions (in fact, I think it's a good idea to use `[[maybe_unused]]` for all the `/* varname */` everywhere else) would be very welcome.

One note, though...

> is most often depicted as `char* argv[]`, not `char* const argv[]`

I think there is a safety argument here; in fact, I would `const`-qualify _more_. The safest representation of `argv` is `char const* const argv[]`.

mikomikotaishi:
This is a simple example, so I don't really think the need to `const`-qualify as much as possible is necessary. For most people, the most recognised form of `main()` is `int main(int argc, char* argv[])` (page 13 of the [C standard](https://www.open-std.org/jtc1/sc22/wg14/www/docs/n3550.pdf) doesn't `const`-qualify `argv` at all; across the rest of the document it remains un-`const`-qualified). I'm not against `const`-qualifying but seeing as it is a simple example, I think there isn't much of a reason to make it more contrived than necessary. (We might not even need to use `argc` and `argv` at all in the example as those parameters aren't even consumed in the example, but I don't mind either way.)

Also, while I didn't return to re-adding `using` statements, perhaps it might be worthwhile considering `using` statements for the modules example, as it is a full example (in a `main()` function) rather than a code snippet - importing symbols into scope is probably worthwhile in making the example easier to read.

mikomikotaishi:
> Any additional wording suggestions (in fact, I think it's a good idea to use [[maybe_unused]] for all the /* varname */ everywhere else) would be very welcome.

There is now a PR for this, #2492.

asuessenbach:
Thanks for your contribution!

PR #2492, opened 2026-02-18
Remove unused parameters from `main()` and apply `[[maybe_unused]]` elsewhere

mikomikotaishi:
Across tests, a lot of `main()` functions had commented-out parameters (like `int main(int /*argc*/, char** /*argv*/)`). Since these parameters aren't going to be used anyway, we should just make these simply `int main()`. I have also replaced function signatures using commented-out parameters to use the `[[maybe_unused]]` attribute, which is clearer and more consistent with modern C++.

### DIFF
--- a/tests/test_macros.hpp
+++ b/tests/test_macros.hpp
@@ -10,3 +10,4 @@
       std::abort();                                                                                    \
     }                                                                                                  \
   } while ( false )
+
--- a/RAII_Samples/01_InitInstance/01_InitInstance.cpp
+++ b/RAII_Samples/01_InitInstance/01_InitInstance.cpp
@@ -22,7 +22,7 @@
 static std::string AppName    = "01_InitInstanceRAII";
 static std::string EngineName = "Vulkan.hpp";
 
-int main( int /*argc*/, char ** /*argv*/ )
+int main()
 {
   /* VULKAN_HPP_KEY_START */
 
--- a/RAII_Samples/02_EnumerateDevices/02_EnumerateDevices.cpp
+++ b/RAII_Samples/02_EnumerateDevices/02_EnumerateDevices.cpp
@@ -32,7 +32,7 @@
 static std::string AppName    = "02_EnumerateDevicesRAII";
 static std::string EngineName = "Vulkan.hpp";
 
-int main( int /*argc*/, char ** /*argv*/ )
+int main()
 {
   try
   {
--- a/RAII_Samples/03_InitDevice/03_InitDevice.cpp
+++ b/RAII_Samples/03_InitDevice/03_InitDevice.cpp
@@ -22,7 +22,7 @@
 static char const * AppName    = "03_InitDeviceRAII";
 static char const * EngineName = "Vulkan.hpp";
 
-int main( int /*argc*/, char ** /*argv*/ )
+int main()
 {
   try
   {
--- a/RAII_Samples/04_InitCommandBuffer/04_InitCommandBuffer.cpp
+++ b/RAII_Samples/04_InitCommandBuffer/04_InitCommandBuffer.cpp
@@ -22,7 +22,7 @@
 static char const * AppName    = "04_InitCommandBufferRAII";
 static char const * EngineName = "Vulkan.hpp";
 
-int main( int /*argc*/, char ** /*argv*/ )
+int main()
 {
   try
   {
--- a/RAII_Samples/05_InitSwapchain/05_InitSwapchain.cpp
+++ b/RAII_Samples/05_InitSwapchain/05_InitSwapchain.cpp
@@ -22,7 +22,7 @@
 static char const * AppName    = "05_InitSwapchainRAII";
 static char const * EngineName = "Vulkan.hpp";
 
-int main( int /*argc*/, char ** /*argv*/ )
+int main()
 {
   try
   {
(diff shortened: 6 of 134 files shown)


asuessenbach:
Thanks for this great PR. Touches a whole lot of files.

I also would prefer to use `[[maybe_unused]]`. But it's a C++17 attribute, and all the samples need to compile with C++11 and C++14 as well. Therefore, it can't be used in general here in this repository.

mikomikotaishi:
How about instead of commenting out the names we just use `(void)`? Also, the examples will still work on C++11, as the compiler will just ignore any unrecognised attributes (thus I think using both the attribute and `(void)`-cast is the best solution for clarity and suppressing the unused parameter warning).

asuessenbach:
> the compiler will just ignore any unrecognised attributes

I'm not sure about this. Worth a try, though.

> thus I think using both the attribute and (void)-cast is the best solution for clarity and suppressing the unused parameter warning

Could you give some example how you're envisioning it?

mikomikotaishi:
> Could you give some example how you're envisioning it?

For example:
```cpp
VKAPI_ATTR vk::Bool32 VKAPI_CALL debugMessageFunc( vk::DebugUtilsMessageSeverityFlagBitsEXT       messageSeverity,
                                                   vk::DebugUtilsMessageTypeFlagsEXT              messageTypes,
                                                   vk::DebugUtilsMessengerCallbackDataEXT const * pCallbackData,
                                                   [[maybe_unused]] void * pUserData )
{
    (void)pUserData; // discard
    std::ostringstream message;
    // ...
}
```

asuessenbach:
... or maybe just keep the arguments unused, and silence the compiler by some pragmas?
Like
```
#if defined( _MSC_VER )
#  pragma warning( disable : 4100 )  // disable warning 4100: unreferenced parameter
#elif defined( __clang__ )
#  pragma clang diagnostic ignored "-Wno-unused-parameter"
#elif defined( __GNUC__ )
#  pragma gcc diagnostic ignored "-Wno-unused-parameter"
#else
// unknown compiler... just ignore the warnings for yourselves ;)
#endif
```

It's not the recommended way, but for those little samples and tests it might be sufficient.

mikomikotaishi:
I think `(void)` casts were the most standard way of suppressing unused parameters until the addition of `[[maybe_unused]]`, but if there's a problem with it please let me know.

Edit: see https://c-faq.com/style/voidcasts.html

mikomikotaishi:
I went ahead with just using `(void)`-casts since this is how it was done since C. Do let me know if you want anything else changed.

mikomikotaishi:
Perhaps I was mistaken. I think the behaviour of ignoring unrecognised attributes was only specified as of C++17, so unfortunately using attributes here is not portable at all.

I guess we can just stick with void casts for tests.

asuessenbach:
Well, I guess not all compilers agree to "just ignore any unrecognised attributes".

Maybe you could add something like
```
#if 17 <= VULKAN_HPP_CPP_VERSION
#  define VULKAN_HPP_MAYBE_UNUSED [[maybe_unused]]
#  define VULKAN_HPP_UNUSED( var )
#else
#  define VULKAN_HPP_MAYBE_UNUSED
#  define VULKAN_HPP_UNUSED( var ) ( (void)( var ) )
#endif
```
to `snippets/MacrosHppTemplate.hpp`?

mikomikotaishi:
It looks like only the tests get compiled for C++11-C++23, while samples are just tested on C++23. I'll just apply it to the tests

asuessenbach:
> It looks like only the tests get compiled for C++11-C++23, while samples are just tested on C++23.

But that might change anytime without further notice. It's just a matter of limited resources.

That is, I would prefer to not intentionally break the samples on C++11 or C++14.

mikomikotaishi:
OK, I replaced it with macros instead as requested.

tests/test_macros.hpp
```diff
@@ -10,3 +10,12 @@
...
     }                                                                                                  \
   } while ( false )
+
+#if 17 <= VULKAN_HPP_CPP_VERSION
+#  define VULKAN_HPP_MAYBE_UNUSED [[maybe_unused]]
+#  define VULKAN_HPP_UNUSED( var )
+#else
+#  define VULKAN_HPP_MAYBE_UNUSED
+#  define VULKAN_HPP_UNUSED( var ) ( (void)( var ) )
+#endif
```
asuessenbach: Why do you add these defines here as well?

mikomikotaishi: Because a few tests use them too and the tests were how we found out that [[maybe_unused]] doesn't work on C++11 (or can't be ignored).

asuessenbach: And those tests don't include `vulkan.hpp` (or `vulkan_hpp_macros.hpp`)?

mikomikotaishi: Ah, you might be right
It's been fixed

mikomikotaishi: @asuessenbach Sorry. it looks like if `VULKAN_HPP_USE_CXX_MODULE` is defined (it imports the `vulkan` module) then the `VULKAN_HPP_MAYBE_UNUSED` macros don't get passed in because modules do not export macros. That's why it was included in `tests/test_macros.hpp`. I restored it put put it behind a `#ifndef VULKAN_HPP_USE_CXX_MODULE`.

asuessenbach: No, please don't do it like this.
Please include vulkan_hpp_macros.hpp in those tests where it's needed, like it's done in ArrayWrapper.cpp, for example.

mikomikotaishi: Okay, fixed

asuessenbach:
Ok, thanks for your contribution.

PR #2499, opened 2026-02-21
Merge ResultValue with specializations

YaaZ:
Implementation of `ResultValue` diverged with its specializations for unique handles: base class has `std::expected`-like functions and specializations have `asTuple()` function. The specializations are mostly boilerplate, as they are identical, with exception of the specialized type, meaning that they can easily be merged with the base class.

So I brought rvalue variant of `asTuple()` into the base class and dropped the specializations.
This means that I have also dropped lvalue `asTuple()` variant, which was deprecated for a few years already: #1605
This should hopefully simplify the maintenance without needing to keep the base class and specializations in sync.

And it also helps with other use cases, like `ResultValue<std::pair<...>>` variant (https://github.com/YaaZ/VulkanMemoryAllocator-Hpp/issues/78), which would otherwise require one more specialization.

Note that I did not run the generator - clang-format gave me a crazy diff for some reason, so I patched vulkan.hpp by hand.

### DIFF
--- a/snippets/ResultValue.hpp
+++ b/snippets/ResultValue.hpp
@@ -27,6 +27,11 @@
       return std::tuple<Result &, T &>( result, value );
     }
 
+    std::tuple<Result, T> asTuple() &&
+    {
+      return std::make_tuple( result, std::move( value ) );
+    }
+
 	// std::expected-look alike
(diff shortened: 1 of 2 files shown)


asuessenbach:
That's a nice little simplification.
I like it, when there's more code removed than added ;)

Thanks for your contribution.

PR #2529, opened 2026-03-20
Fix compiler `import std` detection in CMake

sharadhr:
Resolves #2525.


theHamsta (approved):
Thank you for this prompt response and reporting to the CMake devs!

Works nicely for me with g++-16 and clang-22 (both expose `CMAKE_CXX_COMPILER_IMPORT_STD` for 23 and 26)

PR #2563, opened 2026-05-07
Fix typo: MacOS 26 -> MacOS 16

asuessenbach:


M2-TE:
@asuessenbach MacOS 26 was actually correct (I had to double check a few times in the past too):

https://github.com/KhronosGroup/Vulkan-Hpp/blob/1520807db76d79438c33effe202c9f09411c1a5d/.github/workflows/build.yml#L18-L20

Here's the MacOS image selection: https://github.com/actions/runner-images/tree/main/images/macos

asuessenbach:
Ahm... that's... unexpected.

Ok, I'll try to revert this. Thanks for that hint!

PR #2565, opened 2026-05-07
CI: Updated windows-2025 runner

M2-TE:
As per the notice from the `windows-2025` runner

```
NOTICE: windows-2025 requests are being redirected to windows-2025-vs2026 by May 12, 2026
```

I've replaced the `windows-2025` runner with `windows-2025-vs2026` to perform this transition explicitly, so that we can fix potential issues before it is forced on us.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -15,32 +15,32 @@ jobs:
     strategy:
       matrix:
         env: [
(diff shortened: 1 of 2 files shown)


asuessenbach:
> As per the notice from the windows-2025 runner

Where do you find this notice?

M2-TE:
I found it in the actions tab in a build run that uses the windows runner. Here's one example:

https://github.com/KhronosGroup/Vulkan-Hpp/actions/runs/25504112222

asuessenbach:
> I found it in the actions tab in a build run that uses the windows runner.

Souds reasonable then.

Issue #2569, opened 2026-05-11
Which copyright and license to /generator/snippets/* fall under?


KhronosWebservices:
@M2-TE and/or @asuessenbach Do you know which Copyright and which license the files under https://github.com/KhronosGroup/Vulkan-Hpp/tree/main/generator/snippets fall under? Given the rest of the /generator files are under NVIDIA Corporation and Apache-2.0, is it safe to assume the snippets/* are under: 

```
SPDX-FileCopyrightText: NVIDIA Corporation
SPDX-License-Identifier: Apache-2.0
```

asuessenbach:
@KhronosWebservices Yes, the files under snippets are under the same license as the rest of the /generator files.

M2-TE:
@asuessenbach should we add licence headers to the snippets? They could easily be skipped over when reading them from within the generators.

asuessenbach:
> should we add licence headers to the snippets?

@M2-TE I think, that would be the right move. Would you like to craft such a PR?

M2-TE:
The license compliance lint introduced in #2570 seems to not really check for any compliance at the moment. After skimming the documentation for [reuse-tool](https://codeberg.org/fsfe/reuse-tool), the [REUSE.toml](https://github.com/KhronosGroup/Vulkan-Hpp/blob/main/REUSE.toml) annotations that we have now assign licenses during linting, specifically using the `**` wildcard. This means that even files with no licence/copyright headers will never get flagged.

I believe that there are a few things left to do for actual compliance:
1. Only add files/wildcards for paths in `REUSE.toml` that _do not_ and _should not_ have a license header (`[[annotations]]` are meant to add exceptions from what I understand).
2. Update all license headers for the files we currently have. Either that, or manually add all our licensed files (without wildcards) to `REUSE.toml`. Some of the licenses we have right now cannot be properly parsed by the linter and might need to be adjusted (e.g. [RAII_Samples/01_InitInstance/01_InitInstance.cpp](https://github.com/KhronosGroup/Vulkan-Hpp/blob/main/RAII_Samples/01_InitInstance/01_InitInstance.cpp) is reported to not have licensing information, but the copyright was successfully parsed).
3.  Validate that adding files without licenses actually triggers a compliance linter error.

I have no practical experience with the reuse tool yet, so perhaps I have misunderstood something here. So, before I work on making any changes, please comment on whether or not this seems reasonable @KhronosWebservices @asuessenbach.

[...]

PR #2571, opened 2026-05-12
Generator: Added license headers to generators and snippets

M2-TE:
As per #2569, all `generator/snippets/*` now have the following license header:

```cpp
// SPDX-FileCopyrightText: 2026 NVIDIA CORPORATION
// SPDX-License-Identifier: Apache-2.0
```

This is stripped out when reading the snippet during header generation.
`REUSE.toml` was updated to the minimum of what is now needed.

Auto formatting removed the spurious indents for some snippets. Makes them a bit more readable and does not affect output, so I reckon I should leave it as is.

### DIFF
--- a/generator/XMLHelper.hpp
+++ b/generator/XMLHelper.hpp
@@ -5,7 +5,6 @@
 
 #include <algorithm>
 #include <assert.h>
-#include <format>
 #include <fstream>
 #include <iostream>
 #include <map>
@@ -429,7 +428,7 @@ inline std::string readSnippet( std::string const & snippetFile )
     }
     if ( !compliant )
     {
-      throw std::runtime_error( std::format( "The snippet {} contains a broken license header", snippetFile ) );
+      throw std::runtime_error( "The snippet " + snippetFile + " contains a broken license header" );
     }
   }
   // return the remainder of the snippet
--- a/generator/XMLHelper.hpp
+++ b/generator/XMLHelper.hpp
@@ -5,6 +5,7 @@
 
 #include <algorithm>
 #include <assert.h>
+#include <format>
 #include <fstream>
 #include <iostream>
 #include <map>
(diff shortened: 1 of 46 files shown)


M2-TE:
Updated the license headers for the generators as well.

Only the `.clang-format_*` files remain without a license, so those are excluded via `REUSE.toml`

generator/XMLHelper.hpp
```diff
@@ -422,6 +411,13 @@ inline std::string readSnippet( std::string const & snippetFile )
 {
   std::ifstream ifs( std::string( BASE_PATH ) + "/generator/snippets/" + snippetFile );
   assert( !ifs.fail() );
+  // skip the snippet license header (3 loc)
+  std::string line;
+  for ( uint32_t i = 0; i < 3; i++ )
+  {
+    std::getline(ifs, line);
+  }
```
asuessenbach, line 433: Maybe we should check that the first three lines look like we expect them to look like.
Like
```
  std::string line;
  std::getline( ifs, line );
  assert( line.starts_with( "// SPDX-FileCopyrightText:" ) );
  std::getline( ifs, line );
  assert( line.starts_with( "// SPDX-License-Identifier:" ) );
  std::getline( ifs, line );
  assert( line.empty() );
```

M2-TE: That's a good idea, blindly skipping them would lead to hard-to-track-down issues when new snippets are added without license..

M2-TE:
I've decided against using assert, as the generators are often built/run in Release mode due to the substantially higher header generation times in Debug. Instead, it uses a good old exception:

```cpp
// skip the snippet license header and validate its contents
std::string line;
for ( uint32_t i = 0; i < 3; i++ )
{
  bool compliant = true;
  std::getline( ifs, line );
  switch ( i )
  {
    // REUSE-IgnoreStart
    case 0: compliant = line.starts_with( "// SPDX-FileCopyrightText:" ); break;
    case 1: compliant = line.starts_with( "// SPDX-License-Identifier:" ); break;
    case 2: compliant = line.empty(); break;
    default: compliant = false; // unreachable
    // REUSE-IgnoreEnd
  }
  if ( !compliant )
  {
    throw std::runtime_error( std::format( "The snippet {} contains a broken license header", snippetFile ) );
  }
}
```

Validated that it catches broken snippets:

```
caught exception: The snippet ArrayProxy.hpp contains a broken license header
```

asuessenbach:
> I've decided against using assert, ... Instead, it uses a good old exception

Sounds good.

asuessenbach:

PR #2575, opened 2026-05-13
chore: Update license headers for samples, RAII_Samples, tests, workflows and .gitmodule

### LINKED ISSUE #2569: Which copyright and license to /generator/snippets/* fall under?

M2-TE:
This PR will add the last bit of work for and resolve #2569.
All license headers for `samples/**`, `RAII_Samples/**`, `tests/**`, `.github/workflows/**` and the `.gitmodules` have been updated. The years in the previous licenses were of course retained everywhere and updated to span to 2026.

There are only very few leftovers that are being ignored by the license checker now:
```toml
    "CMakeLists.txt", # has a license that is different to the others (no MIT or Apache 2.0)
    "CMakePresets.json", # json does not support comments
    "VulkanHpp.natvis", # might work to insert comments into this, but I have no VS to test this
    "generator/.clang-format_*", # also does not support comments I believe
```

These are smaller cases though and can be handled on a case-by-case basis, if at all (it's not like there will be a second `CMakeLists.txt` at the repo root). This PR is just for the bulk of the files we have that are actually relevant for licensing.


asuessenbach:
Great work, indeed!

PR #2580, opened 2026-05-20
docs: clarify Vulkan-Hpp module target setup

puneetdixit200:
## Summary
- clarify that the CMake example builds a project-local `Vulkan-HppModule`
- keep the `FindVulkan` / `FetchContent` setup so users do not need a `VulkanHeaders` config package
- note that compiling the named module in the consuming project keeps it matched to that toolchain

Closes #2508

## Testing
- `git diff --check`

### DIFF
--- a/docs/Usage.md
+++ b/docs/Usage.md
@@ -891,6 +891,7 @@ add_executable( YourProject main.cpp )
 target_link_libraries( YourProject PRIVATE Vulkan-HppModule )
 ```
 
+The example creates `Vulkan-HppModule` in the consuming project so CMake compiles the named module with the same compiler and options as the rest of the build.
 Configuring the named module is straightforward; add any required Vulkan-Hpp feature macros listed in [Configuration](./Configuration.md) (or any C macros) to `target_compile_definitions`.
 Ensure the macros are `PUBLIC` so they can be propagated to user code.
 For instance:


theHamsta:
Looks reasonable to me to always use the Vulkan-Hpp provided target.  @M2-TE what do you think?

docs/Usage.md
```diff
@@ -850,18 +851,11 @@ project( vulkan_hpp_modules_example LANGUAGES CXX )
...
-if ( Vulkan_FOUND )
-    # set up Vulkan C++ module as a static library
-    add_library( Vulkan-HppModule STATIC )
-    add_library( Vulkan::HppModule ALIAS Vulkan-HppModule )
-    target_sources( Vulkan-HppModule PUBLIC
-        FILE_SET CXX_MODULES
-        BASE_DIRS ${Vulkan_INCLUDE_DIR}
-        FILES
-            ${Vulkan_INCLUDE_DIR}/vulkan/vulkan.cppm
-            ${Vulkan_INCLUDE_DIR}/vulkan/vulkan_video.cppm)
```
theHamsta, line 863: this recompiles the module from the installed headers. This ensures the compiled module is matching the current toolchain.

Directly using `Vulkan::HppModule` might have been compiled for a different compiler. `Vulkan::HppModule` is not recompiling, but using the installed compiled module ?  @M2-TE

M2-TE: You are half correct; the problem is that the `find_package` version for finding `VulkanHeaders` does not play nice with modules (more of a CMake problem).

We have explicitly forbidden the installation of module build artefacts over in Vulkan-Headers, so compiling locally is a necessity.

docs/Usage.md
```diff
@@ -850,18 +851,11 @@ project( vulkan_hpp_modules_example LANGUAGES CXX )
...
-else()
+# Enable the Vulkan-Hpp module target provided by Vulkan-Headers.
+set( VULKAN_HEADERS_ENABLE_MODULE ON CACHE BOOL "Build Vulkan-Hpp module target" )
```
sharadhr: I'm not sure if we want to set this variable unilaterally; the [check](https://github.com/KhronosGroup/Vulkan-Headers/blob/015e25c3c91b70eb1a754d36fb14c4ba6ad9b0b9/CMakeLists.txt#L47) is there for a reason. If CMake module support is enabled then the option will turn on automatically; it is not really meant to be controlled externally.

docs/Usage.md
```diff
@@ -850,18 +851,11 @@ project( vulkan_hpp_modules_example LANGUAGES CXX )
...
+
+find_package( VulkanHeaders 1.4.344 CONFIG QUIET )
```
sharadhr: We also cannot expect that this package will available. The Vulkan SDK (which is the default path for acquiring these headers) still [does not provide](https://vulkan.lunarg.com/issue/view/6598781b5df1125b58afbb3c) this CMake config file, so we must construct the module manually.

docs/Usage.md
```diff
@@ -871,27 +865,22 @@ else()
...
+
 # link Vulkan-Hpp C++ module into user project
 add_executable( YourProject main.cpp )
-target_link_libraries( YourProject PRIVATE Vulkan-HppModule )
+target_link_libraries( YourProject PRIVATE Vulkan::HppModule )
 ```
 
-Configuring the named module is straightforward; add any required Vulkan-Hpp feature macros listed in [Configuration](./Configuration.md) (or any C macros) to `target_compile_definitions`.
+When `VULKAN_HEADERS_ENABLE_MODULE` is enabled, Vulkan-Headers provides the `Vulkan-HppModule` target and the `Vulkan::HppModule` alias.
+Configuring the named module is straightforward; add any required Vulkan-Hpp feature macros listed in [Configuration](./Configuration.md) (or any C macros) to `target_compile_definitions` on `Vulkan-HppModule`.
```
sharadhr: CMake is rather unclear on how the ABI of the built module interfaces transmit information, so I would prefer users set the module up locally.

M2-TE:
This PR does not implement the changes discussed in the linked issue; it also has a bunch of issues that are already outlined above.

Instead of relying on the configs to provide `Vulkan::HppModule`, it should just be built locally all the time. Using either `find_package` or `FetchContent` should merely be a mechanism to access the headers and module sources. Using the release download for `FetchContent` as outlined in the linked issue should also be the preferred way of doing this.

docs/Usage.md
M2-TE: If anything, it should not be a CACHE BOOL variable either. Either way, building the module manually is probably the most stable way to go about it anyways.

puneetdixit200:
Updated this to keep the example building a project-local `Vulkan-HppModule` and removed the dependency on a `VulkanHeaders` config-package module target. I also refreshed the PR summary to match the current doc change.

docs/Usage.md
M2-TE: Does `VulkanHeaders` even work for find_package? I thought it just works with `Vulkan`.

sharadhr: The Vulkan-Headers CMakeLists [installs such a config](https://github.com/KhronosGroup/Vulkan-Headers/blob/015e25c3c91b70eb1a754d36fb14c4ba6ad9b0b9/CMakeLists.txt#L124) and [I use it too](https://github.com/sharadhr/caldera/blob/96ef30ea22163a2b76a078ace8fe28098b5a2f5b/tools/cmake/modules/libraries.cmake#L13), but this is through a non-standard way of acquiring the Vulkan SDK (I use vcpkg). 

The official installer is half-baked (see the linked issue) and only provides these config files for Linux. Linux system packages offering Vulkan-Headers will probably have this config file too (for instance, [Arch](https://archlinux.org/packages/extra/any/vulkan-headers/)), but we cannot expect to have it in the lowest denominator setup. 

I think there's nothing wrong with having this code branch per se because it clearly works, but we should also enumerate out all install possibilities.

sharadhr:
@puneetdixit200 I think this change is broadly useful, but I would not _delete_ the existing setups in favour of `find_package(VulkanHeaders...)` but rather, as I mentioned above, enumerate and elaborate on all possible configurations. There are many ways to acquire the Vulkan headers, and most of them are listed in the readme; their corresponding variations when setting up the module should be listed as well.

docs/Usage.md
M2-TE: In that case, adding another branch that showcases vcpkg might be useful to add here, that one could use the VulkanHeaders config.

sharadhr:
Suggest closing as a purely AI contribution with little obvious human input.

Issue #2584, opened 2026-05-26
Set up 32-bit Linux CI action and MinGW action


sharadhr:
Relates to #2583. This would be good to test cross-compiling scenarios. It will also catch issues such as #2468.

danyalahmed1995:
Hey, Opened #2585 for the 32-bit Linux CI part.

Added the Ubuntu 24.04 x86 matrix entry using gcc/g++ multilib with `-m32`. 

I did not covered the MinGW part because its a work on its own 😅

danyalahmed1995:
Alright Linux GCC/multilib is done, now I'm gonna do the minGW and clang runner ci . @sharadhr & @M2-TE two separate PR for both of them would be better right? Or you guys prefer one.

M2-TE:
I would also prefer to have them separated. Especially if either one causes problems.
Thanks for tackling this work!

danyalahmed1995:
@sharadhr The MinGW CI job is ready, but it immediately reproduced #2468: MinGW GCC 16.1 fails the C++ module link with duplicate definitions of the local `errorCategory` instance. (verified on my fork)

I split the source fix into #2596. The same MinGW CI job passes all 29 checks 

Once #2596 lands, I’ll update the standalone MinGW CI branch and open it separately for this issue.

Issue #2583, opened 2026-05-26
Build failure on 32 bit (Linux)


Zamundaaa:
This is with version 1.4.341.0 on Fedora: https://bugs.kde.org/show_bug.cgi?id=520589

The errors we got are all about the spaceship operator, and disabling it fixes the build:
```
/usr/include/vulkan/vulkan_structs.hpp: In member function ‘std::strong_ordering vk::DisplayPropertiesKHR::operator<=>(const vk::DisplayPropertiesKHR&) const’:
/usr/include/vulkan/vulkan_structs.hpp:56600:31: error: no match for ‘operator<=>’ (operand types are ‘const vk::DisplayKHR’ and ‘const vk::DisplayKHR’)
56600 |       if ( auto cmp = display <=> rhs.display; cmp != 0 )
      |                       ~~~~~~~ ^~~ ~~~~~~~~~~~
      |                       |               |
      |                       |               const vk::DisplayKHR
      |                       const vk::DisplayKHR
  • there are 7 candidates
    /usr/include/vulkan/vulkan_structs.hpp:56600:31:
    56600 |       if ( auto cmp = display <=> rhs.display; cmp != 0 )
          |                       ~~~~~~~~^~~~~~~~~~~~~~~
    • candidate 1: ‘template<unsigned int N, unsigned int M> std::strong_ordering vk::operator<=>(const ArrayWrapper1D<char, N>&, const ArrayWrapper1D<char, M>&)’ (reversed)
      /usr/include/vulkan/vulkan.hpp:165:24:
        165 |   std::strong_ordering operator<=>( ArrayWrapper1D<char, N> const & lhs, ArrayWrapper1D<char, M> const & rhs ) VULKAN_HPP_NOEXCEPT
            |                        ^~~~~~~~
      • template argument deduction/substitution failed:
        •   ‘const vk::DisplayKHR’ is not derived from ‘const vk::ArrayWrapper1D<char, N>’
          /usr/include/vulkan/vulkan_structs.hpp:56600:39:
          56600 |       if ( auto cmp = display <=> rhs.display; cmp != 0 )
                |                                       ^~~~~~~
    • candidate 2: ‘template<unsigned int N> std::strong_ordering vk::operator<=>(const std::string&, const ArrayWrapper1D<char, N>&)’ (reversed)
      /usr/include/vulkan/vulkan.hpp:211:24:
[...]

PR #2585, opened 2026-06-02
Add 32-bit Linux CI coverage

danyalahmed1995:
Adds a 32-bit Ubuntu CI matrix entry using gcc/g++ multilib and -m32 flags.

The sample build is skipped for this cross-compilation entry to avoid requiring 32-bit graphics/system libraries, while header generation and the unit test builds still run across the supported C++ standards.

This covers the Linux 32-bit CI case discussed in #2584 and helps exercise the kind of build configuration involved in #2583.

Tested on my fork:
- CI Build / ubuntu-24.04 x86, Ninja, g++-14 passed
- Full fork PR checks passed

Refs #2584.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -75,7 +75,7 @@ jobs:
       if: contains(matrix.env.os, 'ubuntu')
       run: |
         sudo apt update
-        sudo apt install -y libgl-dev libxcursor-dev libxi-dev libxinerama-dev libxrandr-dev ${{ matrix.env.ubuntu_packages || '' }}
+        sudo apt install -y libgl-dev libxcursor-dev libxi-dev libxinerama-dev libxrandr-dev ${{ contains(matrix.env.arch, 'x86') && format('{0}-multilib {1}-multilib', matrix.env.c, matrix.env.cxx) || '' }}
     - name: Setup MacOS
       if: contains(matrix.env.os, 'macos')
       run: brew install molten-vk
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -55,7 +55,7 @@ jobs:
       VAR_CMAKE_FLAGS: |
         ${{ matrix.env.c && format('-D CMAKE_C_COMPILER={0}', matrix.env.c) }} \
         ${{ matrix.env.cxx && format('-D CMAKE_CXX_COMPILER={0}', matrix.env.cxx) }} \
-        ${{ matrix.env.c_flags && format('-D CMAKE_C_FLAGS''={0}''', matrix.env.c_flags) }} \
+        ${{ matrix.env.c_flags && format('-D CMAKE_C_FLAGS={0}', matrix.env.c_flags) }} \
         ${{ matrix.env.cxx_flags && format('-D CMAKE_CXX_FLAGS''={0}''', matrix.env.cxx_flags) }} \
         ${{ matrix.env.linker_flags && format('-D CMAKE_EXE_LINKER_FLAGS=''{0}''', matrix.env.linker_flags) }} \
 
@@ -116,7 +116,7 @@ jobs:
 
     # Build samples using highest available C++ standard.
     - name: Build Samples with C++${{matrix.env.cxx_max}}
-      if: (!matrix.env.modules) && (!matrix.env.skip_samples) # Exclude module and cross-compilation runners.
+      if: (!matrix.env.modules) && (!contains(matrix.env.arch, 'x86')) # Exclude module and cross-compilation runners.
       run: |
         for BUILD_TYPE in Debug Release; do
           cmake -B build -G '${{matrix.env.gen}}' --fresh --preset samples \
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -31,7 +31,7 @@ jobs:
           { os: ubuntu-24.04,        cxx: g++-12,           c: gcc-12,         arch: x64, gen: Ninja,                 cxx_max: 23 },
           { os: ubuntu-24.04,        cxx: g++-13,           c: gcc-13,         arch: x64, gen: Ninja,                 cxx_max: 23 },
           { os: ubuntu-24.04,        cxx: g++-14,           c: gcc-14,         arch: x64, gen: Ninja,                 cxx_max: 23 },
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -31,6 +31,7 @@ jobs:
           { os: ubuntu-24.04,        cxx: g++-12,           c: gcc-12,         arch: x64, gen: Ninja,                 cxx_max: 23 },
           { os: ubuntu-24.04,        cxx: g++-13,           c: gcc-13,         arch: x64, gen: Ninja,                 cxx_max: 23 },
           { os: ubuntu-24.04,        cxx: g++-14,           c: gcc-14,         arch: x64, gen: Ninja,                 cxx_max: 23 },
+          { os: ubuntu-24.04,        cxx: g++-14,           c: gcc-14,         arch: x86, gen: Ninja,                 cxx_max: 23, c_flags: '-m32', cxx_flags: '-m32', linker_flags: '-m32', ubuntu_packages: 'gcc-14-multilib g++-14-multilib', skip_samples: true },
```
M2-TE:
```suggestion
          { os: ubuntu-24.04,        cxx: g++-14,           c: gcc-14,         arch: x86, gen: Ninja,                 cxx_max: 23, c_flags: '-m32', cxx_flags: '-m32', linker_flags: '-m32' },
```

I don't think we need to add additional options here. We could instead use the `arch: x86` variable to replace the other ones in the relevant steps. See other comments below.

.github/workflows/build.yml
```diff
@@ -71,7 +73,9 @@ jobs:
         arch: ${{matrix.env.arch}}
     - name: Setup Ubuntu
       if: contains(matrix.env.os, 'ubuntu')
-      run: sudo apt update && sudo apt install libgl-dev libxcursor-dev libxi-dev libxinerama-dev libxrandr-dev
+      run: |
+        sudo apt update
+        sudo apt install -y libgl-dev libxcursor-dev libxi-dev libxinerama-dev libxrandr-dev ${{ matrix.env.ubuntu_packages || '' }}
```
M2-TE:
```suggestion
        sudo apt install -y libgl-dev libxcursor-dev libxi-dev libxinerama-dev libxrandr-dev ${{ contains(matrix.env.arch, 'x86') && 'gcc-14-multilib g++-14-multilib' || '' }}
```

I believe the `contains(...)` should work inside the evaluated expressions `${{...}}`. This removes the need to set `ubuntu_packages` in the runner matrix.

.github/workflows/build.yml
```diff
@@ -112,7 +116,7 @@ jobs:
 
     # Build samples using highest available C++ standard.
     - name: Build Samples with C++${{matrix.env.cxx_max}}
-      if: (!matrix.env.modules) # Exclude module runners.
+      if: (!matrix.env.modules) && (!matrix.env.skip_samples) # Exclude module and cross-compilation runners.
```
M2-TE:
```suggestion
      if: (!matrix.env.modules) && (!contains(matrix.env.arch, 'x86')) # Exclude module and cross-compilation runners.
```

Same here as above, removing the need for `skip_samples`.

.github/workflows/build.yml
```diff
@@ -54,6 +55,7 @@ jobs:
       VAR_CMAKE_FLAGS: |
         ${{ matrix.env.c && format('-D CMAKE_C_COMPILER={0}', matrix.env.c) }} \
         ${{ matrix.env.cxx && format('-D CMAKE_CXX_COMPILER={0}', matrix.env.cxx) }} \
+        ${{ matrix.env.c_flags && format('-D CMAKE_C_FLAGS''={0}''', matrix.env.c_flags) }} \
```
M2-TE: Are C flags really needed here?
Only samples may need them to build the dependencies such as glfw. Or do the Vulkan C headers that get embedded into the Hpp headers need the `-m32` flag as well?

M2-TE (commented):
Good addition, thanks for the contribution!
Just got a few comments regarding the addition of new options in the matrix entry.

.github/workflows/build.yml
M2-TE: The string for the correct libs should of course not hardcode `gcc-14` and instead use the `env.matrix.c` and `env.matrix.cxx` variables. Woops!

sharadhr:
I would also recommend using Clang with `--target=i686-linux-gnu`.

.github/workflows/build.yml
danyalahmed1995: I added the `CMAKE_C_FLAGS` mainly to keep the C and CXX compiler configuration consistent for the 32-bit entry, in case CMake or dependencies enable any C checks during configuration/build.

The project path is primarily CXX, so yeah I can remove it if you prefer keeping the 32-bit targeting limited to `CMAKE_CXX_FLAGS` and linker flags.

danyalahmed1995:
@sharadhr I went with `gcc/g++` plus `-m32` because it matched the existing Ubuntu GCC matrix entries closely and kept the first 32-bit Linux job minimal. It also passed in my fork before opening the PR.

But if you want the CI entry to model cross-compilation more explicitly, I can switch it to Clang with `--target=i686-linux-gnu`.

.github/workflows/build.yml
danyalahmed1995: Updated this manually since the suggestion still had the package names hardcoded.

The install now derives the multilib package names from `matrix.env.c` and `matrix.env.cxx`, so the x86 entry does not need a separate `ubuntu_packages` field.

.github/workflows/build.yml
M2-TE: I think u're right, it wouldn't hurt to also supply the 32-bit flag for C. Particularly in case CMake does some shenanigans when it is missing.

There are a few surplus `'` characters in your added line, could you remove those?

danyalahmed1995: Done.

M2-TE:
Looks good to me, nice work! The clang runner can be a separate PR imo.

M2-TE (approved):

M2-TE:
Recent force-push removed the changes made from a previous one (a7c7c23 to 5fd2c8d), so please add them back in to allow runners to succeed. Specifically, the changes made to ubuntu packages that are to be installed need to be reintroduced.

Should be:
```cpp
sudo apt install -y libgl-dev libxcursor-dev libxi-dev libxinerama-dev libxrandr-dev ${{ contains(matrix.env.arch, 'x86') && format('{0}-multilib {1}-multilib', matrix.env.c, matrix.env.cxx) || '' }}
```

Force pushes are evil! ;)

danyalahmed1995:
Should be good now

asuessenbach:
Thanks @danyalahmed1995 and @M2-TE for your contribution, adding this CI step!

PR #2593, opened 2026-06-11
Add 32-bit Linux Clang CI coverage

danyalahmed1995:
Adds a 32-bit Ubuntu Clang CI matrix entry using --target=i686-linux-gnu.

This is a follow-up to #2585, which added the GCC/multilib 32-bit Linux CI entry. The new Clang entry keeps the same x86 sample-skip behavior while still running header generation and unit test builds across the supported C++ standards.

For the Clang x86 entry, the Ubuntu setup installs the GCC multilib runtime packages instead of deriving non-existent clang*-multilib package names.

Tested on my fork:
- CI Build / ubuntu-24.04 x86, Ninja, clang++-18 passed
- Full fork PR checks passed

Refs #2584.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -59,9 +59,9 @@ jobs:
       VAR_CMAKE_FLAGS: |
         ${{ matrix.env.c && format('-D CMAKE_C_COMPILER={0}', matrix.env.c) }} \
         ${{ matrix.env.cxx && format('-D CMAKE_CXX_COMPILER={0}', matrix.env.cxx) }} \
-        ${{ (contains(matrix.env.os, 'ubuntu') && contains(matrix.env.arch, 'x86') || matrix.env.c_flags) && format('-D CMAKE_C_FLAGS''={0} {1}''', contains(matrix.env.os, 'ubuntu') && contains(matrix.env.arch, 'x86') && (contains(matrix.env.cxx, 'clang') && '--target=i686-linux-gnu' || '-m32') || '', matrix.env.c_flags || '') }} \
-        ${{ (contains(matrix.env.os, 'ubuntu') && contains(matrix.env.arch, 'x86') || matrix.env.cxx_flags) && format('-D CMAKE_CXX_FLAGS''={0} {1}''', contains(matrix.env.os, 'ubuntu') && contains(matrix.env.arch, 'x86') && (contains(matrix.env.cxx, 'clang') && '--target=i686-linux-gnu' || '-m32') || '', matrix.env.cxx_flags || '') }} \
-        ${{ (contains(matrix.env.os, 'ubuntu') && contains(matrix.env.arch, 'x86') || matrix.env.linker_flags) && format('-D CMAKE_EXE_LINKER_FLAGS=''{0} {1}''', contains(matrix.env.os, 'ubuntu') && contains(matrix.env.arch, 'x86') && (contains(matrix.env.cxx, 'clang') && '--target=i686-linux-gnu' || '-m32') || '', matrix.env.linker_flags || '') }} \
+        ${{ contains(matrix.env.os, 'ubuntu') && contains(matrix.env.arch, 'x86') && format('-D CMAKE_C_FLAGS''={0}''', contains(matrix.env.cxx, 'clang') && '--target=i686-linux-gnu' || '-m32') || (matrix.env.c_flags && format('-D CMAKE_C_FLAGS''={0}''', matrix.env.c_flags)) }} \
+        ${{ contains(matrix.env.os, 'ubuntu') && contains(matrix.env.arch, 'x86') && format('-D CMAKE_CXX_FLAGS''={0}''', contains(matrix.env.cxx, 'clang') && '--target=i686-linux-gnu' || '-m32') || (matrix.env.cxx_flags && format('-D CMAKE_CXX_FLAGS''={0}''', matrix.env.cxx_flags)) }} \
+        ${{ contains(matrix.env.os, 'ubuntu') && contains(matrix.env.arch, 'x86') && format('-D CMAKE_EXE_LINKER_FLAGS=''{0}''', contains(matrix.env.cxx, 'clang') && '--target=i686-linux-gnu' || '-m32') || (matrix.env.linker_flags && format('-D CMAKE_EXE_LINKER_FLAGS=''{0}''', matrix.env.linker_flags)) }} \
 
     steps:
     - name: Checkout
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -34,11 +34,11 @@ jobs:
           { os: ubuntu-24.04,        cxx: g++-12,           c: gcc-12,         arch: x64, gen: Ninja,                 cxx_max: 23 },
           { os: ubuntu-24.04,        cxx: g++-13,           c: gcc-13,         arch: x64, gen: Ninja,                 cxx_max: 23 },
           { os: ubuntu-24.04,        cxx: g++-14,           c: gcc-14,         arch: x64, gen: Ninja,                 cxx_max: 23 },
-          { os: ubuntu-24.04,        cxx: g++-14,           c: gcc-14,         arch: x86, gen: Ninja,                 cxx_max: 23, c_flags: '-m32', cxx_flags: '-m32', linker_flags: '-m32' },
+          { os: ubuntu-24.04,        cxx: g++-14,           c: gcc-14,         arch: x86, gen: Ninja,                 cxx_max: 23 },
           { os: ubuntu-24.04,        cxx: clang++-16,       c: clang-16,       arch: x64, gen: Ninja,                 cxx_max: 20 },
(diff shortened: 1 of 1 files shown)


danyalahmed1995:
@M2-TE If this is all good than we can move to the minGW .

sharadhr:
I think instead of passing `--target=i686-linux-gnu` to `CFLAGS`, `CXXFLAGS`, etc, the clang invocation itself should be renamed as `clang++-18 --target=i686-linux-gnu` (for instance).

danyalahmed1995:
@sharadhr Done , should be good now

asuessenbach:
@sharadhr Could you elaborate a bit, why you think the `--target` flags should be part of the `CMAKE_C[XX]_COMPILER`?

.github/workflows/build.yml
```diff
@@ -56,8 +57,8 @@ jobs:
 
     env:
       VAR_CMAKE_FLAGS: |
-        ${{ matrix.env.c && format('-D CMAKE_C_COMPILER={0}', matrix.env.c) }} \
-        ${{ matrix.env.cxx && format('-D CMAKE_CXX_COMPILER={0}', matrix.env.cxx) }} \
+        ${{ matrix.env.c && format('-D CMAKE_C_COMPILER=''{0}''', matrix.env.c) }} \
+        ${{ matrix.env.cxx && format('-D CMAKE_CXX_COMPILER=''{0}''', matrix.env.cxx) }} \
         ${{ matrix.env.c_flags && format('-D CMAKE_C_FLAGS={0}', matrix.env.c_flags) }} \
         ${{ matrix.env.cxx_flags && format('-D CMAKE_CXX_FLAGS''={0}''', matrix.env.cxx_flags) }} \
         ${{ matrix.env.linker_flags && format('-D CMAKE_EXE_LINKER_FLAGS=''{0}''', matrix.env.linker_flags) }} \
```
M2-TE: excess `'` in here, same as in the *_flags below ;)

What if we bake the 32b flags into here, rather than in the runner matrix?
We have all the info we need, so it could boil down to something like:

```
${{ contains(matrix.env.arch, 'x86') && (contains(matrix.env.cxx, 'clang') && '--target=i686-linux-gnu' || '-m32') || '' }}
```

It removes the need to have any flags in the matrix, since they are very much implicit based on architecture and platform.

M2-TE (commented):
Sorry for being late to the party!
I think the 32b flags don't need to be in either the compiler variable or even the runner matrix at all. It should be possible to bake it into the compiler flags on-the-fly.

This would also apply to the `gcc` runner, see below for an example.

.github/workflows/build.yml
danyalahmed1995: Updated thanks , the Linux x86 GCC entry gets -m32, while the Linux x86 Clang entry gets --target=i686-linux-gnu.

M2-TE: To add to what @asuessenbach said above, should we really add these flags to `CMAKE_*_COMPILER`? That is, can't we add them to the three `CMAKE_*_FLAGS` instead?

danyalahmed1995: I am also gonna run a CI check on my fork since this change touches shared compiler flag logic and it affects both Linux x86 jobs.

danyalahmed1995: @M2-TE The change caused CMake to treat clang++-18 --target=i686-linux-gnu as a single compiler executable name, so compiler detection failed.

The target selection is now derived separately from the compiler configuration, which keeps the compiler path valid while still enabling the correct x86 Clang target. Verified on my fork with the full CI matrix passing.

M2-TE: Glad to see the current solution passing.

Could you test moving them into the `CMAKE_*_FLAGS`? While the current solution of setting it in the compiler name and separating by `;` works, it feels like it might invite unexpected behavior in the future (e.g. CMake changing how it consumes the compiler executable names).

sharadhr: > While the current solution of setting it in the compiler name and separating by ; works

This is a CMake-sanctioned way of hard-coding arguments to the compiler regardless of command-line invocation. It can be used in CMake toolchains to great effect.

danyalahmed1995: @M2-TE  Tested the `CMAKE_*_FLAGS` approach on my fork.

The first version applied the derived flag logic too broadly and affected non-Linux-x86 jobs, causing the Windows modules runner to fail. I adjusted it so the derived 32-bit target flags are only applied for Ubuntu x86 jobs:

* Ubuntu x86 GCC gets `-m32`
* Ubuntu x86 Clang gets `--target=i686-linux-gnu`
* existing Windows/macOS/module runner flags stay unchanged

The full fork CI matrix is passing with this version.

M2-TE: I like it. Just feels better to put flags into the flag variables imo. Great job!
@sharadhr are you satisfied with this solution as well?

sharadhr: If it works then I suppose we can merge; pinging @asuessenbach. 

That said I slightly prefer the semicolon-appended way. Having `-m32` (or `--target=...`) in the linker flags is superfluous.

M2-TE:
Before we merge this, I would like to somehow confirm whether we actually do compile in 32-bit. The CI syntax for ternary operators is kind of hard to read and might easily lead to errors.

What's the best way to confirm this?

.github/workflows/build.yml
danyalahmed1995: @sharadhr Both passed on my fork.

I kept the `CMAKE_*_FLAGS` version for now since @M2-TE  preferred putting the target flags there, but I can switch back to the semicolon-appended compiler form if that is the preferred direction.

I can also remove the derived `CMAKE_EXE_LINKER_FLAGS` part if you think the C/CXX flags are enough for this case.

danyalahmed1995: I will let you guys settle this out 😅

danyalahmed1995:
@M2-TE  I added a Linux x86 verification step for this.

It configures and builds a tiny C/C++ CMake project using the same `VAR_CMAKE_FLAGS` as the main jobs, with `static_assert(sizeof(void*) == 4)` to confirm the target is actually 32-bit.

Tested it on my fork as well, and the full CI matrix passed.

.github/workflows/build.yml
```diff
@@ -58,9 +59,9 @@ jobs:
...
-        ${{ matrix.env.c_flags && format('-D CMAKE_C_FLAGS={0}', matrix.env.c_flags) }} \
-        ${{ matrix.env.cxx_flags && format('-D CMAKE_CXX_FLAGS''={0}''', matrix.env.cxx_flags) }} \
-        ${{ matrix.env.linker_flags && format('-D CMAKE_EXE_LINKER_FLAGS=''{0}''', matrix.env.linker_flags) }} \
+        ${{ contains(matrix.env.os, 'ubuntu') && contains(matrix.env.arch, 'x86') && format('-D CMAKE_C_FLAGS''={0}''', contains(matrix.env.cxx, 'clang') && '--target=i686-linux-gnu' || '-m32') || (matrix.env.c_flags && format('-D CMAKE_C_FLAGS''={0}''', matrix.env.c_flags)) }} \
+        ${{ contains(matrix.env.os, 'ubuntu') && contains(matrix.env.arch, 'x86') && format('-D CMAKE_CXX_FLAGS''={0}''', contains(matrix.env.cxx, 'clang') && '--target=i686-linux-gnu' || '-m32') || (matrix.env.cxx_flags && format('-D CMAKE_CXX_FLAGS''={0}''', matrix.env.cxx_flags)) }} \
+        ${{ contains(matrix.env.os, 'ubuntu') && contains(matrix.env.arch, 'x86') && format('-D CMAKE_EXE_LINKER_FLAGS=''{0}''', contains(matrix.env.cxx, 'clang') && '--target=i686-linux-gnu' || '-m32') || (matrix.env.linker_flags && format('-D CMAKE_EXE_LINKER_FLAGS=''{0}''', matrix.env.linker_flags)) }} \
```
asuessenbach, line 64: When we derive the flags here from os, arch, and compiler, are the flags on line 41 needed at all? And those from line 37?

M2-TE: No, those were likely added back on accident in the most recent 95d6da9

danyalahmed1995: Oh, we don't need them anymore . The explicit flags on the Linux x86 matrix rows are redundant now that they’re derived from os, arch, and compiler in VAR_CMAKE_FLAGS.

I’ll remove them.

danyalahmed1995: @asuessenbach removed the flags and tested CI on the fork all good now.

M2-TE:
Talked with @sharadhr privately too regarding his previous comment:

> That said I slightly prefer the semicolon-appended way. Having -m32 (or --target=...) in the linker flags is superfluous.

And having the flags only in the compile identifier (via the semicolon appendage) is not the best solution, as that would not include the flags for the linker.

With that in mind, I think the current solution is good. Having the additional stage to verify that it actually compiles in 32-bit mode should also be kept imo, as it takes little time to execute and provides a solid sanity check.

All in all, I believe this looks ready for merging.

asuessenbach:
Thank you all guys for figuring out how to handle those additional flags for 32bit support. Very much appreciated!

PR #2596, opened 2026-06-24
Fix MinGW C++ module error category linkage

### LINKED ISSUE #2468: Having to patch `vulkan.hpp` when using `vulkan.cppm` with WinLibs GCC
In `vulkan.hpp`, I need to make `const std::error_category & errorCategory()` look like the following

```
  namespace detail
  {
    VULKAN_HPP_INLINE ErrorCategoryImpl errorCategoryInstance;
  }

  VULKAN_HPP_INLINE const std::error_category & errorCategory() VULKAN_HPP_NOEXCEPT
  {
    return detail::errorCategoryInstance;
  }
```

in order to not get linker errors when using `vulkan.cppm`. If anyone has a cleaner solution please let me know.

danyalahmed1995:
Fixes #2468.

MinGW GCC fails to link the C++ module tests because the function-local static `ErrorCategoryImpl` instance is emitted both by the module object and by module consumers.

For C++ module builds, this replaces the local static with a namespace-level inline instance. Header-only builds keep the existing implementation unchanged.

The generated `vulkan/vulkan.hpp` is updated to match the source snippet.

Tested with MinGW GCC 16.1.0:

* Before the change: module tests failed with duplicate definitions of `vk::errorCategory@vulkan()::instance`
* After the change: all 29 CI checks passed

The MinGW CI coverage is being kept as a separate change for #2584.

### DIFF
--- a/generator/snippets/Exceptions.hpp
+++ b/generator/snippets/Exceptions.hpp
@@ -55,10 +55,21 @@ class SystemError : public Error, public std::system_error
   virtual char const * what() const VULKAN_HPP_NOEXCEPT { return std::system_error::what(); }
 };
 
+#if defined( VULKAN_HPP_CXX_MODULE )
+namespace detail
+{
+  inline ErrorCategoryImpl errorCategoryInstance;
+}
+#endif
+
 VULKAN_HPP_INLINE std::error_category const & errorCategory() VULKAN_HPP_NOEXCEPT
 {
+#if defined( VULKAN_HPP_CXX_MODULE )
+  return detail::errorCategoryInstance;
+#else
   static ErrorCategoryImpl instance;
   return instance;
+#endif
 }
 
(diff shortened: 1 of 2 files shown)


sharadhr:
As already mentioned in the linked issue, this is a MinGW GCC toolchain bug. We should not need to fix our behaviour (and in this case, increase complexity by special-casing how the symbol is defined for modules versus headers) just to make a new CI run pass. In fact, having it fail will be a stronger signal for the toolchain developers to fix their bug. 

The existing behaviour is completely legitimate code, especially since other toolchains (MSVC, Clang) can compile and link it just fine. Additionally, I personally prefer the function-local static—it is an obvious Meyers singleton.

danyalahmed1995:
@sharadhr Oh right , Yes the original CI did actually failed without a fix so i thought i need to fix it but this clears it now . I will just open the pr with the MinGW thank you for correcting me.

PR #2597, opened 2026-06-24
Add MinGW GCC CI coverage

### LINKED ISSUE #2584: Set up 32-bit Linux CI action and MinGW action

danyalahmed1995:
Closes #2584.

Adds a dedicated Windows x64 MinGW GCC CI job using MSYS2 UCRT64.

The job:

* installs the MinGW GCC toolchain through MSYS2
* verifies the selected compiler and target
* generates the Vulkan-Hpp headers
* builds the samples and tests with C++23 module support enabled
* runs the test suite

### Fork validation

The validation run completed with 27 passing checks and two failures:

* `Windows x64, MinGW GCC, Ninja`

  * reproduces the known MinGW GCC module linker failure tracked in #2468
  * fails with duplicate definitions of the function-local `errorCategory` instance
* `ubuntu-24.04 x64, Ninja, clang++-18`

  * this is an existing matrix job and is separate from the newly added MinGW job

The MinGW failure is intentionally left visible as a toolchain signal. As discussed in #2596, the existing Vulkan-Hpp implementation is valid and links successfully with MSVC and Clang, so this PR does not add a project-side workaround for the MinGW GCC bug.

Refs #2468.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -19,6 +19,7 @@ jobs:
         shell: bash
 
     # Keep the full compiler matrix visible so CI consumers can inspect every job.
+    # Matrix smoke-test push marker: behavior and coverage remain unchanged.
     strategy:
       matrix:
         env: [
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -188,3 +188,61 @@ jobs:
             ctest -j --output-on-failure --test-dir build -C $BUILD_TYPE
           done
         done
+  # MinGW requires a dedicated MSYS2 toolchain and shell.
+  mingw:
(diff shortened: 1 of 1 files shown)


sharadhr:
@asuessenbach that particular CI is _meant_ to fail; see #2468. I think it is ready for merge otherwise.

asuessenbach:
> that particular CI is meant to fail; see https://github.com/KhronosGroup/Vulkan-Hpp/issues/2468.

I know. And I'll merge it as soon as the root cause of failure is resolved.

PR #2579, opened 2026-05-19
DispatchLoader: Allow passing explicit `vkGetDeviceProcAddr` to dynamic dispatch loader

M2-TE:
This will resolve #2578.

This function signature overload for `DispatchLoaderDynamic::init` previously did not do anything with the `vkGetDeviceProcAddr` param:

```cpp
// This interface does not require a linked vulkan library.
void init( VkInstance                instance,
           PFN_vkGetInstanceProcAddr getInstanceProcAddr,
           VkDevice                  device              = {},
           PFN_vkGetDeviceProcAddr /*getDeviceProcAddr*/ = nullptr ) VULKAN_HPP_NOEXCEPT
{
  VULKAN_HPP_ASSERT( instance && getInstanceProcAddr );
  vkGetInstanceProcAddr = getInstanceProcAddr;
  init( Instance( instance ) );
  if ( device )
  {
    init( Device( device ) );
  }
}
```

This PR does 4 things:
1. Changed aforementioned function to now use the `vkGetDeviceProcAddr` parameter
2. Removed default parameter assignments from aforementioned function
3. Added function overload that _only_ takes instance and `vkInstanceProcAddr`
4. Removed surplus load of `vkGetDeviceProcAddr` in `void init( Device deviceCpp )`

I'm marking this as draft for now, as CI does not run these tests and thus, this change could be dangerous without proper testing. Please give it a try and see if it works for you @PancakeTAS

PR #2615, opened 2026-07-10
DispatchLoader: Added/Changed ctors to reflect #2579

### LINKED ISSUE #2614: Regression in vk::detail::DispatchLoaderDynamic constructor causes assert
In #2579 it was made possible to pass `vkGetDeviceProcAddr` directly via init function again.

I just compiled the latest git version to use in my project, but I encountered this issue.

In the process, the default assignments for the device-level `device` and `getDeviceProcAddr` were removed and a new function for only instance-level initialization was created.

However the constructor was left untouched:
```cpp
      // This interface does not require a linked vulkan library.
      DispatchLoaderDynamic( VkInstance                instance,
                             PFN_vkGetInstanceProcAddr getInstanceProcAddr,
                             VkDevice                  device            = {},
                             PFN_vkGetDeviceProcAddr   getDeviceProcAddr = nullptr ) VULKAN_HPP_NOEXCEPT
      {
        init( instance, getInstanceProcAddr, device, getDeviceProcAddr );
      }
```

If `device` and `getDeviceProcAddr` are unset here, this will trigger an assert in `init(...)`, as the check for `device` was removed:
```cpp
      // This interface does not require a linked vulkan library.
      void init( VkInstance instance, PFN_vkGetInstanceProcAddr getInstanceProcAddr, VkDevice device, PFN_vkGetDeviceProcAddr getDeviceProcAddr )
        VULKAN_HPP_NOEXCEPT
      {
        init( instance, getInstanceProcAddr );
        VULKAN_HPP_ASSERT( device && getDeviceProcAddr );
        vkGetDeviceProcAddr = getDeviceProcAddr;
        init( Device( device ) );
      }
```

A second constructor for just `instance` and `getInstanceProcAddr` should be added.

@M2-TE @asuessenbach

M2-TE:
Resolves #2614.

### DIFF
--- a/generator/snippets/DispatchLoaderDynamic.hpp
+++ b/generator/snippets/DispatchLoaderDynamic.hpp
@@ -27,8 +27,8 @@ public:
     // This interface does not require a linked vulkan library.
     DispatchLoaderDynamic( VkInstance                instance,
                             PFN_vkGetInstanceProcAddr getInstanceProcAddr,
-                            VkDevice                  device            = {},
-                            PFN_vkGetDeviceProcAddr   getDeviceProcAddr = nullptr ) VULKAN_HPP_NOEXCEPT
+                            VkDevice                  device,
+                            PFN_vkGetDeviceProcAddr   getDeviceProcAddr) VULKAN_HPP_NOEXCEPT
     {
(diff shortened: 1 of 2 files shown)


asuessenbach:
Thanks for fixing this issue that fast!

PR #2622, opened 2026-07-20
Support '+' separator in feature depends

dgkoch:
The Vulkan dependency syntax defines '+' as logical AND and ',' as logical OR.
The `depends` attribute on `<feature>` elements always intends AND semantics,
but historically used ',' as the separator. Future XML corrections will use
the proper '+' separator (see https://gitlab.khronos.org/vulkan/vulkan/-/issues/4912).

Switch from `tokenize(..., ",")` to `tokenizeAny(..., ",+")` so the parser
accepts both forms — backward-compatible with current XML, forward-compatible
with the corrected XML.

▎ GenAI disclosure: Draft assisted by Claude (claude-sonnet-4-6[1m], Anthropic, commercial SaaS — Claude Code CLI).

### DIFF
--- a/generator/VkXMLParser.cpp
+++ b/generator/VkXMLParser.cpp
@@ -1410,7 +1410,9 @@ Feature parseFeature( tinyxml2::XMLElement const * element )
     }
     else if ( attribute.first == "depends" )
     {
-      feature.depends = tokenize( attribute.second, "," );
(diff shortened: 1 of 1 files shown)


asuessenbach:
Thanks for this little adjustment.

PR #2647, opened 2026-08-04
Do not `export module std`

sharadhr:
We don't need to publish the entire interface of `module std;`; anything that is visible and reachable will automatically be populated by `module vulkan`. 

This is an experiment to test how reducing the public interface of `module vulkan` changes its usability. [CI runs](https://github.com/sharadhr/Vulkan-Hpp/actions/runs/30912610921/job/92003545651?pr=3) already suggest that fully-qualified types for return values of Vulkan functions such as `vk::raii::Instance::enumeratePhysicalDevices` will require `import std` in consumer code.


sharadhr:
CI failures are interesting...

tests/FunctionCallsRAII/FunctionCallsRAII.cpp
sharadhr, line 47: Could also be fixed with:
```suggestion
    auto physicalDevices = instance.enumeratePhysicalDevices();
```

sharadhr:
Ready for review. 

This reworks the public interface of `module vulkan`; certain symbols will now have reachability but not visibility, for instance the return type of `vk::raii::Instance::enumeratePhysicalDevices` is an instantiation of `std::vector` which will now need to be _explicitly_ `import`ed with `import std;`. An illustration from a previous CI failure:

```
/home/runner/work/Vulkan-Hpp/Vulkan-Hpp/tests/FunctionCallsRAII/FunctionCallsRAII.cpp:46:10: error: declaration of 'vector' must be imported from module 'std' before it is required
   46 |     std::vector<vk::raii::PhysicalDevice> physicalDevices = instance.enumeratePhysicalDevices();
      |          ^
/lib/share/libc++/v1/std/vector.inc:12:14: note: declaration here is not visible
   12 |   using std::vector;
      |              ^
```

[This blog post](https://vector-of-bool.github.io/2019/03/31/modules-2.html) goes into much greater detail about these rules, which we are now seeing in practice.

tests/FunctionCallsRAII/FunctionCallsRAII.cpp
asuessenbach: This might fix the issue, but would contradict the goal of those tests: make ever call explicit to automatically catch API changes.

asuessenbach:
Looks good.
Thanks for your contribution, again!

PR #2664, opened 2026-08-20
Include <cstdlib> on APPLE in order to resolve std::getenv

### LINKED ISSUE #2662: Missing <cstdlib> include in vulkan.hpp for std::getenv causes build failures with C++ modules
vulkan.hpp makes a call to std::getenv when attempting to resolve the Vulkan library path on macOS, but it does not include <cstdlib>. 

It is transitively/implicitly included from other headers so will be fine for non-module compiles, which is why it was working before. But when using clang modules it causes a compile error, because visibility and unit boundaries are more tightly enforced.

Specifically this is being seen during a dependency roll for Chromium (on MacOS bots), causing errors like this:
```
   1 ../../third_party/swiftshader/include/vulkan/vulkan.hpp:22912:37: error: declaration of 'getenv' must be imported from module
     '//build/modules:system.std.cstdlib' before it is required
   2  22912 |           if ( !m_library && ( std::getenv( "DYLD_FALLBACK_LIBRARY_PATH" ) == NULL ) )
   3        |                                     ^
   4 ../../third_party/libc++/src/include/cstdlib:130:9: note: declaration here is not visible
   5   130 | using ::getenv _LIBCPP_USING_IF_EXISTS;
```

I believe that this should be resolvable by making sure that `#include <cstdlib>` is added to the system library includes at the top of vulkan.hpp

asuessenbach:
Resolves #2662

### DIFF
--- a/generator/snippets/DynamicLoader.hpp
+++ b/generator/snippets/DynamicLoader.hpp
@@ -31,32 +31,32 @@
         {
           m_library = dlopen( "libvulkan.1.dylib", RTLD_NOW | RTLD_LOCAL );
         }
-		// modern versions of macOS don't search /usr/local/lib automatically contrary to what man dlopen says
-		// Vulkan SDK uses this as the system-wide installation location, so we're going to fallback to this if all else fails
-		if ( !m_library && ( std::getenv("DYLD_FALLBACK_LIBRARY_PATH") == NULL ) )
-		{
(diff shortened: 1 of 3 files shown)


sharadhr:
We should probably add some tests to catch this.

zoddicus:
I have confirmed in a local build that this solves the issue I was seeing

PR #2666, opened 2026-08-20
cmake: reduce duplication in samples build setup

sharadhr:
Some cleanup for the samples build logic, in preparation for `module`ifying all the samples.


asuessenbach:
Great reduction in complexity!

Thanks a lot!

PR #2675, opened 2026-08-27
Parse required properties

christophe-lunarg:
https://gitlab.khronos.org/vulkan/vulkan/-/merge_requests/8501 introduced required properties in vk.xml.

This caused warnings in Gitlab tests.

Another thing that might be a problem, the MR improved the completness of the feature requirements for Vulkan core versions and C.I. generate this message:

```
caught exception: vk.xml: Spec error on line 21226: feature <samplerMirrorClampToEdge> with struct <VkPhysicalDeviceVulkan12Features> already listed as required for feature <VK_VERSION_1_2>
```

It looks like the parser code has some handling for the "depends" attributes but it's not clear to me how much.

With Vulkan 1.2, samplerMirrorClampToEdge requires 'VK_KHR_sampler_mirror_clamp_to_edge' support.
With Vulkan 1.4, samplerMirrorClampToEdge must be supported inconditionally.

I would appriciate some helps regarding these issues.

### DIFF
--- a/generator/VkXMLParser.cpp
+++ b/generator/VkXMLParser.cpp
@@ -49,6 +49,7 @@ ExtensionRequireEnumConstant parseExtensionRequireEnumConstant( tinyxml2::XMLEle
 Extensions                   parseExtensions( tinyxml2::XMLElement const * element );
 Feature                      parseFeature( tinyxml2::XMLElement const * element );
 FeatureElement               parseFeatureElement( tinyxml2::XMLElement const * element );
+PropertyElement              parsePropertyElement( tinyxml2::XMLElement const * element );
 Require                      parseFeatureRequire( tinyxml2::XMLElement const * element );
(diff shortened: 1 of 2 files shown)


asuessenbach:
@christophe-lunarg Thanks for this PR, I'll have a closer look at it asap.

Regarding the exception: I think, that should be resolved with #2674.

asuessenbach:
It looks really good. I'll add some checks later on.

christophe-lunarg:
Oh great, thanks ! :)

asuessenbach:
@christophe-lunarg Note, with #2677, I'll add a couple of checks for that property elements.
I had encountered some errors with the vk.xml I tried from [https://gitlab.khronos.org/vulkan/vulkan/-/merge_requests/8501](https://gitlab.khronos.org/vulkan/vulkan/-/merge_requests/8501)

